PR #72, opened 2015-07-07
More than one playfield

jabdoa2:
Refactor Playfield and BallDevice to allow more than one playfield.


jabdoa2:
Sorry wrong branch

PR #96, opened 2015-08-13
implement auto_disable_time

jabdoa2:


toomanybrians:
Can you add self.delay.remove('disable_shoot_again') to the disable() method?

jabdoa2:
check

PR #117, opened 2015-10-19
Balldevice problems

jabdoa2:
Just tests. Three out of four currently fail because of the count_balls() problems.


jabdoa2:
added a fix for both problems

PR #129, opened 2016-01-12
Add ball search

jabdoa2:
initial ball search which just fires the coils of BallDevices


toomanybrians:
Will merge updated branch instead

PR #138, opened 2016-03-25
Implemented LED stacks and removed cache, updated tests

toomanybrians:
Now led_player and shows fully restore lower priority led states when they stop or when a mode ends.


jabdoa2:
Looks very good to me. Two small notices:
1. You may use self.debug_log instead of if self.debug: self.log.debug()
2. Please fix the two PEP8 style issues

PR #140, opened 2016-03-25
Support gottlieb trough

jabdoa2:
Support a) from https://missionpinball.com/forum/topic/gottlieb-trough-configuration/#post-8508

### DIFF
--- a/mpf/core/config_validator.py
+++ b/mpf/core/config_validator.py
@@ -141,6 +141,7 @@ coils:
     number: single|str|
     number_str: single|str|
     pulse_ms: single|int|None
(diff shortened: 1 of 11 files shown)


jabdoa2:
also added support for b)

mpf/core/config_validator.py
```diff
@@ -140,6 +141,7 @@
     number: single|str|
     number_str: single|str|
     pulse_ms: single|int|None
+    long_pulse_ms: single|int|None
```
jabdoa2: This is not used anywhere. Can we remove it again?

PR #253, opened 2016-05-06
Refactor fast serial communicator

jabdoa2:
- On top of #252 (merge first)
- Please test on real hardware


jabdoa2:
Does not work on real hw: "ok. yeah, very strange. SA: is getting sent, but it seems that no commands are being received? But that can’t be true because it can ID the connections, so I don’t know"

jabdoa2:
superseeded by tickless_mpf branch

Issue #233, opened 2016-05-04
Implement checks for FAST triggers & board location


toomanybrians:
FAST I/O boards can only program hardware rules "locally" within the board, e.g. the switch and driver must be on the same board.

The exception is switches 00-07 (which are the first switches on whichever board is connected first in the loop to the controller). Those switches are put in a priority packet on the loop network for minimal latency and sent to all other boards immediately.

The idea is those are typically flipper buttons which would be in an I/O board in the cabinet controlling drivers on an I/O board under the pf. Otherwise this is typically a non-issue, since for example a pop bumper switch and coil are near each other and can be plugged into the same board no prob.

So I want to implement checks in the FAST platform to ensure that hw rules are set only with switches that are on the same board as their drivers, with the exception of switches 00-07 which can control any driver.

jabdoa2:
fixed in dev

Issue #221, opened 2016-05-02
Allow FAST driver & switch configs by board number


toomanybrians:
Currently all FAST drivers and switches are cumulatively numbered starting with 00. Proposal is to allow an alternate number scheme where a dash in the number signifies board number, switch number (or driver number), like this:

```
switches:
  switch1:
    number: 0-0   # board 0, sw 0
  switch2:
    number: 2-5  # board 2, sw 5
```

same for drivers

Currently we do query each IO board at boot, and we know how many switches and drivers are on each and the order, so this would be fairly straightforward to implement

jabdoa2:
fixed in dev

PR #332, opened 2016-05-26
Tickless mpf

jabdoa2:
WIP. changes for platforms to work without polling for better latency


jabdoa2:
fixed #221 and fixes #233

PR #367, opened 2016-06-19
get opp switch input status every tick

HarryXS:
Changed the poll rate of the switch-inputs from every tenth tick to every tick.
Still not really happy with this since it should be faster than every 33ms.


jabdoa2:
For 0.30, you can increase the tick rate in config using:

```
mpf:
  hz: 100
```

In 0.31 this will be different and I will make sure OPP runs with more then 30Hz by default.

PR #369, opened 2016-07-05
Fix bug with show token in led express config

qcapen:
Fixed bug with show tokens in led express config when also providing a
fade value (ex: (secondcolor)-f500ms).  Added test to catch bug.


qcapen:
This resolves issue #371

Issue #370, opened 2016-07-08
Gamma Color Correction Profile


BrewinBombers:
The LEDs commonly used with pinball machines do not produce terribly accurate colors. Quinn whipped up this gamma correction scriptlet that makes oranges and darker colors much closer to their intended colors. 

This should be baked right into MPF at some point.

`from mpf.core.scriptlet import Scriptlet
from mpf.core.rgb_color import RGBColorCorrectionProfile

class GammaCorrectionProfile(Scriptlet):
    """To 'activate' your scriptlet:
        1. Copy it to your <your_machine_folder>/scriptlets/ folder
        2. Add an entry to the 'Scriptlets:' section of your machine config files
        3. That entry should be 'your_scriptlet_file_name.YourScriptletName'
    """

```
def on_load(self):
    """Called automatically when this scriptlet is loaded."""

    self.log.info("Adding custom gamma_correction_profile color correction profile")

    # you can register handlers to act on system events
    self.machine.events.add_handler('init_phase_5', self.add_color_correction_profile)

def add_color_correction_profile(self):
    profile = RGBColorCorrectionProfile('gamma_correction_profile')

    channel = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
               0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1,
               1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2,
               2, 3, 3, 3, 3, 3, 3, 3, 4, 4, 4, 4, 4, 5, 5, 5,
               5, 6, 6, 6, 6, 7, 7, 7, 7, 8, 8, 8, 9, 9, 9, 10,
               10, 10, 11, 11, 11, 12, 12, 13, 13, 13, 14, 14, 15, 15, 16, 16,
               17, 17, 18, 18, 19, 19, 20, 20, 21, 21, 22, 22, 23, 24, 24, 25,
               25, 26, 27, 27, 28, 29, 29, 30, 31, 32, 32, 33, 34, 35, 35, 36,
               37, 38, 39, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 50,
               51, 52, 54, 55, 56, 57, 58, 59, 60, 61, 62, 63, 64, 66, 67, 68,
[...]

PR #372, opened 2016-07-08
Added default color correction profile

qcapen:
Added new 'default' color correction profile (and associated tests).  These changes add the profile to the machine, but do not automatically associate it with any led devices. It can be overridden by the user in their machine config.

This addresses issue #370.

You can decide whether or not to pull this into 0.30 (even though it is a new feature, although very limited in scope) or wait for 0.31.


qcapen:
I recommend including this in 0.30 as there is some important new validation for LED color correction profiles (exception is now thrown on a bad color correction profile value whereas before the profile was simply skipped and a warning was logged).

PR #737, opened 2017-01-04
Mode timer reset on mode start

### LINKED ISSUE #732: Mode timer continue on next ball
Add option to continue a mode_timer on the next ball of the same player.

jabdoa2:
fix #732 

@toomanybrians should this default to true or false? true would be backward compatible


toomanybrians (approved):
True seems like the most intuitive, and since that's also backwards compatible, I like it.

PR #743, opened 2017-01-05
Default highscores

### LINKED ISSUE #232: Add default high scores
Currently you set default high scores by hacking the high_scores data file. Should add a way to specify these via the config, as well as high score reset timers (days, months, etc) as well as reset methods

jabdoa2:
fix  #232

### DIFF
--- a/mpf/modes/high_score/config/high_score.yaml
+++ b/mpf/modes/high_score/config/high_score.yaml
@@ -17,6 +17,13 @@ high_score:
       - HIGH SCORE 2
       - HIGH SCORE 3
       - HIGH SCORE 4
+  defaults:
+    score:
+      - BRI: 4242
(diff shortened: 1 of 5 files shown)


mpf/modes/high_score/config/high_score.yaml
```diff
@@ -17,6 +17,13 @@ high_score:
       - HIGH SCORE 2
       - HIGH SCORE 3
       - HIGH SCORE 4
+  defaults:
```
jabdoa2, line 20: @toomanybrians what do you think about the syntax? If you are fine with this we can merge.

toomanybrians: Looks good to me

toomanybrians (approved):

PR #748, opened 2017-01-15
Add psu device and refactor score_reels

jabdoa2:


jabdoa2:
Superseeded by #825 and #821

Issue #754, opened 2017-01-22
Crash in Shots


jabdoa2:
```
INFO : SwitchController : <<<<< switch: s_switch4, State:0 >>>>>
INFO : SwitchController : <<<<< switch: s_switch4, State:1 >>>>>
INFO : Mode.attract : Mode Starting. Priority: 10
INFO : Machine : Shutting down...
INFO : Machine : Shutting down...
ERROR : asyncio : Exception in callback EventManager._done(<Task finishe...scriptable",)>)
handle: <Handle EventManager._done(<Task finishe...scriptable",)>)>
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/events.py", line 125, in _run
    self._callback(*self._args)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/mpf/core/events.py", line 530, in _done
    future.result()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/mpf/core/events.py", line 462, in _run_handlers_sequential
    handler.callback(queue=queue, **merged_kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/mpf/core/mode.py", line 487, in _control_event_handler
    callback(mode=self)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/mpf/devices/shot.py", line 696, in reset
    self.jump(mode, state=0)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/mpf/devices/shot.py", line 655, in jump
    if state == self.player[self.get_profile_by_key('mode', mode)['settings']['player_variable']]:
```

From: https://groups.google.com/forum/#!topic/mpf-users/WgK4AOVWEvM

jabdoa2:
This is also a symptom of #838 since it also happens during fuzzing

jabdoa2:
fixed in dev

PR #820, opened 2017-03-14
event priorities and cleanup

jabdoa2:
- add explicit event priorities
- validate event priorities. two handlers on the same class instance
  cannot bind on the same event with the same priority
- fix some types with the help of mypy
- cleanup some code


jabdoa2:
fixes #754

jabdoa2:
merged into next

PR #827, opened 2017-03-22
Prevent 'ff' value propagating to color conversion for leds

unRARed:
Setting a specific value (tried "white" and "red") for `color` in the leds config caused **mpf** to hang. According to the docs, `color` is supposed to be optional, but leaving it out caused the value `"ff"` to propagate into the RGBColor utility function as the `color` variable which seems incorrect.

### DIFF
--- a/mpf/config_players/led_player.py
+++ b/mpf/config_players/led_player.py
@@ -47,7 +47,7 @@ class LedPlayer(DeviceConfigPlayer):
 
     @staticmethod
     def _led_color(led, instance_dict, full_context, color, **s):
-        if color == "on":
+        if color in ['on', 'ff']:
             color = led.config['default_color']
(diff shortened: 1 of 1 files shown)


unRARed:
Actually, this might be an issue with the led docs. Don't merge yet...

toomanybrians:
Which `color` config is this? In an led_player / show? Can you post a config example? I'm just curious as what the root cause of this is... like where is the `ff`coming from?

unRARed:
I think I got mixed up with the docs for lights vs leds. I got it figured out.

toomanybrians:
In the next version of MPF, we're merging lights and LEDs to prevent this exact kind of confusion. Don't worry.. not the first time! :)

PR #829, opened 2017-03-23
Bcp changes

qcapen:
Updated BCP version to 1.1.  Update code to restore some of the old behavior that was in version 0.21 and used in the Unity MPF interface.  Removed some unused BCP features.  Added ability to monitor various groups of events: all events, devices, switches, modes, core events (player, ball, turn, etc.), player vars, machine vars.  Added ability to stop monitoring and unregister triggers.  Added new tests.

### DIFF
--- a/mpf/_version.py
+++ b/mpf/_version.py
@@ -10,9 +10,9 @@ PyPI.
 
 """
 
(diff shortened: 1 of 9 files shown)


jabdoa2 (approved):
Looks very good

PR #841, opened 2017-03-27
add mod placeholder and test

### LINKED ISSUE #832: Use modulus operators in conditional events?
Trying to use event player to post an event every 8th time a conditional event is posted. Specifically, every time the hit_jukebox event from a counter logic block posts with a value divisible by 8, I want it to post the same event. 

So, I tried to do this:
```
event_player:
  hit_jukebox{count%8==0}:
    start_light_lock
```
but it crashes my game.

There are other ways around this, but this seemed like the simplest (i.e. fewest lines of code/config) to me. 

Log attached. No rush. I can do what I need to do in mode code. 
[crash using modulus in conditional event.zip](https://github.com/missionpinball/mpf/files/869673/crash.using.modulus.in.conditional.event.zip)

jabdoa2:
fix #832


GabeKnuth:
Works great! Thanks.

PR #845, opened 2017-03-30
Force 'utf8' encoding when loading config files

unRARed:
We ran into an encoding issue trying to run mpf for a machine config (that I have been iterating on from a Mac) on a Windows 10 system. It seems like this was addressed prior via https://github.com/missionpinball/mpf/commit/cbaff2b4b15268de039fba173d6817aa381d0cb8, but I checked and neither desktop.ini, Thumbs.db or .DS_Store were found among my machine config files. Passing the encoding when loading the yaml files as changed here allowed Windows to load my machine config.

### DIFF
--- a/mpf/file_interfaces/yaml_interface.py
+++ b/mpf/file_interfaces/yaml_interface.py
@@ -188,7 +188,7 @@ class YamlInterface(FileInterface):
     @staticmethod
     def get_config_file_version(filename):
         """Return config file version."""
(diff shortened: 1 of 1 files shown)


jabdoa2:
Good catch. I assumed that utf8 would be the default encoding nowadays but that seems to only hold for byte to string conversations and not file open (see: http://stackoverflow.com/questions/36303919/python-3-0-open-default-encoding)

PR #871, opened 2017-04-13
fix #598 by changing event args order

### LINKED ISSUE #598: Bug in timer control event handling
When trying to use an event with args as a control event in a timer, it screws up the timer. Here's an example config:

```
timers:
    world_tour_timer:
        start_value: 20
        max_value: 20
        tick_interval: 1s
        direction: down
        control_events:
          - event: timer_wt_intro_timer_complete
            action: start
          - event: player_world_tour_world_tour_timer_tick{value == 5} # This is the problem
            action: change_tick_interval
            value: 1.5
          - event: add_time
            action: add
            value: 3
          - event: s_subwayopto_active
            action: stop
```

With this config, I want the timer to change the tick interval to 1.5s for the last 5 seconds. If I put a normal event in there (like a switch that I push to test it), it works, but with the event that's in the above config, it just resets the timer to 0.

jabdoa2:
Please carefully review the change in event.py. Do we want this?

### DIFF
--- a/mpf/core/events.py
+++ b/mpf/core/events.py
@@ -525,8 +525,8 @@ class EventManager(MpfController):
             # in while we were processing previous handlers
 
             # merge the post's kwargs with the registered handler's kwargs
-            # in case of conflict, posts kwargs will win
-            merged_kwargs = dict(list(handler.kwargs.items()) + list(kwargs.items()))
+            # in case of conflict, handlers kwargs will win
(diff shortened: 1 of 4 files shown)


qcapen (approved):
Simple change to give preference to handler kwargs in case of a conflict makes sense. Those kind of kwarg name collisions certainly could be difficult to debug.

Issue #480, opened 2016-09-13
Game Lifecycle


jabdoa2:
Extend the game lifecycle and define events around it. One atomic event per transition and a queue event before/after the transition.

States: not running -> starting -> running -> starting_ball -> play -> ending_ball -> (...) -> stopping -> not running

Events: (TODO: look up previously existing events and add them)
- player_will_add/adding/added (parameters: player_num, late_add_in_game)
- game_will_start
- game_starting (queue)
- game_started
- player_turn_will_start/ing/ed (parameters: player_num) 
- ball_will_start (parameters: player_num, ball, player_change, is_extra_ball, opt: extra_ball_num_in_game, extra_ball_num_in_ball) 
- ball_starting (queue)
- ball_started
- ball_will_end (parameters: player_num, ball, player_will_change, next_will_be_extra_ball) 
- ball_ending (queue)
- ball_ended
- player_turn_will_stop/ing/ed
- game_will_stop
- game_stopping (queue)
- game_stopped

qcapen:
The following events have been removed in 0.50 and will need to be migrated:

- player_turn_start (replaced with player_turn_will_start or player_turn_started)
- player_turn_stop (replaced with player_turn_will_end, player_turn_ending, player_turn_ended depending upon how event was used)
- player_add_success (replaced with player_added)

qcapen:
Implemented in 0.50

Issue #479, opened 2016-09-13
Mode Lifecycle


jabdoa2:
Extend the mode lifecycle and define events around it. One atomic event per transition and a queue event before/after the transition.

States: not running -> starting -> running -> stopping -> not running

Events:
- mode_will_start, we setup config players here, devices, etc.
- mode_starting (queue), mode configs and config players can hook this
- mode_started (like before), enable devices
- mode_will_stop
- mode_stopping (queue), mode configs and config players can hook this
- mode_stopped

qcapen:
Implemented in 0.50

PR #879, opened 2017-04-19
Game and mode lifecycle events

qcapen:
Implements issues #479 and #480


qcapen:
Also converted game mode to asyncio

mpf/modes/game/code/game.py
```diff
@@ -31,16 +32,174 @@ def __init__(self, machine, config, name, path):
...
+                    '%', self.machine.config['game']['add_player_switch_tag']),
+                self.request_player_add)
+
+        yield from self._start_game()
+
+        # Game loop
+        while True:
+            # Wait for end ball event to be set
+            yield from self._end_ball_event.wait()
+            if self._end_ball_event.is_set():
```
jabdoa2: We do not need the if here because the previous line will only return once this is set

mpf/modes/game/code/game.py
```diff
@@ -31,16 +32,174 @@ def __init__(self, machine, config, name, path):
...
+                yield from self._end_ball()
+                self._end_ball_event.clear()
+
+    @asyncio.coroutine
+    def _start_game(self):
+        """Start a new game"""
```
jabdoa2: Comment should end with a period or pylint will complain

jabdoa2 (approved):
looks good to me

mpf/modes/game/code/game.py
qcapen: Nice catch.  That was left from when I had multiple events in that loop.

PR #921, opened 2017-05-07
extend blocking test

jabdoa2:


unRARed:
Per Slack discussion, I think it was determined that `block` is broken in `0.33.x` but working on `dev` and should be back-ported.

jabdoa2:
@unRARed did you test that? We do not have a test in 0.33 so it might be broken.

jabdoa2:
And i would say that it is working as well: https://github.com/missionpinball/mpf/pull/925

PR #925, opened 2017-05-10
test block in scoring

jabdoa2:


unRARed:
I fetched and reset to the latest `0.33.x` for both `mpf` and `mpf-mc` and still getting this test failure:

       File "/Users/unrared/Repos/pinball/combat/tests/test_bombs_dropped.py", line 33, in test_scoring
         self.assertEqual(current_score, self.machine.game.player.score)
     AssertionError: 36500 != 35500

Here is my mode test. You can see from the `tigger_mode()` helper that we assert the players score has changed by 5000 after hitting the drop target 5 times (1000 points in base). After the mode is triggered, when we hit the drop again, it this time scores 10,000 (but not 11,000 since the 1,000 from the base mode is blocked as expected). However, when we hit the switch `s_stationary_special`, this higher priority mode is ended while the base mode continues to ru. Hitting the drop target again at this point produces 0 points so it's as if the `block` is never lifted.

```python
 def test_scoring(self):
    # before mode
    current_score = 0
    self.begin_gameplay()
    self.hit_and_release_switch("s_stationary_special")
    current_score += 500
    self.assertEqual(current_score, self.machine.game.player.score)
    self.trigger_mode()
    current_score += 5000
    self.assertEqual(current_score, self.machine.game.player.score)
    # mode is now live
    self.hit_and_release_switch("s_drop_target")
    current_score += 10000 # 10x multi on drop targets during mode
    self.assertEqual(current_score, self.machine.game.player.score)
    self.hit_and_release_switch("s_stationary_special")
    current_score += 20000
    self.assertEqual(current_score, self.machine.game.player.score)
    self.assertModeNotRunning('bombs_dropped')
    self.assertModeRunning('base')
    print('mode no longer running, should NOT block')
    # hitting special ends mode
    # assert our return to base mode scoring
    self.hit_and_release_switch("s_drop_target")
    current_score += 1000
    self.assertEqual(current_score, self.machine.game.player.score)

 ## Helpers

 def begin_gameplay(self):
   self.hit_and_release_switch("s_start")
   self.advance_time_and_run(1)
   self.hit_and_release_switch("s_plunger_lane")
   self.advance_time_and_run(4) # 3+1
   self.assertEqual(1, self.machine.playfield.balls)

 def trigger_mode(self):
   for x in range(0, 5):
       self.hit_and_release_switch("s_drop_target")
   # the mech itself triggers this switch
   # as soon as all 5 targets have dropped
   self.hit_and_release_switch("s_drop_target_reset")
   self.assertModeRunning('bombs_dropped')
```

jabdoa2:
Can you try to add a advance_time_and_run() after hit_and_release_switch? The mode might be no longer active but not unloaded yet.

jabdoa2:
@unRARed Can you just extend this test? Your private test is not helping us here because nobody can reproduce it except you.

PR #928, opened 2017-05-11
Ball end on tilt

jabdoa2:


qcapen (approved):
Nice simple fix.  Good catch about missing player_turn_ending at the end of a game.

PR #932, opened 2017-05-12
Create failing test for scenario blocking multiple events scoring

unRARed:
Here is a test that reproduces the scenario I was describing on #925. It appears the bug might be in`clear_context()` found in `mpf/mpf/config_players/score_player.py` as it seems to only clear the tuple for the switch that triggered the end of the mode. In this case, we have two different score blocking events occur, so one of the tuples ( namely `(300, 'mode_for_logic_block', 's_counter_target_active')` ) remains in the `blocks` attribute of the game state. This prevents the scoring of both switches' lower priority modes.


jabdoa2:
superseeded by #933

PR #942, opened 2017-05-23
Fix multiple BCP clients crash

qcapen:


qcapen:
Problem was a lambda function in the Bcp._setup_bcp_connections function.  Since lambda functions bind variable names (not values like partial), the last connection setup in the loop would always be used as the client variable for all self.transport.register_transport callbacks.  Switching to a functools.partial for the callback fixed it.

Issue #939, opened 2017-05-18
clear machine vars from disk when they're deleted


toomanybrians:
Currently the method which removes machine vars only removes them from the running instance of MPF, but it does not remove them from the persistent disk file. For example, if you play a multiplayer game, you get the last game scores stored on disk, and then you start a single player game, it will write just player 1 to disk but players 2+ are also still on disk from before.

ashanhol:
Hi @toomanybrians / @jabdoa2! Is this issue still up for grabs? I'm new to OSS contribution and would like to take a stab at this if possible.

jabdoa2:
Hi @ashanhol. This issue is up for grabs! You can have a look at this test: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_MachineVariables.py to see how loading of machine vars works. It uses a mock DataManager (https://github.com/missionpinball/mpf/blob/dev/mpf/tests/TestDataManager.py) which would persist those variables to disk outside of a test.

ashanhol:
Hey @jabdoa2 + @toomanybrians. I looked through the tests and have been poking through the regular DataManager and code/game.py. I wanted to be clear on the issue, it seems that when the game ends you're [supposed to persist the variables](https://github.com/missionpinball/mpf/blob/dev/mpf/modes/game/code/game.py#L391-L393) to display previous scores in "attract mode". 
If that's the case, should I modify _start_game() to check and remove player variables on disk? 

Also, I wanted to clarify the documentation in game.py.  The [documentation for end_game()](https://github.com/missionpinball/mpf/blob/dev/mpf/modes/game/code/game.py#L423) says "this method calls :meth:`game_end`.". However, I don't see any method called game_end. Does it mean to say _end_game?

toomanybrians:
Hi @ashanhol, thanks for jumping in!

Looking at the code, I'm actually not sure this is an issue still. It looks like @jabdoa2 [added code to _end_game()](https://github.com/missionpinball/mpf/blame/dev/mpf/modes/game/code/game.py#L402-L404) that clears out any `playerX_score` machine vars that might have existed from a prior game that had more players than the current game. So maybe that just means we need a better test for this specific case?

You're correct about the documentation for `end_game()` being wrong. It should probably just mention that it posts those two events which will end the game, and not mention the call to `game_end()`

ashanhol:
I edited the documentation/comment in code/game.py and submitted a PR. 

[...]

PR #989, opened 2017-07-21
Edit documentation comment for end_game()

ashanhol:
Erased mention of game_end() as the method isn't actually calling that (as discussed in #939).


jabdoa2:
Thanks for taking care about this! Will make it easier for the next person touching that function.

PR #991, opened 2017-07-23
allow placeholders in event_player. fix #973

### LINKED ISSUE #973: Event name as placeholder in event_player
Replace placeholders in events in event_player. 

See #959 for string placeholders

jabdoa2:
dunno if this is such a great idea. syntax is crude


jabdoa2:
This code is suboptimal and needs a better string template implementation

Issue #972, opened 2017-07-01
Ball search failed action: end_ball


jabdoa2:
Implement `end_ball` as ball_search_failed_action in ball_search. It should just end the ball and not the game as `end_game`.

ashanhol:
Hi @jabdoa2! I was looking through ball_search, were you thinking of another if statement in  _compensate_lost_balls()?

jabdoa2:
Yes,  that would be the right place to implement this. Thanks for looking at this!

jabdoa2:
This was a feature request from one of our users: https://groups.google.com/forum/m/#!topic/mpf-users/F60DMIKll_g

ashanhol:
While this seems to be a quick fix #992, I found that the unit tests written for [ball_search](https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_BallSearch.py) seem a bit complicated. I'm not sure how to go about adding one. 

On a separate note, I think it might be a good idea to add additional links to the contributing page http://docs.missionpinball.org/en/latest/about/contributing_to_mpf.html as it doesn't list that this project requires python v 3.4 there.

jabdoa2:
thanks for implementing this @ashanhol

PR #992, opened 2017-07-25
Added end_ball to ball search

ashanhol:
Addressed #972, unit tests still run.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,5 +14,6 @@ their spare time, unpaid, for the love of pinball!
  * Brian Dominy <brian@oddchange.com>
  * Philip Dixon <epotech@ntlworld.com>
  * Thomas O'Connell <shastada@gmail.com>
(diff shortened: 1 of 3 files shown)


jabdoa2 (approved):
Hi Adina, this looks to the point. We can merge it as it is. Do you plan to write a test for this as well? A test would go here:  https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_BallSearch.py and could probably reuse the config as well (https://github.com/missionpinball/mpf/blob/dev/mpf/tests/machine_files/ball_search/config/config.yaml). You could set ball_search_failed_action at the beginning of the test using ```self.machine.playfields.playfield.config['ball_search_failed_action'] = "end_ball" ```.

ashanhol:
So I've never written a test in Python or like this before. I modeled it off the end_game ones and right now it's passing, but I'd feel better if someone looked it over.

Issue #993, opened 2017-07-26
PEP8 Style Warnings


colin-mullikin:
I came across your repo on YourFirstPR and, while looking around, noticed that you have quite a few (709) Style messages according to Landscape. I don't have a lot of Python experience, but I figure fixing these might be a good way to familiarize myself with the language and your code base.

Is this something that you guys are open to, @toomanybrians?

As there are so many and it could take quite a while, it might be a good idea to create a public branch so that any people working on it don't step on each other's toes/waste time by fixing the same issues.

jabdoa2:
Hi colin, 

We are totally up to that. Feel free to create PRs. Unfortunately, some of the messages are false positives. E.g. most (probably all) unused imports are caused by type annotations. Feel free to fix all the others. 


Jan

jabdoa2:
I looked into this a bit. You can install prospector (via pip3) locally and run it from within the MPF directory to get the same messages locally without pushing the code all the time. This also shows less false positives for me. You can also do PRs per file (or directory) if you want. We will merge that.

PR #994, opened 2017-07-28
Command and config_players folders.

colin-mullikin:
Issue #993. Fixed style warnings in command and config_players folders.

### DIFF
--- a/mpf/config_players/plugin_player.py
+++ b/mpf/config_players/plugin_player.py
@@ -53,16 +53,15 @@ class PluginPlayer(DeviceConfigPlayer):
 
         self.bcp_client = self._get_bcp_client(config)
 
-        m = self.machine
         for event in config:
-            event_name, _ = m.events.get_event_and_condition_from_string(event)
-            m.bcp.interface.add_registered_trigger_event_for_client(
+            event_name, _ = self.machine.events.get_event_and_condition_from_string(event)
+            self.machine.bcp.interface.add_registered_trigger_event_for_client(
                 self.bcp_client, event_name)
             event_list.append(event_name)
 
-        m.bcp.interface.add_registered_trigger_event_for_client(
+        self.machine.bcp.interface.add_registered_trigger_event_for_client(
             self.bcp_client, '{}_play'.format(self.show_section))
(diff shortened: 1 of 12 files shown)


mpf/config_players/plugin_player.py
```diff
@@ -50,30 +53,36 @@ def register_player_events(self, config, mode=None, priority=0):
 
         self.bcp_client = self._get_bcp_client(config)
 
+        m = self.machine
```
jabdoa2: I think this makes the code less readable. Can you revert this particular one?

mpf/config_players/score_player.py
```diff
@@ -40,15 +41,19 @@ def play(self, settings: dict, context: str, calling_context: str, priority: int
...
+        firstChar = priority_sorted[0]
+        return firstChar.priority > priority and firstChar.context != context
 
-    def _score(self, var: str, entry: dict, placeholder_parameters: dict) -> None:
+    def _score(self, var: str, entry: dict,
+               placeholder_parameters: dict) -> None:
+        mac = self.machine
         if entry['string']:
-            self.machine.game.player[var] = entry['string']
+            mac.game.player[var] = entry['string']
```
jabdoa2: Same here

jabdoa2 (commented):
Looks good to me. Only two minor issues.

mpf/config_players/plugin_player.py
colin-mullikin: The reason I did this was because of line length issues. My linter pointed out a ton of these, and sometimes this was the only way I could resolve it.

jabdoa2: We increased max-line-length to 120 in the landscape/prospector config (https://github.com/missionpinball/mpf/blob/dev/.landscape.yml). It should ist this one if you run prospector from the git root.

For this you could also break the lines into a multi-line statement instead.

colin-mullikin: Ah, okay. I'll make sure those settings get used. That should eliminate a lot of the messages on my side.

jabdoa2:
@colin-mullikin do you plan to make those changes? would like to merge the rest of the PR.

Issue #794, opened 2017-02-28
Service Menu - Items sorted in wrong order


MarkInc666:
Items in the service menu, for example switches, are sorted in alphabetical order.
This causes confusion.  They appear as 0, 1, 10, 11, 12, ...19, 2, 20, 21, 22, 23, ...
Should be 0, 1, 2, 3, 4, 5, 6, ...

This may be a problem for other lists too.

F4b1-:
Hi there, 
I picked this issue as a starting point for working on mission pinball. I might just need someone to point me in the right direction. I know there's a service menu I can access while a machine is running. (I am using demo-man right now) but this only contains the entries "settings", "coil test" and "switch test". No overview of all the switches or other items. If I run one of the dev examples I get a UI in the terminal that contains such a list, alphabetically ordered, that's is created by the TextUI class but that's not what you were referring to here, right?

MarkInc666:
If you go into service menu.  Then select coil test.
If you use the UP/DOWN keys to cycle through the drivers, they are neither in coil number order or alphabetical order.  I expect them to be listed as 0,1,2,3,...30,31,32,... etc.   
I think the   self.machine.service.get_coil_map() function (called around line 149 in service.py) needs to sort them properly.

F4b1-:
Hey, I opened a pull request a while back.  A couple of tests are failing. A lot of them don't really seem to be connected to the changes I made. I somehow get these test failures even I undo my changes and run them again. Any insight on this? Thanks

jabdoa2:
@F4b1- I did not look at the PR because of the failed tests. Sorry for that.

Looks like the test failures are caused by unrelated problems which existed at the time where you branched off dev. You could rebase to current dev. However, since this is a very small change it should be fine anyways. I added some comments to the PR itself.

RoseAndres:
Is this issue still open or has it been fixed with @F4b1-'s commit?

jabdoa2:
This is already fixed. We use natural sort on the strings now.

PR #1003, opened 2017-08-19
#794: Service Menu - Items sorted in wrong order

F4b1-:
Hey,
i believe the problem was that strings were sorted and therefore lexicographic sorting was used. So the coils were sorted by 1,2,10,3 for example. Taking into account the length of the objects gets rid of this problem. I haven't done a lot of python so far so feedback is always welcome. Adding a unit test to cover this issue should also be possible.

### DIFF
--- a/mpf/core/service_controller.py
+++ b/mpf/core/service_controller.py
@@ -75,4 +75,5 @@ class ServiceController(MpfController):
 
         # sort by board + driver number
         coil_map.sort(key=lambda x: x[0] + str(x[1].hw_driver.number))
+        coil_map.sort(key=lambda x: len(x[0] + str(x[1].hw_driver.number)))
         return coil_map
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,5 +15,6 @@ their spare time, unpaid, for the love of pinball!
  * Philip Dixon <epotech@ntlworld.com>
  * Thomas O'Connell <shastada@gmail.com>
  * Adina Shanholtz <ashanhol@gmail.com>
+ * Fabian Gand <gandfabian@gmail.com>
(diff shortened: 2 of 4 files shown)


F4b1-:
Huh, those test failures are pretty odd. Any advice on this?

mpf/core/service_controller.py
```diff
@@ -75,4 +75,5 @@ def get_coil_map(self) -> List[CoilMap]:
 
         # sort by board + driver number
         coil_map.sort(key=lambda x: x[0] + str(x[1].hw_driver.number))
+        coil_map.sort(key=lambda x: len(x[0] + str(x[1].hw_driver.number)))
```
jabdoa2, line 78: I guess you want to sort by len + the string itself, right? Currently the order of 000 and 111 would be random because both have a length of 3.

jabdoa2, line 78: Can you add a minimal test for this in test_ServiceMode.py? Just to make sure that it works for a basic example. Everything should be already set up in the test.

jabdoa2 (changes requested):

mpf/core/service_controller.py
F4b1-: Hey, I completed the coil test in in  test_ServiceMode.py.

F4b1-: Correct me if I am wrong but I wanted to sort by board/driver number(first) AND by length(second). This can be done (one after another) because python's sort is stable. As far as I can tell (also based on the tests) the order of 000 and 111 would not be random.

jabdoa2:
manually merged. thanks!

PR #1017, opened 2017-10-06
Added OPP support for processing switch matrices.  This includes adding

jabdoa2:
OPPMatrixCard and support for READ_MATRIX_INP command.  Small amounts of
code cleanup so processing is similar in all cases.  Insure msgs contain
enough bytes for crc8 before verifying crc8 is correct.  Removed use of
CFG_SOL_USE_SWITCH since using SET_SOL_INP_CMD.  Added config2.yaml for
some additional unit tests.


openpinballproject (approved):
Merge looks great.

PR #1045, opened 2017-11-22
Check duplicate lights on Spike by including node in address uniqueness

avanwinkle:
The recent update to check for duplicate light addresses triggered a false positive on Spike platforms, which use `device.node` plus `device.number` to identify a unique address. During the duplicate light check, a light with address `10-14` would fail as duplicating a light with address `8-14`.

This PR checks for the presence of `device.node` and includes the value when creating the unique key for a light, thereby ensuring that lights of the same number on different nodes are not considered duplicates.

### DIFF
--- a/mpf/devices/light.py
+++ b/mpf/devices/light.py
@@ -97,7 +97,8 @@ class Light(SystemWideDevice):
         check_set = set()
         for light in machine.lights:
             for driver in light.hw_drivers.values():
-                key = (light.platform, driver.number, type(driver))
+                # Spike platforms identify unique addresses by node + number, so include node if available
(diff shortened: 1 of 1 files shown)


jabdoa2:
Good find. I think i want to solve this differently by including node + index in number (and probably separate variables for index and node).

jabdoa2:
Thanks for reporting this and proposing a fix.  That is really great! I hope I did not discourage you from contributing to MPF by modifying the approach slightly. This makes it a bit more universal in a sense that all platforms should behave according to their interface (and Spike was incorrect here).

avanwinkle:
Awesome, thanks for checking it out! You're the expert, I trust your approaches over mine :)

Issue #1049, opened 2017-11-28
Missing HTML Colours


cfbenn:
Two colours are missing from rgb_color.py.

Aqua, #00FFFF / rgb(0, 255, 255)
Fuchsia, #FF00FF / rgb(255, 0, 255)

PR #1061, opened 2017-12-26
Update rgb_color.py

MatthewDietrich:
Add the two standard web colors that were missing from this file, as identified in Issue #1049


jabdoa2:
Thanks!

Issue #777, opened 2017-02-16
Add temp disable to autofire devices if they fire too fast/long (anti machine-gunning)


toomanybrians:
Autofire coils (and related devices based on autofires like in #636 like slingshots, pop bumpers, etc) should have a config option that will temporarily disable them if they fire too fast or too long. A config option could be something like "if it fires X times in Y secs then disable it for Z secs".

Since autofires are handled in hw rules, we'd have to watch for switch events which come on while the autofire is active and then temporarily disable the autofire.

jabdoa2:
We got that. I think it is called recycle time. Not all platforms support it though

toomanybrians:
I'm thinking this feature though can go longer than the 255ms that the platform(s) support. For example, in WoZ, it's possible that a ball can get "stuck" in the upper slingshot area for too long and not come down. So this setting would be more like "if this slingshot fires more than 10 times in 10 seconds, disable it for 5 secs so the ball will come down and the game will continue".

toomanybrians:
this is a real setting from woz btw, though I forget exactly how they phrase it

jabdoa2:
Makes sense. With the next platform refactoring we will probably support longer recycle ms in software but this will not help here because it actually counts the hits.

Aiden166:
Is this still open? I think I might know how to do it but I'll need to explore the codebase some more to figure out the specifics, such as where I would even put it.

jabdoa2:
Hi Aiden,

Yes this is still open. Let me know if you have any questions. 


Jan

Aiden166:
Okay. I'll get on this then. Thanks for responding so quickly!

Aiden166:
I'm sorry. This is much more complicated than I thought it would be. I will keep working on it, but it will probably take me a while.

Aiden166:
I think I've figured out how to do it, but I have two questions. One, is _hit method inside of the AutofireCoil class called whenever the device each instance represents activated, and two, would it be better to edit the class itself rather than use events? It seems to me that would be the better option as that way you don't have to deal with a separate event for each instance.

Aiden166:
@jabdoa2  Submitted a pull request. I'd be very grateful if you could review it. I am currently unable to perform any tests due to the limitations of the laptop I'm currently using, so this is completely untested. Also, know me, I probably did something incredibly stupid and broke the whole build. So, sorry.

Aiden166:
[...]

PR #1066, opened 2018-01-11
Autofire timeout

Aiden166:
More of a request to test than to merge. I am currently unable to do any tests myself. I was wondering if someone could look over this. It *should* work but I'm not confident in it. What it's supposed to do is add the ability for autofire devices to automatically disable themselves for X amount of seconds if they are triggered Y times in Z seconds. X, Y, and Z are options in the config file. There's also an option to disable the timeout entirely. Basically, when disabled it shouldn't change anything. Sorry for anything I did wrong. I'm new to GitHub and okay-ish at Python. Thanks in advance.

### DIFF
--- a/mpf/devices/autofire.py
+++ b/mpf/devices/autofire.py
@@ -47,6 +47,13 @@ class AutofireCoil(SystemWideDevice):
                 self.config['ball_search_order'], self._ball_search, self.name)
         # pulse is handled via rule but add a handler so that we take notice anyway
         self.config['switch'].add_handler(self._hit)
+        if self.config['enable_timeouts']:
+            self._timeout_watch_time = self.config['timeout_watch_time']
+            self._timeout_max_hits = self.config['timeout_max_hits']
+            self._timeout_disable_time = self.config['timeout_disable_time']
+            self._timeout_hits = []
+            from time import time as _time
+            from threading import Timer as _Timer
 
     @event_handler(10)
     def enable(self, **kwargs):
@@ -110,6 +117,17 @@ class AutofireCoil(SystemWideDevice):
         """Rule was triggered."""
         if not self._ball_search_in_progress:
             self.config['playfield'].mark_playfield_active_from_device_action()
+        if self.config['enable_timeouts']:
+            self._timeout_hits.append(_time())
+            while True:
+                if self._timeout_hits[-1]-self._timeout_hits[0]>self._timeout_watch_time:
+                    self._timeout_hits.pop(0)
+                else:
+                    break
+            if len(self._timeout_hits)>self._timeout_max_hits:
+                self.disable()
+                _timeout_timer=_Timer(self._timeout_disable_time,self.enable)
+                _timeout_timer.start()
 
     def _ball_search(self, phase, iteration):
         del phase
--- a/mpf/devices/autofire.py
+++ b/mpf/devices/autofire.py
@@ -48,9 +48,9 @@ class AutofireCoil(SystemWideDevice):
         # pulse is handled via rule but add a handler so that we take notice anyway
         self.config['switch'].add_handler(self._hit)
         if self.config['enable_timeouts']:
-            self._timeout_watch_time = self.config['timeout_watch_time']
+            self._timeout_watch_time = self.config['timeout_watch_time']/1000
             self._timeout_max_hits = self.config['timeout_max_hits']
-            self._timeout_disable_time = self.config['timeout_disable_time']
+            self._timeout_disable_time = self.config['timeout_disable_time']/1000
             self._timeout_hits = []
             from time import time as _time
             from threading import Timer as _Timer
--- a/mpf/devices/autofire.py
+++ b/mpf/devices/autofire.py
@@ -42,6 +42,10 @@ class AutofireCoil(SystemWideDevice):
         super().__init__(machine, name)
         self.delay = DelayManager(self.machine.delayRegistry)
         self._ball_search_in_progress = False
+        self._timeout_watch_time = None
+        self._timeout_max_hits = None
+        self._timeout_disable_time = None
+        self._timeout_hits = []
 
     def _initialize(self) -> None:
         if self.config['ball_search_order']:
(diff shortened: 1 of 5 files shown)


Aiden166:
Let's hope that fixes it.

Aiden166:
I guess not.

mpf/devices/autofire.py
```diff
@@ -47,6 +47,13 @@ def _initialize(self) -> None:
                 self.config['ball_search_order'], self._ball_search, self.name)
         # pulse is handled via rule but add a handler so that we take notice anyway
         self.config['switch'].add_handler(self._hit)
+        if self.config['enable_timeouts']:
+            self._timeout_watch_time = self.config['timeout_watch_time']/1000
+            self._timeout_max_hits = self.config['timeout_max_hits']
+            self._timeout_disable_time = self.config['timeout_disable_time']/1000
+            self._timeout_hits = []
+            from time import time as _time
```
jabdoa2: Can you move the import to the top of the file? Pylint complains about this.

mpf/devices/autofire.py
```diff
@@ -47,6 +47,13 @@ def _initialize(self) -> None:
...
+            from threading import Timer as _Timer
```
jabdoa2: This timer will interfere with asyncio. You can use self.machine.clock instead or instantiate DelayManager(self.machine.delayRegistry) for a nicer interface.

mpf/devices/autofire.py
```diff
@@ -110,6 +117,17 @@ def _hit(self):
...
+        if self.config['enable_timeouts']:
+            self._timeout_hits.append(_time())
+            while True:
+                if self._timeout_hits[-1]-self._timeout_hits[0]>self._timeout_watch_time:
+                    self._timeout_hits.pop(0)
+                else:
+                    break
+            if len(self._timeout_hits)>self._timeout_max_hits:
+                self.disable()
+                _timeout_timer=_Timer(self._timeout_disable_time,self.enable)
```
jabdoa2: I like the approach. There is a race if the pops are disabled during the disable period. With DelayManager you could just cancel the delay in the disable function.

jabdoa2 (dismissed):
Thanks for your efforts! The general approach is good. A test would be nice. You can just run:

pip3 install -e .

Within the mpf directory to get everything installed. Then you can run unit tests via:

python3 -m unittest discover -s mpf.tests

See: http://docs.missionpinball.org/en/dev/about/contributing_to_mpf.html (you can ignore all the mc related points if you only make changes to mpf).

jabdoa2 (commented):
Your test could go here: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_Autofire.py

Aiden166:
I fixed the import, changed the Timer to a delay, and fixed the race condition. I think. About the tests, I am currently working on a desktop at my school, which means I have to do everything in-browser as I have limited conditions. I have Python installed but it's 3.6 and the installation help says it will only work with 3.4. Additionally, I can't access the command line at all because it's been disabled. I am unable to do any tests at this time. I should be able to once I get home, but I need to find a way to be able to call Python 3.4 from the command line instead of 3.6. I've looked at the automated tests, but it seems to be unable to find the config options I added, so I need to figure out what I screwed up on that, which is what is giving me the most trouble right now. I will try to write some tests for it but I'll need to figure out how first. Thanks for all the help you've given me, and sorry to trouble you so much.

jabdoa2:
Python 3.6 should be fine for MPF. I think it currently is only an issue with Kivy on Windows.

Linter is currently unhappy on travis:

```
mpf/devices/autofire.py
  Line: 9
    pylint: wrong-import-order / standard import "from time import time as _time" should be placed before "from mpf.core.delays import DelayManager"
  Line: 53
    pep8: E226 / missing whitespace around arithmetic operator (col 73)
    pylint: attribute-defined-outside-init / Attribute '_timeout_watch_time' defined outside __init__ (col 12)
  Line: 54
    pylint: attribute-defined-outside-init / Attribute '_timeout_max_hits' defined outside __init__ (col 12)
  Line: 55
    pylint: attribute-defined-outside-init / Attribute '_timeout_disable_time' defined outside __init__ (col 12)
  Line: 56
    pylint: attribute-defined-outside-init / Attribute '_timeout_hits' defined outside __init__ (col 12)
  Line: 108
    pep8: W293 / blank line contains whitespace (col 1)
  Line: 125
    pep8: E226 / missing whitespace around arithmetic operator (col 42)
    pep8: E225 / missing whitespace around operator (col 64)
  Line: 129
    pep8: E225 / missing whitespace around operator (col 39)
  Line: 131
    pep8: E231 / missing whitespace after ',' (col 58)
```

Aiden166:
Fixed the style. Still have no idea why it can't find "enable_timeouts" in the config file.

Aiden166:
Okay, I have no idea why it can't find it. If someone could tell me why it's unable to find the option, I would be extremely grateful.

mpf/devices/autofire.py
```diff
@@ -40,6 +42,11 @@ def __init__(self, machine: "MachineController", name: str) -> None:
         super().__init__(machine, name)
         self.delay = DelayManager(self.machine.delayRegistry)
         self._ball_search_in_progress = False
+        if self.config['enable_timeouts']:
```
jabdoa2: The problem here is that config is set after __init__ is done. I guess all you have to do is move this to _initialize which is called after self.config is set.

Aiden166: It actually had been in _initialize before and I was still getting the same error. I actually moved it to __init__ because linter was complaining about the variable declarations being outside of __init__ in that error log you posted. I'll try moving it back into _initialize but unfortunately I don't think that's it. It's probably something stupid like I put the options in the wrong config file.

Aiden166:
So, why does that work now? I see that you removed the "enable_timeouts" config check and used the "timeout_watch_time" to decide whether to enable auto-timeout instead, but would you mind explaining to me why that works?

jabdoa2:
That is unrelated. 1. I copied the config to kickbacks which subclass autofires. That fixes the issue with missing config. 2. I added the variable to init to make the linter happy.

Aiden166:
I see. Thank you very much. So I needed to put the config options in "kickbacks" instead of "autofires". Thank you for explaining that to me. Sorry for being so much of a noob.

jabdoa2:
added a test and fixed some issues with the timeout. this should be enough to fix #777

Aiden166:
Should I delete the branch then now?

jabdoa2:
The branch is in your repository so it's up to you. However, there is little reason to keep it.

PR #1067, opened 2018-01-17
Conditional event handlers and asset groups

avanwinkle:
This PR adds functionality in `event_player` to support the use of conditions on event _handlers_ in addition to the triggers. For example:

```
reenable_nonrecruit_modes:
    - start_mode_shadowbroker_base
    - start_mode_n7_assignments
    # Prior to collectorship, enable overlord; afterwards, enable arrival
    - start_mode_overlordlight{device.achievements.collectorship.state!="complete"}
    - start_mode_arrival{device.achievements.collectorship.state=="complete"}
```

This update also includes conditional support for asset pools (e.g. `sound_pools`) that evaluate the inclusion of assets during random/sequential selection:
```
sound_pools:
  randomsquad_recruitcomplete:
    type: random
    track: voice
    sounds:
      - missioncomplete_voice_miranda
      - missioncomplete_voice_jack
      - missioncomplete_voice_garrus{current_player.status_garrus<=1}
      - missioncomplete_voice_jack{current_player.status_jack<=1}
      - missioncomplete_voice_grunt{current_player.status_grunt<=1}
      - missioncomplete_voice_mordin{current_player.status_mordin<=1}
      - missioncomplete_voice_samara{current_player.status_samara<=1}
 ```

### THE LOGIC
This PR adds a new method to `placeholder_manager` called `parse_conditional_template()`, which standardizes the parsing of a template string (e.g. _"some_event_name{condition!=False}|2"_) that was previously duplicated across players. This method returns a dictionary object as follows:
```
parse_dict = { 
    name: "some_event_name", 
    condition: <BoolTemplate 'condition!=False'>, 
    number: 2
}
```
The method supports an optional keyword argument `default_number`. If provided, the method will attempt to type-conform the template's number value to match the `default_number` type, or populate the default value if the template has no number specified. If not provided, the method will return the template's number value in its original string form (e.g. _"500ms"_)

### THE IMPLEMENTATION
The `parse_conditional_template` method is implemented in `EventPlayer` to determine whether each event handler in a list should be posted, and is done in realtime so that condition changes mid-mode are always reflected. Currently the event player parses the handler strings every time the trigger is activated, but future work might store the parsed strings (with names, conditions, and delays) the way asset pools do.

[...]

PR #1069, opened 2018-01-19
Support kwargs in event handler conditions

avanwinkle:
This PR is an update to https://github.com/missionpinball/mpf/pull/1067 that allows event kwargs to be used in the conditions of event handlers.

The original feature was built and tested using only machine/player variables, but this fix is needed to evaluate a condition based on the `Args` of the event that triggered the handler.

I know `delete kwargs` is commonly used to prevent propagation and distinguish between regular and relay events. I removed the deletion line for the `play` method but don't believe it should have any behavioral impacts, since there are no other calls that reference/relay it. Right?


jabdoa2:
Makes sense! Sorry i missed this for two days.

Issue #1075, opened 2018-01-23
mpf both crashes with termios.error: (19, 'Operation not supported by device')


muffler-aus:
When running mpf both I get an error: termios.error: (19, 'Operation not supported by device') If I run MPF and MC in seperate terminals it appears to work fine. After some discussion with fellow MPF users/contributors I was pointed at commands/both.py. 

Looking through this I found that I can reproduce the error with this minimal code:

```
import sys
import termios
import multiprocessing
from time import sleep


def _start():
     tbs = termios.tcgetattr(sys.stdin)
     sleep(10)     

mpf = multiprocessing.Process(target=_start)
mpf.start()
mpf.join()

When termios calls tcgetattr on sys.stdin inside the multiprocess it errors out.

I will submit a PR that works around this issue but I need to have it tested on other platforms (Linux/Windows).

jabdoa2:
fixed by #1076

PR #1076, opened 2018-01-23
Issue #1075: mpf both crashes with termios.error

muffler-aus:
Changing both.py to run MC in mutliprocessing and leave MPF running in the main thread should sidestep the issue with termios unable to access the tty stdin.


jabdoa2:
Awesome. Thanks! Will test it tonight on Linux

jabdoa2:
Works fine on Linux. I guess Windows will also be fine. Thanks!

PR #1080, opened 2018-01-29
Pypiwin32 workaround

jabdoa2:
lets give it a try


qcapen:
Looks good and works for me

PR #1082, opened 2018-01-30
Fix multiball_lock event unregister and duplicates

avanwinkle:
This PR fixes two minor bugs in `multiball_lock` behavior events.

1. When enabled, a multiball lock registers two event handlers `_lock_ball` and `_post_events`, but when the lock is disabled only `_lock_ball` is removed. This PR includes the removal of `_post_events` when unregistering.

2. When posting events, a multiball lock creates an array of events and then iterates posting them. This array is never cleared, so each new event posting (e.g. `total_balls=2` includes all previous events e.g. `total_balls=1`). This PR clears the array of events after they have been posted.


jabdoa2:
Thanks!

PR #1084, opened 2018-01-31
Restore "(shot_group)_(profile)_(state)_hit" event

avanwinkle:
This PR reverts one change from https://github.com/missionpinball/mpf/pull/1033, specifically [refactor Shots and ShotGroups](https://github.com/missionpinball/mpf/commit/c90b0d73f88375c069b87c30a9b01e228bc4bd5c).

The changes above simplified and streamlined the shot_group logic, but in the process eliminated the profile-specific events.

**Events posted before refactor:**
* `(shot_group)_hit Args={profile: profile, state: state}`
* `(shot_group)_(profile)_hit Args={profile: profile, state: state}`
* `(shot_group)_(profile)_(state)_hit Args={profile: profile, state: state}`

**Events posted after refactor:**
* `(shot_group)_hit Args={}`

**Use Case:**
The use case affected is a collection of shots in a group with one or more shots in various states. The shot group provides a means to rotate which of the shots are in which states, and when a player hits a shot in the group it is desired to have event handlers based on the state of the shot hit. Examples: a "chase" mode, an elimination mode, or a random shot mode.

Currently, an event handler must be attached for every shot in the group in order to capture events based on state. This PR restores one event to the shot group that includes the profile and state: `(shot_group)_(profile)_(state)`

I'm open to other ideas and suggestions on how to handle the above use cases, if shot_groups are not ideal or if passing arguments is preferred. Thanks!

### DIFF
--- a/mpf/devices/shot_group.py
+++ b/mpf/devices/shot_group.py
@@ -110,9 +110,11 @@ class ShotGroup(ModeDevice):
         """One of the member shots in this shot group was hit.
 
         Args:
-            kwarg: unused
+            kwarg: {
+                profile: the current profile of the member shot that was hit
+                state: the current state of the member shot that was hit
+            }
         """
-        del kwargs
         if advancing:
             self._check_for_complete()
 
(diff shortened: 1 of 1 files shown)


jabdoa2:
Makes sense. I think (profile) might not be needed anymore since there can be only one. That is probably why I removed it. Besides that I'm fine with the change.

avanwinkle:
Sure, I'm happy enough with `(shot_group)_(state)_hit` just as well! I'll make that change.

jabdoa2:
Thanks

PR #1100, opened 2018-02-20
Update fast communicator to delay longer

Blackbaud-JohnMarsh:
Update fast communicator to delay longer after a reset before trying to read the switches.


jabdoa2:
Did you test with 2s? I thought that would already be a long delay.

Issue #1085, opened 2018-01-31
Switch sequence with two times the same switch


jabdoa2:
Causes duplicate event warnings.

- Test
- Fix

jabdoa2:
fixed in #1103

PR #1103, opened 2018-02-22
1085 - Avoid duplicate handlers for shot sequence event duplicates

avanwinkle:
This PR addresses issue #1085 by restricting the creation of event handlers for `sequence_shots` events.

Previously, each item in the list of switches/events for a sequence shot was given an event handler. With this change, that list is reduced to a non-duplicate set prior to the event handler creation, ensuring only one handler is created per unique switch/event name.


jabdoa2:
Looks good. Thanks!

PR #1107, opened 2018-02-27
Common method for expecting config/show version tag

avanwinkle:
Recent updates to config loading/processing (mainly related to caching and consistent version number tagging) changed how config files were parsed.

These changes were implemented in **config_processor.py** in the `_load_config_file_and_return_loaded_files()` method but not in the `load_config_files()` method. The latter is a public method that can be used by other scripts to interpret MPF config files, so consistent behavior with MPF's internal processing is optimal.

This PR extracts the config/show version tag parser into a new method `get_expected_version()` and calls that method from both places where version tags are expected.

### DIFF
--- a/mpf/core/config_processor.py
+++ b/mpf/core/config_processor.py
@@ -123,12 +123,7 @@ class ConfigProcessor(object):
         """Load a config file and return loaded files."""
         # config_type is str 'machine' or 'mode', which specifies whether this
         # file being loaded is a machine config or a mode config file
-        if config_type in ("machine", "mode"):
-            expected_version_str = "#config_version={}".format(__config_version__)
-        elif config_type == "show":
-            expected_version_str = "#show_version={}".format(__show_version__)
-        else:
-            raise AssertionError("Invalid config_type {}".format(config_type))
+        expected_version_str = ConfigProcessor.get_expected_version(config_type)
 
         config = FileManager.load(filename, expected_version_str, True)
         subfiles = []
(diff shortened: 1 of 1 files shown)


jabdoa2:
That is a good one. Thanks for noticing! Are you running any external scripts which use that?

Just one small linter issue:
```
Messages
========
mpf/core/config_processor.py
  Line: 209
    pep257: D102 / Missing docstring in public method
```

avanwinkle:
Oh those pesky linter errors. I need to fix my local linter so it matches and I can stop checking in goofs! Fixed now :)

Yes to your question, I wrote a python script to traverse and maintain my audio files, copying them in from my source folder to the appropriate mode folder when added to a config file and moving/deleting audio files that are no longer in those modes. I'm constantly shuffling sounds around my configs :)

jabdoa2:
Try: pip3 install prospector. Then just run prospector from the root of the repository. 

You are welcome to upstream your scripts if they would help others.

avanwinkle:
Ah hah! I didn't see that .landscape.yml was there for the prospector settings, since it's hidden. Now I can run it with the same params as Travis, and I'll do my best to clean up before committing.

For some reason I'm still getting the `pylint` failure message with that old `TypeError: 'Uninferable' object is not iterable` when I run prospector. I guess if that's the only error I get locally, my code should be good to PR. Have you a way to suppress that pylint error?

jabdoa2:
Look good. Thank!

That pylint error should be fixed upstream. Try: pip3 install --upgrade prospector

PR #1108, opened 2018-02-27
Allow multiple "eject_coil_enable_time" values

avanwinkle:
In the continuing work for supporting ball devices that physically lock multiple balls, this PR extends the `eject_coil_enable_time` setting to (optionally) support multiple time values.

The reasoning here is that eject-coil-enable devices, by definition, rely on gravity to pull the ejecting ball from the device. When multiple balls are stacked in the device it is critical to time the coil to release only one ball: too short and the ejecting ball will jam, too long and a second ball will attempt to escape.

However the exact timing of the coil depends on how many balls are in the device, as the added weight of a second or third ball will "push" the ejecting ball out faster. As a result, an eject-coil-enable ball device will need shorter and shorter enable times as the number of held balls increases.

This change only affects machines when explicitly specified. Config settings that have only a single `eject_coil_enable_time` value will continue to behave as they currently do.

### DIFF
--- a/mpf/core/config_spec.py
+++ b/mpf/core/config_spec.py
@@ -118,7 +118,7 @@ ball_devices:
     eject_coil_retry_pulse: single|ms|None
     eject_coil_reorder_pulse: single|ms|None
     eject_coil_max_wait_ms: single|ms|200ms
-    eject_coil_enable_time: single|ms|None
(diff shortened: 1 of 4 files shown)


jabdoa2:
Always wanted to implement that. Thanks for doing that!

PR #1113, opened 2018-03-03
Issue 1112: Allow setting x,y,z location to switch

muffler-aus:
I would like to set the location (copy from mpg-monitor) the location of switches. This can then be used in code to react to switch events and drive lightshows based on switch location.

Currently I get "Your config contains a value for the setting "switches:s_start:x", but this is not a valid setting name."

I will raise a PR that will allow this setting in switches config.


jabdoa2:
Thanks

PR #1117, opened 2018-03-09
Python 35 36 support

qcapen:
Includes new pypinproc binaries for Windows that support python 3.5 and 3.6 (both 32 and 64 bit)


jabdoa2:
Awesome!

jabdoa2 (approved):

jabdoa2:
The test failure was caused by my changes (and it fixed in dev). This should be good to be merged.

Issue #1120, opened 2018-03-12
Relative index names for LightGroup lights


avanwinkle:
The behavior for a `LightGroup` device (i.e. a `LightStrip(e)` or `LightRing`) is to create an array of light devices with sequential names.

Currently, these names are based on the (zero-index) **absolute** address of the light in the LED chain, for example the third light on a group starting at 0 is named _group_name_light_2_ and the third light on a group starting at 64 is _group_name_light_66_.

Unfortunately this prevents the use of show tokens to play shows on light groups at different places in the chain, which could readily be done if the light names were based on the **relative** address of the light within its group. For example, the third light on a group starting at 64 would be _group_name_light_2_ instead of _group_name_light_66_.

In that case, group-based light shows like the following would be supported:
```
#show_version=5
- time: 0
  lights:
    (led_group)_light_1: red
    (led_group)_light_2: orange
    (led_group)_light_3: yellow
    (led_group)_light_4: green
    (led_group)_light_5: blue
    (led_group)_light_6: indigo
    (led_group)_light_7: purple
    (led_group)_light_0: pink
- time: '+1'
  lights:
    (led_group)_light_2: red
    (led_group)_light_3: orange
    (led_group)_light_4: yellow
    (led_group)_light_5: green
    (led_group)_light_6: blue
    (led_group)_light_7: indigo
    (led_group)_light_0: purple
    (led_group)_light_1: pink
- time: '+1'
  lights:
    (led_group)_light_3: red
    (led_group)_light_4: orange
    (led_group)_light_5: yellow
    (led_group)_light_6: green
    (led_group)_light_7: blue
    (led_group)_light_0: indigo
    (led_group)_light_1: purple
    (led_group)_light_2: pink
- time: '+1'
  lights:
    (led_group)_light_4: red
    (led_group)_light_5: orange
    (led_group)_light_6: yellow
    (led_group)_light_7: green
    (led_group)_light_0: blue
    (led_group)_light_1: indigo
    (led_group)_light_2: purple
    (led_group)_light_3: pink
```
This is a quick change that I'm happy to make a PR for, if there's support for it.  It's potentially breaking for existing configurations, but I'm not sure how much support for LightGroups expected given their newness.

jabdoa2:
Sounds reasonable. Go for it!

jabdoa2:
fixed by #1121

PR #1121, opened 2018-03-13
Relative indexes for LightGroup light names

avanwinkle:
This PR addresses #1120 by creating lights for LightGroups using the light's relative index in the group, rather than absolute address index.


jabdoa2:
awesome! thanks!

Issue #1086, opened 2018-02-02
Support Servo Speed and Acceleration


jabdoa2:
Expose speed and acceleration in servo config.
Pololu maestro supports that in hardware. Support it in software for other platforms.

F4b1-:
Hi, I will take a look at this as my second issue. After reading the mpf documentation on the pololu mastro and looking at the code a bit it seem that the set_acceleration and set_speed methods are already implemented. So this is issue is just about parsing the speed/acceleration values from the config file and passing them on? Thanks!

jabdoa2:
Yeah right. This issue is about adding it to config in a platform independent way (normalized values e.g. 0 to 1). Pololu supports it already. For other platforms we should emulate that but that part could also become a second issue.

PR #1122, opened 2018-03-14
#1086 Support Servo Speed and Acceleration

F4b1-:
Hi, there are some things I am a bit unsure about:
* the general implementation: Since it's only implemented in the Maestro the set_speed/acceleration methods are currently not implemented for the other platforms. As discussed in #1086 this should be emulated for the other platforms (new issue)
* the default values for speed and acceleration (both are 0 currently)

Just let me know if any of this or something else needs fixing.

### DIFF
--- a/mpf/core/config_spec.py
+++ b/mpf/core/config_spec.py
@@ -974,6 +974,8 @@ servos:
     servo_max: single|float|1.0
     ball_search_min: single|float|0.0
     ball_search_max: single|float|1.0
+    speed: single|float|0.0
+    acceleration: single|float|0.0
     ball_search_wait: single|ms|5s
     include_in_ball_search: single|bool|True
     reset_position: single|float|0.5
--- a/mpf/core/config_spec.py
+++ b/mpf/core/config_spec.py
@@ -974,8 +974,8 @@ servos:
     servo_max: single|float|1.0
     ball_search_min: single|float|0.0
     ball_search_max: single|float|1.0
-    speed: single|float|0.0
-    acceleration: single|float|0.0
+    speed_limit: single|float|-1.0
+    acceleration_limit: single|float|-1.0
     ball_search_wait: single|ms|5s
     include_in_ball_search: single|bool|True
     reset_position: single|float|0.5
--- a/mpf/devices/servo.py
+++ b/mpf/devices/servo.py
@@ -24,6 +24,8 @@ class Servo(SystemWideDevice):
         self.hw_servo = None
         self.platform = None        # type: ServoPlatform
         self._position = None
+        self.speed = None
+        self.acceleration = None
         self._ball_search_started = False
         self.delay = DelayManager(machine.delayRegistry)
         super().__init__(machine, name)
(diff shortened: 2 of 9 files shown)


jabdoa2:
What is the scale of acceleration and speed? 0 does not make much sense. Maybe this should be speed limit and acceleration limit? Default would be infinite I guess

F4b1-:
According to the documentation and the comments in the methods a value of 0 means unrestricted.

jabdoa2:
Yes for pololu. However, our interface should be easy to understand.

jabdoa2:
The linter is a bit unhappy:

```
Messages
========
mpf/platforms/fast/fast_servo.py
  Line: 30
    pep8: N804 / first argument of a classmethod should be named 'cls' (col 9)
  Line: 35
    pep8: N804 / first argument of a classmethod should be named 'cls' (col 9)
mpf/platforms/i2c_servo_controller.py
  Line: 106
    pep8: N804 / first argument of a classmethod should be named 'cls' (col 9)
  Line: 111
    pep8: N804 / first argument of a classmethod should be named 'cls' (col 9)
mpf/platforms/rpi/rpi.py
  Line: 91
    pep8: N804 / first argument of a classmethod should be named 'cls' (col 9)
  Line: 96
    pep8: N804 / first argument of a classmethod should be named 'cls' (col 9)
mpf/platforms/virtual.py
  Line: 357
    pep8: N804 / first argument of a classmethod should be named 'cls' (col 9)
  Line: 362
    pep8: N804 / first argument of a classmethod should be named 'cls' (col 9)
Check Information
=================
         Started: 2018-03-14 17:19:41.934344
        Finished: 2018-03-14 17:21:14.234246
      Time Taken: 92.30 seconds
       Formatter: grouped
        Profiles: .landscape.yml, doc_warnings, strictness_veryhigh, no_member_warnings, no_test_warnings
      Strictness: from profile
  Libraries Used: 
       Tools Run: dodgy, mccabe, pep257, pep8, profile-validator, pyflakes, pylint
  Messages Found: 8
```

What do think about just renaming those to max_speed? or speed_limit? With unit max position_change/s [1/s] (because position is just 0-1 and not degree). We would have to look up how that translates to the pololu (in the manual). They probably state something about it. Same for acceleration limit with unit max_speed_change/s and because speed_change is [1/s] this will be [1/s^2] (as any other acceleration).  What do you think?

F4b1-:
Thanks for your feedback. Renaming sounds good. About the errors/warnings in the travis builds: Travis basically complained about the unimplemented set_speed/acceleration methods in the other platforms. (_Method could be a function_) Any suggestions on what's the best way to handle this?

From the meastro docs: _The speed limit is given in units of (0.25 μs)/(10 ms)_
Example: Value is 140 -> 0.25 * 140 / 10 = 3.5 microseconds per second
1.0 in our config equals 9000 microseconds for the maestro (correct?):
So we got a value in our config of let's say: 0.5 (/s)
That is: 0.5 * 9000 / s = 4500/s
4500/100 = 45/10ms
4.5 / 0.25 = 18 -> value we send to the meastro


For acceleration: _The acceleration limit is a value from 0 to 255 in units of (0.25 μs)/(10 ms)/(80 ms)_
So we would basically do the same thing as before just for the adjusted acceleration formulas.

Pretty new to most of this so let me know what you think;-)

jabdoa2:
Sounds good. Go for it!

The linter only complains that the first parameter should be cls (because the method is in a class).

jabdoa2:
Looks good. Only some small linter issues:
```

mpf/platforms/pololu_maestro.py
  Line: 5
    pylint: wrong-import-order / standard import "import math" should be placed before "import serial"
  Line: 158
    pylint: no-self-use / Method could be a function (col 4)
  Line: 159
    pep8: E501 / line too long (127 > 120 characters) (col 121)
```

jabdoa2:
Perfect! Thanks! Dunno why the other linter errors suddenly appear. Will check that later.

Issue #792, opened 2017-02-24
Add custom "code:" section to machine configs


toomanybrians:
This is to replace "scriptlets" (which is a dumb name). I guess there's no real need to get rid of scriptlets, since they don't hurt anything, but we should add a `code:` section to the machine config, users add their custom code in via `module.Class`, and put their modules in their machine folder's `code` folder and we call their constructor and just pass the machine.

PR #1148, opened 2018-04-24
rename scriptlets to custom_code. fix #931

### LINKED ISSUE #931: Load scriptlets from any folder, change to "code"
Currently scriptlets need to be loaded from a machine's "scriptlets" folder. I want to change it so they can be loaded from anywhere. Rationale is the term "scriptlet" is sort of weird and suggests that custom code is some kind of hack or script and not a "real" MPF thing.

In the config file, I want to change the `scriptlets:` section to `code:`, and then it can be a list of
modules (from the machine root) and classes or methods. e.g.

```
code:
  code.my_custom_code.Foo
  code.some_other_code.some_func
```

The first would be the `Foo` class in the machine folder's `code/my_custom_code.py` module, and the second would be the `some_func` function in `code/some_other_code.py`

I'm thinking maybe we can inspect to see if the link is to a class or method, and if it's a class we create an instance of it, and if it's a method, we call it? In both cases we can pass the machine reference.

We can also save a reference to whatever is returned in a custom_code dict or something (just to keep it alive), and of course the person writing the code can do whatever they want if they want to refer to it differently.

jabdoa2:


jabdoa2:
also fixed #792

PR #1160, opened 2018-05-01
call related handlers first

jabdoa2:


qcapen (approved):
Looks good.  This fixes some race conditions?  So events that are created in handlers are processed first (prior to calling the next handler), regardless of priority. Does that cause any other strange side effects or design considerations?

jabdoa2:
I always thought it would work that way but it did not. Does not cause any races with ball devices. Without this its hard to design logic using logic blocks because two concurrent switch hits will trigger logic concurrently before any disable_events from the other block (see my test)

Issue #1177, opened 2018-05-15
Add configuration option to log in JSON format 


muffler-aus:
It would be nice to have JSON logging to enable better search filtering using tools like Splunk/ElasticSearch. I have created a patch that will give the -j command line option to enable JSON logging. Im not sure if this is the way you would like to have this option exposed but I guess with this issue and a PR we can start a conversation about how you would like it done.

jabdoa2:
contained in 0.50.9

jabdoa2:
this has been merged

PR #1178, opened 2018-05-15
Issue1177 - Added support for JSON logging via --json-logging command line option

muffler-aus:
Initial PR to allow JSON logging as per https://github.com/missionpinball/mpf/issues/1177


jabdoa2:
awesome! thanks!

PR #1183, opened 2018-05-29
Allow list of flashers as light value

avanwinkle:
This PR adds functionality to `flasher_player` to support a list of lights as the target. The iteration was copied directly from `light_player` and brings the two players to parity.

Without this fix, using a list of lights in `flasher_player` (or as show tokens for it) throws a KeyError. This is presumed to be a bug, since the expectation is that `flasher_player` will behave like other players.


jabdoa2:
Good one. Thanks!

PR #1199, opened 2018-07-09
Support for ‘—no-sound’ argument (required for corresponding MC update)

avanwinkle:
This is the sibling PR for https://github.com/missionpinball/mpf-mc/pull/340 to allow `--no-sound` as a command line argument.

This exists because of the dependency on arguments when using `mpf both`, but only MC actually cares about the new arg.


jabdoa2:
+1

PR #1211, opened 2018-08-21
Bug fix: prevent achievements from enabling after restoring state

avanwinkle:
This PR addresses a bug in Achievement behavior wherein achievements without `enable_events` inherit behavior from `mode_device.add_control_events_in_mode()` and automatically enable themselves when their modes start.

This inherited behavior is contradictory to the achievement's `device_loaded_in_mode()` behavior, which to calls `_reset()` or `_restore_state()` and uses sophisticated logic to determine the desired state of the achievement (including automatically enabling achievements without `enable_events` defined).

Unfortunately `add_control_events_in_mode()` creates a callback which is called _after_ `device_loaded_in_mode()`, so any achievement without `enable_events` will subsequently have its restored state changed to "enabled".

This PR overwrites the `add_control_events_in_mode()` behavior for Achievements to do nothing. The expected result is that `_restore_state()` will handle all of the desired behavior.

### DIFF
--- a/mpf/devices/achievement.py
+++ b/mpf/devices/achievement.py
@@ -220,6 +220,15 @@ class Achievement(ModeDevice):
             self._show.stop()
             self._show = None
 
+    def add_control_events_in_mode(self, mode: Mode) -> None:
+        """Override the default mode device behavior of automatically enabling devices
+        that don't explicitly define enable_events.
+
+        Achievements use sophisticated logic to handle their mode-starting states
+        during device_loaded_in_mode(). Therefore no default enabling is required.
+        """
+        pass
+
     def add_to_group(self, group):
         """Add this achievement to an achievement group.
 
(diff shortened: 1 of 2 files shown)


jabdoa2:
Prospector is unhappy:

```
Messages
========
mpf/devices/achievement.py
  Line: 223
    pep257: D205 / 1 blank line required between summary line and description (found 0)
    pep257: D400 / First line should end with a period (not 's')
Check Information
=================
         Started: 2018-08-21 23:15:29.046836
        Finished: 2018-08-21 23:17:15.629410
      Time Taken: 106.58 seconds
       Formatter: grouped
        Profiles: .prospector.yaml, doc_warnings, strictness_veryhigh, no_member_warnings, no_test_warnings
      Strictness: from profile
  Libraries Used: 
       Tools Run: dodgy, mccabe, pep257, pep8, profile-validator, pyflakes, pylint
  Messages Found: 2
```

Can you fix that? Should we add a test for this case?

avanwinkle:
Yes, and I'll get some tests up too!

jabdoa2:
Awesome! That looks really good

PR #1212, opened 2018-08-22
Support placeholder values and value types for event player args

avanwinkle:
This PR adds support for placeholder values in `event_player` events that use arguments.

```
event_player:
  some_event_occurred:
    # Current events can only pass static args
    event_with_basic_args:
      string_arg: "Hey there"
      int_arg: 120
    # New feature: pass dynamic (placeholder value) args
    show_pophits_count:
      count:
        value: device.counters.pophits.value
        type: int
    show_player_phase:
      phase:
        value: current_player.active_phase
        type: string
```

It supports types of `int`, `float`, `bool`, and `string` (which is default if no type supplied).

The motivation for this new functionality is to pass device and player values to MC event handlers. For example, having a widget update on a `logicblock_(name)_updated` works great except for when the slide first activates and there's no update event. 

Here's the actual example from my code:
```
event_player:
  logicblock_n7_hits_counter_complete: start_mode_n7_assignments
  # Force an event to show n7/iff count when the first bumper hits
  slide_minerals_slide_active:
    logicblock_n7_hits_counter_hit{device.achievements.normandyattack.state!="enabled"}:
      count:
        value: device.counters.n7_hits_counter.value
        type: int

widget_player:
  logicblock_n7_hits_counter_hit:
    n7_progress_widget:
      action: update
      slide: minerals_slide

widgets:
  n7_progress_widget:
    - type: text
      text: "N7 assignment in (count)"
      style: dmd_small
```

The first draft just checked for whether the argument value was a string and always created a `TextTemplate`, which was sufficient for use in slides/widgets. The first test I wrote expected an integer, which is when I realized it might be more useful to explicitly define what type of value to templatize. This required the argument value to become a dict, but that also makes it easier for `event_player` to know when it's supposed to try creating a placeholder.

Thoughts/feedback welcome!

### DIFF
--- a/mpf/config_players/event_player.py
+++ b/mpf/config_players/event_player.py
@@ -40,7 +40,21 @@ class EventPlayer(FlatConfigPlayer):
 
     def _post_event(self, event, s):
         event_name_placeholder = TextTemplate(self.machine, event.replace("(", "{").replace(")", "}"))
+        for key, param in s.items():
+            if isinstance(param, dict):
+                s[key] = self._evaluate_event_param(param)
         self.machine.events.post(event_name_placeholder.evaluate({}), **s)
+
+    def _evaluate_event_param(self, param):
+        if param.get("type") == "float":
+            placeholder = self.machine.placeholder_manager.build_float_template(param["value"])
+        elif param.get("type") == "int":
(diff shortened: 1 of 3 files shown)


jabdoa2:
Very cool. I like it. We could also use the new string placeholders like in segment displays. But that can also be added later

PR #1214, opened 2018-08-24
Bugfix: timer "pause" delay passed in secs when expecting ms

avanwinkle:
This PR fixes a bug where the `Timer.pause()` method handles pause duration in seconds but `DelayManager.add()` expects an argument in milliseconds. The result is that pauses are uselessly brief.

This PR takes the timer argument and multiplies by 1000 before passing the delay. Verified working via log event timestamps.

### DIFF
--- a/mpf/devices/timer.py
+++ b/mpf/devices/timer.py
@@ -299,7 +299,7 @@ class Timer(ModeDevice):
 
         self.running = False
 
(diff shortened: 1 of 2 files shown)


jabdoa2:
Looks like we got a test which relies on the wrong behavior

PR #1216, opened 2018-09-10
Ball Device: Entrance switch ignore window

avanwinkle:
This PR adds a new configuration setting to ball devices: `entrance_switch_ignore_window_ms`

This new setting replicates the `ignore_window_ms` behavior of switches, and is proposed because of limitations on that setting. The switch ignore window only applies to the switch posting events, whereas ball devices bind directly to the switch trigger itself. As a result, a ball device will count an entrance on every switch hit regardless of any debounce or ignore setting.

The use case for the new `entrance_switch_ignore_window_ms` setting is a ball device with an upfield switch that captures a ball rolling back down. The ball may hit the switch twice, once on the way up and again on the way down, but the device should only register a single entrance. If the switch is more than a balls-width upfield, it cannot be used as a `ball_switches` switch and must therefore be an `entrance_switch`.

This PR also fixes (?) the behavior of `autofire` coils, wherein autofire coil hits currently post *playfield_active* events even when they are disabled. This causes unexpected behavior because a given switch may alternate between being an autofire trigger and a ball device entrance. By design, ball device entrance switches cannot be `playfield_active` because a ball in a device is not on the playfield. Disabling the autofire to enable the ball device **should** prevent *playfield_active* events from posting, but in the current behavior they do not (and therefore cause ball count issues and *unexpected_ball_on_playfield*.

Tests and documentation updates to follow

### DIFF
--- a/mpf/devices/autofire.py
+++ b/mpf/devices/autofire.py
@@ -126,7 +126,10 @@ class AutofireCoil(SystemWideDevice):
 
     def _hit(self):
         """Rule was triggered."""
+        if not self._enabled:
+            return
         if not self._ball_search_in_progress:
+            self.machine.log.info("Autofire {} setting playfield_active event".format(self.name))
             self.config['playfield'].mark_playfield_active_from_device_action()
         if self._timeout_watch_time:
             current_time = self.machine.clock.get_time()
@@ -136,6 +139,9 @@ class AutofireCoil(SystemWideDevice):
             if len(self._timeout_hits) >= self._timeout_max_hits:
                 self.disable()
                 self.delay.add(self._timeout_disable_time, self.enable, "_timeout_enable_delay")
+
+        for event in self.config.get('events_when_fired', []):
+            self.machine.events.post(event)
 
     def _ball_search(self, phase, iteration):
         del phase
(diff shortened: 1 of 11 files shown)


mpf/devices/autofire.py
```diff
@@ -126,6 +126,8 @@ def disable(self, **kwargs):
 
     def _hit(self):
         """Rule was triggered."""
+        if not self._enabled:
```
jabdoa2, line 129: Good question what the expected behavior should be. In some cases (such as with kickbacks in an outlime) it would make sense to mark the playfield as active. Nevertheless it would also not hurt to not mark it active. I guess this is ok.

jabdoa2 (approved):

mpf/devices/autofire.py
avanwinkle: Yeah, I wasn't sure if this would break any expectations. My assumption is that _most_ kickbacks would be on switches that explicitly defined `tags: playfield_active` so the autofire-playfield-active event is superfluous. But maybe I've just been wrong in tagging my outlanes as `playfield_active`?

And in any case, if the kickback is enabled and actually kicks the ball back, the *playfield_active* event will still post. If the kickback is disabled, presumably the outlane switch would send the ball to the drain so it's not important.

jabdoa2:
Feel free to merge this yourself. Looks perfect to me.

PR #1220, opened 2018-09-15
Opp bus improvements

jabdoa2:
- do not start bus early to prevent user confusion
- batch incand updates


avanwinkle (approved):
Seems clean and straightforward to me!

PR #1228, opened 2018-10-02
Config spec support for multiple widget styles

avanwinkle:
This PR updates `config_spec` to process a widget's `styles:` setting as a list instead of a string. This is the companion commit to the MPF-MC update to support multiple widget styles.

See the MPF-MC change here: https://github.com/missionpinball/mpf-mc/pull/349


jabdoa2:
Good idea!

PR #1229, opened 2018-10-03
validate that ball_count is reasonable in multiball

jabdoa2:
Prevent ball_count <= 1 for ball_count_type total and ball_count <= 0 for ball_count_type add


avanwinkle (approved):
Coolcoolcool

PR #1243, opened 2018-10-18
Create mc.widget_player.rst

cloudjor:
Description on the mc.widget_player  and how to use it python code.
I hope this helps people with writing code for when they really have to


jabdoa2:
Thanks!

PR #1272, opened 2018-12-21
Rebuild windows pinproc libraries

qcapen:
Includes bugfix in pypinproc.cpp to fix DMD output.  I don't have hardware to test with so these are untested.


qcapen:
All these will work with 0.50, 0.51, and dev.

jabdoa2:
awesome. thanks :-)

PR #1285, opened 2019-01-04
disable coils when removing rules in OPP. fix #1284

### LINKED ISSUE #1284: Coils are sometimes stuck on in OPP when they are pulsed when we remove the rule
Fix this by disabling the coil explicitly

jabdoa2:


jabdoa2:
fixed in OPP firmware

PR #1295, opened 2019-01-13
Update smart_matrix_dmd_teensy_code.ino

aaronmatthies:
Added changes to accept new cookie and V4 of Teensy shield. The old cookie can cause issues with GIFs that use very dark colors.


jabdoa2:
perfect. thanks!

Issue #1257, opened 2018-11-24
Add setting to change high score order


jabdoa2:
Currently, more is always better in highscore. Allow categories were lower is better (e.g. faster/less time).

Change this sort based on a config setting: https://github.com/missionpinball/mpf/blob/dev/mpf/modes/high_score/code/high_score.py#L151

yensho:
Hi, I found this repo this afternoon, and would like to try contributing. Would it be alright if I took this issue?

jabdoa2:
Hi yensho,

Contributions are welcome! There is some documentation about the process here: http://docs.missionpinball.org/en/dev/about/contributing_to_mpf.html. To get familiar with MPF try the tutorial here: http://docs.missionpinball.org/en/dev/tutorial/index.html.

Jan

jabdoa2:
this has been solved in #1296

PR #1296, opened 2019-01-17
High score new config

yensho:
#1257 
Added new config value in config spec, and referenced that new value within high score. All tests run through unittest locally passed. No new unit test was added, though. Is that sufficient for testing these changes, or should I build my own local machine for a full test?

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -582,7 +582,7 @@ hardware_sound_player:
 high_score:
     __valid_in__: mode
     award_slide_display_time: single|ms|4s
-    high_score_reverse_sort: single|bool|False
+    reverse_sort: single|bool|False
     categories: omap|str:list|
     defaults: dict|str:list|None
     enter_initials_timeout: single|secs|20s
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -582,10 +582,10 @@ hardware_sound_player:
 high_score:
     __valid_in__: mode
     award_slide_display_time: single|ms|4s
-    reverse_sort: single|bool|False
     categories: omap|str:list|
     defaults: dict|str:list|None
     enter_initials_timeout: single|secs|20s
+    reverse_sort: single|bool|False
 info_lights:
     __valid_in__: machine                            # todo add to validator
 image_pools:
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -582,6 +582,7 @@ hardware_sound_player:
 high_score:
     __valid_in__: mode
     award_slide_display_time: single|ms|4s
+    high_score_reverse_sort: single|bool|False
     categories: omap|str:list|
     defaults: dict|str:list|None
     enter_initials_timeout: single|secs|20s
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -585,7 +585,7 @@ high_score:
     categories: omap|str:list|
     defaults: dict|str:list|None
     enter_initials_timeout: single|secs|20s
-    reverse_sort: single|bool|False
+    reverse_sort: list|str|
 info_lights:
     __valid_in__: machine                            # todo add to validator
 image_pools:
--- a/mpf/modes/high_score/code/high_score.py
+++ b/mpf/modes/high_score/code/high_score.py
@@ -148,7 +148,7 @@ class HighScore(AsyncMode):
                     new_list.append((player, player[category_name]))
 
             # sort if from highest to lowest
-            new_list.sort(key=lambda x: x[1], reverse=(not self.high_score_config['reverse_sort']))
+            new_list.sort(key=lambda x: x[1], reverse=(not category_name in self.high_score_config['reverse_sort']))
 
             # scan through and see if any of our players are in this list
             i = 0
--- a/mpf/modes/high_score/code/high_score.py
+++ b/mpf/modes/high_score/code/high_score.py
@@ -148,7 +148,7 @@ class HighScore(AsyncMode):
                     new_list.append((player, player[category_name]))
 
             # sort if from highest to lowest
-            new_list.sort(key=lambda x: x[1], reverse=True)
+            new_list.sort(key=lambda x: x[1], reverse=(not self.high_score_config['high_score_reverse_sort']))
 
             # scan through and see if any of our players are in this list
             i = 0
--- a/mpf/modes/high_score/code/high_score.py
+++ b/mpf/modes/high_score/code/high_score.py
@@ -148,7 +148,7 @@ class HighScore(AsyncMode):
                     new_list.append((player, player[category_name]))
 
             # sort if from highest to lowest
-            new_list.sort(key=lambda x: x[1], reverse=(not self.high_score_config['high_score_reverse_sort']))
+            new_list.sort(key=lambda x: x[1], reverse=(not self.high_score_config['reverse_sort']))
 
             # scan through and see if any of our players are in this list
             i = 0
--- a/mpf/tests/machine_files/high_score/modes/high_score/config/high_score.yaml
+++ b/mpf/tests/machine_files/high_score/modes/high_score/config/high_score.yaml
@@ -19,3 +19,4 @@ high_score:
       - MPF: 23
     loops:
       - JK: 42
+  reverse_sort: False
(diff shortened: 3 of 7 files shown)


mpf/config_spec.yaml
```diff
@@ -582,6 +582,7 @@ hardware_sound_player:
 high_score:
     __valid_in__: mode
     award_slide_display_time: single|ms|4s
+    high_score_reverse_sort: single|bool|False
     categories: omap|str:list|
     defaults: dict|str:list|None
```
jabdoa2, line 586: I think this would make more sense per category. For instance shortest_time_to_wizard would be get reverse_sort True and loop_champ would not.

jabdoa2:
You could add a test case here: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_HighScoreMode.py. You can find the configs for that test here: https://github.com/missionpinball/mpf/tree/dev/mpf/tests/machine_files/high_score

mpf/config_spec.yaml
```diff
@@ -582,6 +582,7 @@ hardware_sound_player:
 high_score:
     __valid_in__: mode
     award_slide_display_time: single|ms|4s
+    high_score_reverse_sort: single|bool|False
```
avanwinkle: Since this config setting is part of `high_score`, is it necessary to prefix the setting name with it? Why not just `reverse_sort`?

avanwinkle (commented):
Looks like a nice feature. Unit tests would be great, to ensure (a) that it works as expected and (b) that we don't break it down the line. Welcome to the MPF dev community!

yensho:
Thanks! Am currently working on getting the full test built for this, will push back to this request when I have finished it.

mpf/config_spec.yaml
yensho: Good point. I will correct that as well in my next push.

yensho:
I am currently having issues with my test, and while setting my value to True lets me successfully test my features, doing so also breaks quite a few other tests. I would like all of the high score tests to always use a value of false regardless of what the player/owner has input, and then I can change the actual value to what I need during my own specific test. Is that what the config located here allows me to do?

https://github.com/missionpinball/mpf/tree/dev/mpf/tests/machine_files/high_score

Just would like confirmation, since I don't have a full grasp on how all of the yamls for these configs work together.

jabdoa2:
That is the right file for this test. You can also copy the test and the config to create your own instance if that's easier.

yensho:
I created a new test, and was able to set up the test config to work as needed. Also changed the field to just reverse_sort. Please feel free to review.

mpf/config_spec.yaml
```diff
@@ -582,6 +582,7 @@ hardware_sound_player:
 high_score:
     __valid_in__: mode
     award_slide_display_time: single|ms|4s
+    reverse_sort: single|bool|False
```
avanwinkle: Nit: it'd be nice to keep this list in alphabetical order

mpf/tests/machine_files/high_score/modes/high_score/config/high_score.yaml
```diff
@@ -19,3 +19,4 @@ high_score:
       - MPF: 23
     loops:
       - JK: 42
+  reverse_sort: False
```
avanwinkle: This explicit call makes sense, but it could mask regressions for users who don't have a `reverse_sort` set. Since `False` is the default, this test should still pass without the line you've added here.

mpf/tests/test_HighScoreMode.py
```diff
@@ -472,3 +472,18 @@ def testInvalidData(self):
 
         self.assertEqual(new_score_data, self.machine.modes.high_score.high_scores)
         self.assertEqual(new_score_data, self.machine.modes.high_score.data_manager.written_data)
+
+    def test_reverse_sort(self):
+        self.start_game(4)
+        self.machine.game.game_ending()
+        self.machine.modes.high_score.high_score_config['reverse_sort'] = True
```
avanwinkle: @jabdoa2 Do you have a preference about modifying the mode config during the test run, as opposed to creating a different mode config? Are there possible holes that could be introduced by not having an explicit yaml file?

avanwinkle (commented):
Looking good, @yensho ! A few comments but I like the new feature!

mpf/config_spec.yaml
yensho: Makes sense. I can fix that.

mpf/tests/machine_files/high_score/modes/high_score/config/high_score.yaml
yensho: Maybe this is from me misunderstanding how the system works in a real life example, but the reason I've put this is here is that it needs to always be `False` when running the tests, as having a value of `True` breaks multiple other tests.

If a user has built their own config for their machine, sets `reverse_sort: True`, and for some reason or another needs to run tests, would the value in the default config still be used for the test, or would it use the user's config?

avanwinkle: My concern was less about users who set `reverse_sort: True`, which should work according to the changes you've made. I'm more curious about users who don't have _anything_ set for `reverse_sort`—which right now, is everybody. 

Since you've set the default value to be `False`, then in theory all of the existing tests should pass with no `reverse_sort` setting at all. This is important, because we don't want to force everybody to add `reverse_sort`. 

If you remove this line, will all the tests still pass? If so, that's great and we should keep it that way 😃 If not, then the default value isn't being used and it'll need a fix before we merge.

mpf/tests/test_HighScoreMode.py
jabdoa2: I guess we would need another machine directory in this case. Alternatively, we would need to change all other tests. Changing one value within a test made problems with ball devices before. So I guess copying the folder would be a good thing in this case. 

We could do something like this:
```
    def getMachinePath(self):
        if self._testMethodName = "test_reverse_sort":
            return 'tests/machine_files/high_score_reverse/'
        else:
            return 'tests/machine_files/high_score/'
```

I will commit some helper decorators for unit tests later which make this nicer but we can refactor that later.

yensho: alright, I will move forward with this direction: make a new folder for the reverse test config, and modify getMachinePath for this specific test.

mpf/config_spec.yaml
yensho: Would you mind expanding on this a bit more? Like I said elsewhere, I am still new to yaml files, so any guidance you might be able to offer with how this might be structured would be greatly appreciated. Thanks!

yensho:
Hi @jabdoa2 , I have resubmitted with changes. Could you please review?

mpf/modes/high_score/code/high_score.py
```diff
@@ -148,7 +148,7 @@ def _run(self) -> Generator[int, None, None]:
                     new_list.append((player, player[category_name]))
 
             # sort if from highest to lowest
-            new_list.sort(key=lambda x: x[1], reverse=True)
+            new_list.sort(key=lambda x: x[1], reverse=(not self.high_score_config['reverse_sort']))
```
jabdoa2: I guess we could go with ``category_name in self.high_score_config['reverse_sort']`` (or similar) here. That would allow a setting per category. Usually you want the normal score to be sorted normally (highest score first). However, scores such as "time to wizard mode" would be reverse sorted. That is why I would put this per category.

Config would then look like this:

```
high_score:
  _overwrite: True
  categories: !!omap
  - score:
      - GRAND CHAMPION
      - HIGH SCORE 1
      - HIGH SCORE 2
      - HIGH SCORE 3
      - HIGH SCORE 4
  - loops:
      - LOOP CHAMP
  - time_to_wizard:
      - FASTEST WIZARD
      - ALMOST FASTEST WIZARD
  defaults:
    score:
      - BRI: 4242
      - GHK: 2323
      - JK: 1337
      - QC: 42
      - MPF: 23
    loops:
      - JK: 42
    time_to_wizard_mode:
      - JK: 300
      - BM: 350
  reverse_sort:
     - time_to_wizard_mode
```

Makes sense?

yensho:
- `reverse_sort` is now a category
- `config` in both new machine and previous have been updated to account for category
- new machine files are still required for test because of new test data breaking old tests

mpf/modes/high_score/code/high_score.py
yensho: yes, this cleared it up for me, thank you.

jabdoa2 (approved):
Awesome! Thanks!

PR #1297, opened 2019-01-19
Event kwargs in event placeholders

avanwinkle:
This PR extends the work of evaluating event_player events (https://github.com/missionpinball/mpf/pull/1212) by including kwargs from the triggering event in the evaluation.

This change merely passes `kwargs` when evaluating the event name to be posted (_event_name_placeholder_) as well as when calculating the values of any args to be posted with that event (_event_param_).

Includes tests!


jabdoa2 (approved):
Very clean change. Like it.

PR #1303, opened 2019-02-01
fix yaml parsing issue.

jabdoa2:
for some reason we assumed 2:10 would be int and failed to process it


avanwinkle (commented):
Huh, interesting. Looks like the `float` resolver also includes the colon separator, though that doesn't seem to be hurting anything. I wonder what would use that, if anything?

jabdoa2:
> Huh, interesting. Looks like the `float` resolver also includes the colon separator, though that doesn't seem to be hurting anything. I wonder what would use that, if anything?

Good point. Will remove that also. Probably some weird notation which exists somewhere but we do not need it

PR #1327, opened 2019-04-10
Separate event handlers from normal code

jabdoa2:
Prefix all event handlers with event_. Previously, all kinds of methods had `**kwargs` and would swallow additional/incorrect arguments. Mark experienced this in custom code. This is a try to increase correctness. This only covers control events but it should be 90% of the handlers.

### DIFF
--- a/mpf/devices/ball_hold.py
+++ b/mpf/devices/ball_hold.py
@@ -68,17 +68,12 @@ class BallHold(SystemWideDevice, ModeDevice):
 
         self.source_playfield = self.config['source_playfield']
 
-    @event_handler(10)
-    def enable(self, **kwargs):
+    def enable(self):
         """Enable the hold.
 
         If the hold is not enabled, no balls will be held.
-
-        Args:
-            **kwargs: unused
         """
-        del kwargs
-
+        super().enable()
         if not self.enabled:
             self.debug_log("Enabling...")
             self._register_handlers()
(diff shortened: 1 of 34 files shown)


mpf/devices/ball_routing.py
```diff
@@ -26,9 +27,14 @@ def __init__(self, machine: "MachineController", name: str) -> None:
         self._balls_at_target = 0
         self._handler = []
 
-    def enable(self, **kwargs):
-        """Enable routing."""
+    @event_handler(2)
+    def event_enable(self, **kwargs):
```
avanwinkle: For `event_enable` handlers like this, would the default behavior inherited from `ModeDevice` suffice? Or is there a reason to explicitly define it here?

mpf/devices/ball_hold.py
```diff
@@ -68,17 +68,12 @@ def _initialize(self):
...
         """Enable the hold.
 
         If the hold is not enabled, no balls will be held.
-
-        Args:
-            **kwargs: unused
         """
-        del kwargs
-
+        super().enable()
```
avanwinkle: I'm wondering about the choice to override `self.enable()` here and call super, instead of using `EnableDisableMixin` and overriding `self._enable()`.

mpf/devices/extra_ball_group.py
```diff
@@ -128,10 +131,15 @@ def award_lit(self, **kwargs):
         else:
             posted_unlit_events = True
 
-        self.award(posted_unlit_event=posted_unlit_events)
```
avanwinkle, line 131: Ooh, sneaky little typo bug! Good catch!

mpf/devices/multiball.py
```diff
@@ -216,48 +228,49 @@ def add_a_ball(self, **kwargs):
...
 
         If the multiball is not enabled, it cannot start.
-
-        Args:
-            **kwargs: unused
         """
-        del kwargs
+        super().enable()
         self.debug_log("Enabling...")
         self.enabled = True
```
avanwinkle: This is a `ModeDevice` so it needs its own `self.enabled = True` set here. Would it be advantageous to use `EnableDisableMixin` instead?

avanwinkle (commented):
Ausgezeichnet! This is a great load of work, simple in theory but a lot in execution. I like the direction and the separation of event handlers from device behaviors, and enforcing kwargs on device methods is nice.

A few questions about enable/disable regarding `ModeDevice` and `EnableDisableMixin` overrides, but overall great work!

mpf/devices/ball_hold.py
jabdoa2: Good idea. I will review that. Maybe we should add disable to the base class also (but that would something for another PR).

mpf/devices/ball_routing.py
jabdoa2: Looks like I oversaw that. Will fix it.

mpf/devices/multiball.py
jabdoa2: Yes I guess so.

jabdoa2: done

mpf/devices/ball_hold.py
jabdoa2: done

mpf/devices/ball_routing.py
jabdoa2: done

Issue #1119, opened 2018-03-12
Show machine and player variables in Text UI


jabdoa2:
Show variables in the text UI. Show all by default. Make it configurable to always show certain vars on top (because space is limited)

woosle1234:
I would like to try to implement this issue. Can you explain a bit more about what variables you want? and how do you want the configurable variables to work?

jabdoa2:
Hi Woosle,

The idea is to show all machine variables (https://github.com/missionpinball/mpf/blob/dev/mpf/core/machine.py#L792) and player variables (https://github.com/missionpinball/mpf/blob/dev/mpf/core/player.py). Machine vars are generally accessible via self.machine.machine_var. Player vars are accessible (during a game only) via self.machine.game.player.vars. 

The text UI is rendered here: https://github.com/missionpinball/mpf/blob/dev/mpf/core/text_ui.py

woosle1234:
So would that mean I would have to update the def _draw_player_header(self): in the text_ui.py and add in self.machine.machine_var and self.machine.game.player.vars. in there?

jabdoa2:
Either there or in _update_player. There should be some space left there.

woosle1234:
Is there a way to get all the variables for machine and player? I'm thinking of using a for loop or would I have to declare each name? Also is the documentation on the player and machine variables correct in listing all the variables?

jabdoa2:
The documentation is mostly accurate but there might be more during running. I guess we could to add a getter for all machine variables and player vars. However, you could also loop over the variable i listed in my earlier post. That kind of bypasses the getter.

jabdoa2:
fixed by #1328

PR #1328, opened 2019-04-10
Adding Variables to Text UI

woosle1234:
This is for : [https://github.com/missionpinball/mpf/issues/1119](https://github.com/missionpinball/mpf/issues/1119)
I added the player index to text ui 
Facing problems with other variable types and player types
Cannot seem to iterate through self.machine.game.player.vars as an array 
Every time I try to use a for loop doing so often gives me errors of "task destroyed but it was pending" or 
image
Without the ability to access the variables as an array I can't display any more player variables
Also can't seem to access the machine variables might be the wrong name I tried using self.machine.vars, self.machine.game.machine.vars and self.machine.machine_var. 
All of the was being tested inside the text_ui.py

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,6 +19,7 @@ their spare time, unpaid, for the love of pinball!
  * Fabian Gand <gandfabian@gmail.com>
  * Anthony van Winkle <mpf@anthonyvanwinkle.com>
  * Stewart Font <yensho@gmail.com>
+ * Woosle Park <whyuwatchingthis@gmail.com>
 
 MPF is inspired by pyprocgame and skeletongame which were written by:
 
--- a/mpf/core/text_ui.py
+++ b/mpf/core/text_ui.py
@@ -274,6 +274,9 @@ class TextUi(MpfController):
             self.screen.print_at(
                 'SCORE: {:,}'.format(self.machine.game.player.score),
                 self.columns[3], self.player_start_row + 2)
+            self.screen.print_at(
+                'INDEX: {}'.format(self.machine.game.player.index),
+                self.columns[3], self.player_start_row + 3) 
         except AttributeError:
             self._update_player_no_game()
 
--- a/mpf/core/text_ui.py
+++ b/mpf/core/text_ui.py
@@ -275,8 +275,16 @@ class TextUi(MpfController):
                 'SCORE: {:,}'.format(self.machine.game.player.score),
                 self.columns[3], self.player_start_row + 2)
             self.screen.print_at(
-                'INDEX: {}'.format(self.machine.game.player.index),
-                self.columns[3], self.player_start_row + 3) 
+                'ADDITIONAL VARIABLES',
+                self.columns[3], self.player_start_row + 3)
+            add_row = 1
+            player_vars = self.machine.game.player.vars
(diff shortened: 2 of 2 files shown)


jabdoa2:
Looks great! I will test it tonight.

Player variables only exist during a game. A game only exists after the user pressed the start button. It is probably best to check for self.machine.game. It will be None if there isn't any game. Usually we use _update_player_no_game outside of a game but you can also just is an if block.

woosle1234:
I managed to add the for loop and have all player variables show up
working on machine variables next

woosle1234:
Added machine variables also changed it so that the same player variables do not get displayed multiple times Is there anything else I need to do

jabdoa2:
Looks good! Maybe we could also show the machine variables outside of games.

The test previously have been unhappy because there was a trailing whitespace. Besides that it I guess we can merge it soon.

If you want you can add a acreenshot to the docs (e.g. here http://docs.missionpinball.org/en/dev/running/commands/game.html) or also to http://missionpinball.org. You could write a blog post on http://missionpinball.org/blog.html also since this is an often requested feature. All if you want.

jabdoa2:
```
Messages
========
mpf/core/text_ui.py
  Line: 288
    pylint: unused-variable / Unused variable 'x' (col 16)
  Line: 306
    pylint: unused-variable / Unused variable 'k' (col 20)
```

Just replace those with a underscore and you should be good. If you are at it could give the other vars a longer name too.

woosle1234:
> # Messages
> mpf/core/text_ui.py
> Line: 288
> pylint: unused-variable / Unused variable 'x' (col 16)
> Line: 306
> pylint: unused-variable / Unused variable 'k' (col 20)
> 
> Just replace those with a underscore and you should be good. If you are at it could give the other vars a longer name too.

I shall

jabdoa2 (approved):
Looks awesome. Thanks!

avanwinkle:
Excellent! I wonder if there should be a config setting in the `player_vars:` configuration to enable/disable variables from this display? I have 46 player variables at the moment and most of them are immaterial, but there are a handful that I would _love_ to see. Unfortunately a straight dump of them hides the ones I'm after below the ones I don't care about.

Something like...
```
player_vars:
  research_points:
    initial_value: 0
    show_in_ui: true
```

woosle1234:
> Excellent! I wonder if there should be a config setting in the `player_vars:` configuration to enable/disable variables from this display? I have 46 player variables at the moment and most of them are immaterial, but there are a handful that I would _love_ to see. Unfortunately a straight dump of them hides the ones I'm after below the ones I don't care about.
> 
> Something like...
> 
> ```
> player_vars:
>   research_points:
>     initial_value: 0
>     show_in_ui: true
> ```

I'm not sure how I would go about this because the variables are created at game time and I wouldn't know the names to pop so I can't hard code any filtering.

jabdoa2:
One thing I noticed: You did not yet add your name to the Authors file.

woosle1234:
> One thing I noticed: You did not yet add your name to the Authors file.

My mistake I will add it now

jabdoa2:
Thanks for adding this great feature!

PR #1337, opened 2019-04-17
Allow dynamic values in `timer: control_events: value:` settings

avanwinkle:
Based on a conversation in slack, this PR extends the Timer `control_events:value` setting to support dynamic values.

```
timers:
  timer_with_player_var_control_events:
    start_value: 0
    control_events:
      - event: start_player_var_timer
        action: start
      - event: add_player_var_timer
        action: add
        value: current_player.timer_amount
      - event: subtract_player_var_timer
        action: subtract
        value: current_player.timer_amount
```

I edited _config_spec_ to set the `value:` to be a template, which grants it an `evaluate()` method that is called during the Timer event handler.

**I'm not pleased with** defining the template as `template_float` and converting it to `int` after evaluation. Every control event value is an `int` except for _change_tick_interval_, which is a `float`.  I'd love a better way to enforce the proper validation without coercing floats -> ints.

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -1542,7 +1542,7 @@ timer_control_events:  # subconfig for mode timers
     __valid_in__: None
     action: single|enum(add,subtract,jump,start,stop,reset,restart,pause,set_tick_interval,change_tick_interval,reset_tick_interval)|
     event: single|str|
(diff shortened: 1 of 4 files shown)


jabdoa2:
Leet! 👍

Issue #1180, opened 2018-05-19
Control_events in counters


jabdoa2:
Increase by two/decrease etc

kirakrishnan:
Hi jabdoa2, can I work on this.?

jabdoa2:
Yes you can. Start by setting up MPF in dev mode: http://docs.missionpinball.org/en/dev/about/contributing_to_mpf.html. Let me know if you got any questions. We will help you to find the right places to make that change.

kirakrishnan:
Hi @jabdoa2 . I have followed the steps and installed mpf and mpf-mc. I am able to see version when I do mpf --version. Please tell me what to do next

jabdoa2:
Awesome! The issue is about adding control events to the Counter class here: https://github.com/missionpinball/mpf/blob/dev/mpf/devices/logic_blocks.py#L266. 

Those should work similar to timer control_events: http://docs.missionpinball.org/en/dev/config/timers.html#control-events. Those are defined here https://github.com/missionpinball/mpf/blob/dev/mpf/config_spec.yaml#L1401 and implemented here: https://github.com/missionpinball/mpf/blob/dev/mpf/devices/timer.py#L145.

The idea is to add a ``control_events`` entry to counters in config_spec which will allow those entries in config. You can then add a test for it here: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_LogicBlocks.py and put the config for it in a mode: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/machine_files/logic_blocks/modes/mode1/config/mode1.yaml. Then go and implement some logic inside the counter (similar to the timer). I guess the first action would be to decrement the counter (first by one, then by an arbitrary amount). Then the same for incrementing and setting the value.

jonroby2:
Hi! is this still an issue? I'm looking through the code and trying to figure out how this all connects together.

jabdoa2:
Hi jonroby2,

This is still open. Feel free to ask any questions! Either here or in our forum https://groups.google.com/forum/#!forum/mpf-users

Jan

jonroby2:
Hi

Been working on this and I have a question - how does the increment/decrement value get passed to the counter? is it passed as a parameter with the event or is it set as an event and then the incement/decrement events use that value?

avanwinkle:
Hey @jonroby2-

[...]

PR #1342, opened 2019-04-23
Add control events to counters

dziedada:
Implements control events for the counter class, as well as unit tests. Implements the feature requested in issue 1180.
https://github.com/missionpinball/mpf/issues/1180

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -705,7 +705,7 @@ counters:
     control_events: list|subconfig(counter_control_events)|None
 counter_control_events:  # subconfig for mode counters
     __valid_in__: None
-    action: single|enum(add,subtract,set_value)|
+    action: single|enum(add,subtract,jump)|
     event: single|str|
     value: single|int|None  # int for all
 sequences:
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -702,6 +702,12 @@ counters:
     count_interval: single|int|1
     direction: single|str|up
     starting_count: single|template_int|0
+    control_events: list|subconfig(counter_control_events)|None
+counter_control_events:  # subconfig for mode counters
+    __valid_in__: None
+    action: single|enum(add,subtract,set_value)|
+    event: single|str|
+    value: single|int|None  # int for all
 sequences:
     __valid_in__: machine, mode
     events: list|str|
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -707,7 +707,7 @@ counter_control_events:  # subconfig for mode counters
     __valid_in__: None
     action: single|enum(add,subtract,jump)|
     event: single|str|
-    value: single|int|None  # int for all
+    value: single|template_int|None
 sequences:
     __valid_in__: machine, mode
     events: list|str|
--- a/mpf/devices/logic_blocks.py
+++ b/mpf/devices/logic_blocks.py
@@ -320,6 +320,87 @@ class Counter(LogicBlock):
             self.hit_value *= -1
         elif self.config['direction'] == 'up' and self.hit_value < 0:
             self.hit_value *= -1
+        # Add control events if included in the config
+        if self.config['control_events']:
+            self._setup_control_events(self.config['control_events'])
+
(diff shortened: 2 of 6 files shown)


jabdoa2:
Good change. The linter has some nitpicks:

```
Messages
========
mpf/devices/logic_blocks.py
  Line: 329
    pep257: D300 / Use """triple double quotes""" (found '''-quotes)
    pep257: D212 / Multi-line docstring summary should start at the first line
    pep257: D205 / 1 blank line required between summary line and description (found 0)
    pep257: D400 / First line should end with a period (not 'd')
    pep257: D401 / First line should be in imperative mood ('Return', not 'Returns')
  Line: 351
    pep257: D300 / Use """triple double quotes""" (found '''-quotes)
  Line: 364
    pep257: D300 / Use """triple double quotes""" (found '''-quotes)
  Line: 377
    pep257: D300 / Use """triple double quotes""" (found '''-quotes

```

mpf/config_spec.yaml
```diff
@@ -697,6 +697,12 @@ counters:
     count_interval: single|int|1
     direction: single|str|up
     starting_count: single|template_int|0
+    control_events: list|subconfig(counter_control_events)|None
+counter_control_events:  # subconfig for mode counters
+    __valid_in__: None
+    action: single|enum(add,subtract,set_value)|
+    event: single|str|
+    value: single|int|None  # int for all
```
jabdoa2: You could use template_int here. That would allow placeholders which are evaluated during runtime (whem you call .evaluate(additional_vars) on them). Could also be a separate future change.

mpf/config_spec.yaml
```diff
@@ -697,6 +697,12 @@ counters:
     count_interval: single|int|1
     direction: single|str|up
     starting_count: single|template_int|0
+    control_events: list|subconfig(counter_control_events)|None
+counter_control_events:  # subconfig for mode counters
+    __valid_in__: None
+    action: single|enum(add,subtract,set_value)|
```
avanwinkle: For consistency with timers, what about `add,subtract,jump` ?

mpf/devices/logic_blocks.py
```diff
@@ -304,6 +304,87 @@ def _initialize(self):
...
+
+    def check_complete(self, count_complete_value=None):
+        '''
+        Returns true if the counter has reached or surpassed its specified
+        completion value, returns False if no completion criteria or is
+        not complete.
+        '''
+        # If count_complete_value was not passed, obtain it
+        if count_complete_value is None:
+            count_complete_value = self.config['count_complete_value']\
```
avanwinkle: `count_complete_value` is already set to be `None`, so there's no need to re-set it.

```
if count_complete_value is None and self.config.get("count_complete_value"):
  count_complete_value = self.config["count_complete_value"].evaluate([])
```

mpf/devices/logic_blocks.py
```diff
@@ -304,6 +304,87 @@ def _initialize(self):
...
+                .evaluate([]) if self.config['count_complete_value']\
+                is not None else None
+
+        if count_complete_value is not None:
+
+            if self.config['direction'] == 'up' and self._state.value >= count_complete_value:
```
avanwinkle: Could be simplified?

```
if self.config['direction'] == 'up':
  return self._state.value >= count_complete_value
```

Same with the "down" version.

mpf/devices/logic_blocks.py
```diff
@@ -304,6 +304,87 @@ def _initialize(self):
...
+                return True
+
+            elif self.config['direction'] == 'down' and self._state.value <=\
+                    count_complete_value:
+                return True
+        return False
+
+    def add(self, **kwargs):
```
avanwinkle: See @jabdoa2 's recent work on event handlers to consider separating the handler into method `self.event_add` that accepts kwargs and calls `self.add` which does _not_ accept kwargs. Same for the other handler methods.

avanwinkle (changes requested):
Looks good! A few notes on the boolean evaluations, and check on the new method of handling events with kwarg-accepting methods. Thanks for the work!

jabdoa2:
@dziedada Do you still want to work on this or should we take it over?

PR #1361, opened 2019-05-08
Util string_to_list with braces

avanwinkle:
The changes in #1327 caused new evaluations of `Util.string_to_list` that can result in the splitting of conditional events when merging dictionaries.

This PR extends the logic of `string_to_list` to maintain spaces within braces when it splits, resulting in the expected behavior
```
>>> Util.string_to_list("event1 event2{not some_condition} event3")
["event1", "event2{not some_condition}", "event3"]
```

Because the regex is more computionally demanding than the native `String.replace().split()`, the new logic is only performed if the string has a brace. Otherwise, the old logic is used.

### DIFF
--- a/mpf/core/utility_functions.py
+++ b/mpf/core/utility_functions.py
@@ -102,8 +102,12 @@ class Util:
 
         """
         if isinstance(string, str):
-            # Convert commas to spaces, then split the string into a list
-            new_list = string.replace(',', ' ').split()
+            if "{" in string:
+                # Split the string on spaces/commas EXCEPT regions within braces
+                new_list = re.findall(r'([\w|-]+?\{.*?\}|[\w|-]+)', string)
+            else:
+                # Convert commas to spaces, then split the string into a list
+                new_list = string.replace(",", " ").split()
             # Look for string values of "None" and convert them to Nonetypes.
             for index, value in enumerate(new_list):
                 if isinstance(value, str) and len(value) == 4 and value.lower() == 'none':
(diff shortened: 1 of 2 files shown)


jabdoa2:
Wow where did I break that? Thanks for the fix!

mpf/core/utility_functions.py
```diff
@@ -102,8 +102,12 @@ def string_to_list(string: Union[str, List[str], None]) -> List[str]:
...
         """
         if isinstance(string, str):
-            # Convert commas to spaces, then split the string into a list
-            new_list = string.replace(',', ' ').split()
+            if "{" in string:
+                # Split the string on spaces/commas EXCEPT regions within braces
+                new_list = re.findall(r'([\w|-]+?\{.*?\}|[\w|-]+)', string)
+            else:
+                # Convert commas to spaces, then split the string into a list
+                new_list = string.replace(",", " ").split()
```
jabdoa2, line 110: I guess whitespace and/or comma is also not such a good idea, right? We should convert this into something more explicit (not necessarily now)

avanwinkle: It feels like an "it was good enough at the time" decisions, and generally speaking it's sufficient. I wouldn't want to change anything that would break existing configs, although accepting both commas and spaces is more complex than doing strict space separation.

avanwinkle:
Honestly @jabdoa2 I couldn't find exactly where it was broken. I tracked the bug based on the logs throwing for unknown slides, widgets, et cetera with names like `slide_singleplayer{ball==3 ` and that led me to the spaces in conditionals.

I then started working back through commits and found that the event player merge was what broke it, but I couldn't find where. Just made a fix instead :)

PR #1369, opened 2019-05-17
Bugfix settings dict lookup

avanwinkle:
This PR fixes a bug in `SettingsController` that prevented programmatic access to the machine settings.

The settings are stored as `machine.__dict__["_settings"]` (with underscore) but the lookup item's check was against `machine.__dict__["settings"]` (without underscore). As a result, any attempt to access `machine.settings["foo"]` would throw.

This PR adds the underscore so the lookup is in the correct place and machine settings values are accessible.

### DIFF
--- a/mpf/core/settings_controller.py
+++ b/mpf/core/settings_controller.py
@@ -63,7 +63,7 @@ class SettingsController(MpfController):
 
     def __getattr__(self, item):
         """Return setting."""
-        if "_settings" not in self.__dict__ or item not in self.__dict__['settings']:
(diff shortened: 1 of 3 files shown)


jabdoa2:
I guess we should also add a test to keep it working.

avanwinkle:
Now with tests!

Also, noticed that I accidentally pushed this branch to the main repo instead of my fork. Whoops!

jabdoa2:
Thanks for adding tests. PR looks good to merge for me!

PR #1372, opened 2019-05-20
Fix "enabled" value in placeholder for devices

avanwinkle:
This PR reverts a behavior change in #1327 that introduced a bug on Device `enabled` values in placeholders. I didn't sort out exactly the issue, but I think something was double-escaping the underscored property or doubling the alias lookback. Suffice it to say, `event{device.type.devicename.enabled}` conditions always evaluated to false, throwing on the DevicePlaceholder not having an "enabled" property.

Changing the placeholder to be `event{device.type.devicename._enabled}` behaved as expected, but that feels incorrect.

This PR reverts the alias underscore of the "enabled" property for BallHold, MultiBall, and MultiballLock, and adds tests to confirm the expected behavior.

NOTE: BallRouting also includes the "_enabled" change, but I'm not familiar enough with how routing devices work to write tests for them.

### DIFF
--- a/mpf/devices/ball_hold.py
+++ b/mpf/devices/ball_hold.py
@@ -10,7 +10,7 @@ from mpf.core.mode_device import ModeDevice
 from mpf.core.system_wide_device import SystemWideDevice
 
 
-@DeviceMonitor("balls_held", "enabled")
+@DeviceMonitor("balls_held", _enabled="enabled")
 class BallHold(EnableDisableMixin, SystemWideDevice, ModeDevice):
 
     """Ball hold device which can be used to keep balls in ball devices and control their eject later on."""
--- a/mpf/devices/ball_hold.py
+++ b/mpf/devices/ball_hold.py
@@ -10,7 +10,7 @@ from mpf.core.mode_device import ModeDevice
 from mpf.core.system_wide_device import SystemWideDevice
 
 
-@DeviceMonitor("balls_held", enabled="_enabled")
+@DeviceMonitor("balls_held", "enabled")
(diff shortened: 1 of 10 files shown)


mpf/devices/ball_hold.py
```diff
@@ -10,7 +10,7 @@
 from mpf.core.system_wide_device import SystemWideDevice
 
 
-@DeviceMonitor("balls_held", enabled="_enabled")
+@DeviceMonitor("balls_held", "enabled")
```
jabdoa2: This will fix evaluations for one shot placeholders. However, it will not allow subsections because the placeholder will not notice changes to the getter. That is the reason why I tried to use the internal variable. Maybe I just inverted the order? It is good that we have tests now. Maybe we can create a generic test which just queries all placeholders to make sure that they do not crash at least?

avanwinkle: Hey @jabdoa2 , thanks for the feedback. I tried swapping the positions to `_enabled="enabled" and my new tests still pass, is that sufficient? Could you provide tests for the situation you're describing above?

The trouble with the current bug is that it doesn't actually crash, the placeholder substitution swallows the error and returns None. But I'm all in favor of more tests!

jabdoa2: This can be tested like this: https://github.com/missionpinball/mpf/commit/1239ef4c9feb27c09b39ccd53a711976e7280aed#diff-e74bb5ba8adfd77096ec130b2124abf4R16. I'm pretty confident that your solution is right!

avanwinkle: I've pushed an update that restores the kwargs but swaps the underscore position. All the tests (including my new ones) pass, so... hurray?

PR #1374, opened 2019-05-21
Honor disable_random setting on random_event_player

avanwinkle:
This PR adds support for a documented-but-unimplemented setting for RandomEventPlayer: `disable_random`.

The underlying Randomizer accepts a disable_random setting to return items sequentially, but RandomEventPlayer never passed the setting to the Randomizer. This PR makes that change, for both machine and player RandomEventPlayers.

Also changed: currently, the RandomEventPlayer sets the `force_different` and `force_all` properties on its Randomizer every time the randomizer is accessed. This PR increases the indentation of that behavior so that the properties are only set when the randomizer is created, not every time its used.

Also also, tests! Lots of tests!

### DIFF
--- a/mpf/config_players/random_event_player.py
+++ b/mpf/config_players/random_event_player.py
@@ -36,11 +36,14 @@ class RandomEventPlayer(ConfigPlayer):
                 variables that start with "random\_" are.
                 '''
 
-            if settings['force_all']:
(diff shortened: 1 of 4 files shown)


jabdoa2:
Awesome. Looks very good!

PR #1381, opened 2019-06-07
Support Conditional Carousel Items

avanwinkle:
This PR extends the Carousel mode's `selectable_items:` list to support conditional entries. These conditions are evaluated when the mode starts and support both player and machine variable conditions.

*CHANGES:*
* Carousel stores the complete list as `self._all_items` during _mode_init_ and parses the list to create `self._items` during _mode_start_. Previously, `self._items` was created during init.
* Carousel items are no longer stored as player variables. 
  * This logic was designed to allow carousel items to be eliminated on a per-player basis. Conditional logic in the events will support this instead.
  * Player variables were written for carousel items but never modified, so this logic was incomplete and not implemented.

*QUESTIONS:*
* Carousel will throw an error on mode start if there are no valid items to select. Perhaps it should fail silently with an event instead?

*EXAMPLE: Choosing an attract-mode slide based on the number of players in the last game*
```
mode:
  code: mpf.modes.carousel.code.carousel.Carousel

mode_settings:
  selectable_items:
    - title
    - 1player_score{not machine.player2_score}
    - 2player_score{machine.player2_score and not machine.player3_score}
    - 3player_score{machine.player3_score and not machine.player4_score}
    - 4player_score{machine.player4_score}
  next_item_events: s_flipper_right_active, timer_attract_timer_complete
  previous_item_events: s_flipper_left_active
```


jabdoa2:
> QUESTIONS:
> 
>     Carousel will throw an error on mode start if there are no valid items to select. Perhaps it should fail silently with an event instead?

Maybe an empty_list event or something? Make sense to me.

avanwinkle:
Updated behavior: Now an empty list will post *(carousel_name)_items_empty* and stop the mode (without finishing the start), so game logic can handle that accordingly.

jabdoa2:
Thanks!

Issue #1385, opened 2019-06-20
Review diverter enable/activation behavior


GabeKnuth:
Why is it that a diverter with an activate_event configured instead of an activation_switch activates as soon as it's enabled instead of waiting for the activate_event? I have a diverter that can't be tied to a switch, so this is a problem.

This is well-documented in diverter code, but I don't understand the reason why it would activate right away if an activate_event is configured. Maybe it's just an oversight.

I understand how to work around this (using the same event for enable and activation, or making the enable event and deactivation event the same), but that seems counter-intuitive, especially when following the config file reference's description of the difference between "enabling" and "activating" diverters.

My suggestion would be to modify the diverter code to also look for "activate_events". Starting at line 130 in diverter.py, it currently looks like:

 ```
       if self.config['activation_switches']:
            self._enable_switches()
        else:
            self.activate()
```
The fix is probably simple, like:

```
        if self.config['activation_switches']:
            self._enable_switches()
        elif self.config['activate_events']:
            pass
        else:
            self.activate()
```
Thoughts?

GabeKnuth:
PR #1386

PR #1386, opened 2019-06-20
Don't activate diverter if activate_event present

GabeKnuth:
#1385 

Seems pretty straightforward. Checked on my end and it works as desired. Could probably be merged upstream, too, but that's up to you.

If this needs a Unit Test, let me know. I've never written one, so I don't really know the process. I can figure it out, though.


jabdoa2:
Looks good to me!

jabdoa2 (approved):

PR #1396, opened 2019-07-25
1190 modes in subfolders

pmansukhani:
Do I need more tests for the modes loaded from subfolders?

### DIFF
--- a/mpf/tests/machine_files/mode_tests/modes/mode_in_sub_folder/sub_mode/mode6/config/mode6.yaml
+++ b/mpf/tests/machine_files/mode_tests/modes/mode_in_sub_folder/sub_mode/mode6/config/mode6.yaml
@@ -1,0 +1,14 @@
+#config_version=5
+mode:
+  start_events: start_mode1
+  stop_events: stop_mode1
+  priority: 200
+  start_priority: 1
+  stop_on_ball_end: false
+  game_mode: False
+
+mode_settings:
+  this: true
+
+config:
+  - test.yaml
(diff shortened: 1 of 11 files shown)


mpf/tests/machine_files/mode_tests/modes/mode_in_sub_folder/sub_mode/mode6/config/mode6.yaml
```diff
@@ -0,0 +1,14 @@
+#config_version=5
+mode:
+  start_events: start_mode1
```
jabdoa2: Small nitpick: This should be start_mode6. Otherwise other tests will accidentally start this mode. Same for the other new modes

jabdoa2 (approved):
Looks good to me. Thanks for taking the time!

jabdoa2:
> Do I need more tests for the modes loaded from subfolders?

The test is sufficient. You could add a negative test that we do not find subfolders/more modes within another mode.

pmansukhani:
I'll fix both and also squash the commits. 

The documentation would probably need to change as well. Would there be a separate issue created for that? How is it usually handled?

jabdoa2:
Feel free to just update the configuration in mpf-docs. We rarely use issues in the docs (mostly just as a reminder for stuff to do later).

pmansukhani:
These changes are ready for merge @jabdoa2 . I'll work on the documentation at some point next week.

jabdoa2:
Awesome. Thanks!

PR #1404, opened 2019-08-07
Added placeholder on shoot_again time

pmansukhani:
Please verify if this is all that's needed.
I would also like some assistance on what type of a unit test would be needed for this. 

P.S.: I can see a linting error but it's not on one of the files I changed.


mpf/devices/multiball.py
```diff
@@ -123,7 +123,7 @@ def start(self):
         if self.balls_added_live - balls_added > 0:
             self.source_playfield.add_ball(balls=self.balls_added_live - balls_added)
 
-        if not self.config['shoot_again']:
+        if not self.config['shoot_again'].evaluate([]):
```
jabdoa2: I would only call evaluate once and store the value in a variable. It could chamge between calls (not likely here but possible).

jabdoa2:
Looks good. I will have a look at the linter. I recently made it more strict. 

To unit test this I would copy this test: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_MultiBall.py#L16. Use a machine or player variable as placeholder for shoot_again. Then use self.machine.variables.set_machine_var to set it and see if it works (i.e. No shoot again with false or 50s). Makes sense?

pmansukhani:
> Looks good. I will have a look at the linter. I recently made it more strict.
> 
> To unit test this I would copy this test: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_MultiBall.py#L16. Use a machine or player variable as placeholder for shoot_again. Then use self.machine.variables.set_machine_var to set it and see if it works (i.e. No shoot again with false or 50s). Makes sense?

Not really. First of all, isn't shoot again value taken from the config? Then how does setting it on the machine variable using `self.machine.variables.set_machine_var` allow me to test that the config value works? Am I getting it all wrong?

Also, how do I actually test that shoot again works? Is it by 1. draining the balls, 2. advancing the time, 3. depending on time advanced the ball would either be re-added or not. (Is it re-added if the time is less than the shoot again time)?

Does this make sense at all?

jabdoa2:
Previously we had a fixed shoot again value like this:

```
multiballs:
   my_multiball:
      ball_count: 3
      shoot_again: 10s
```
With your change you can use any templates:
```
multiballs:
   my_multiball:
      ball_count: 3
      shoot_again: machine.your_variable
```
Here are some examples for templates: http://docs.missionpinball.org/en/dev/config/instructions/dynamic_values.html.

Your idea to test shoot again is to the point. That is exactly what we want.

jabdoa2:
I rebased, added a test, did some minor cleanup and refactored one test method (which was named incorrectly).

PR #1406, opened 2019-08-14
Text ui config

avanwinkle:
@jabdoa2 refactored the TextUi as part of https://github.com/missionpinball/mpf/pull/1339, which improved the stability of the screen and provided a bunch more information (hurray!). Unfortunately for machines with 100+ machine and player variables, the overhead to maintain the UI had negative impacts on performance.

This PR introduces a config section `text_ui:` that allows a user to configure _which_ machine_vars and player_vars appear in the text UI. If unspecified, all will appear.

If approved, I'll update the docs to reflect the new options. Feedback/suggestions welcome!

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -1518,6 +1518,10 @@ system11:
     file_log: single|enum(none,basic,full)|basic
 text_strings:
     __valid_in__: machine, mode                 # todo add to validator
+text_ui:
+    __valid_in__: machine
+    player_vars: list|str|None
+    machine_vars: list|str|None
 tic_stepper_settings:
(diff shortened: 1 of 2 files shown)


jabdoa2:
We should have a look why this hits us so hard performance wise. I already fixed a few issues upstream to improve performance.

avanwinkle:
I can't speak much to the performance, but there's also the benefit of being able to see relevant information. With 100+ variables they will never all fit on the screen, and the machine vars are cluttered with data that's not useful in many cases. Being able to customize the output seems like an easy win, and the performance boost is a freebie :)

jabdoa2:
Yeah makes sense! Feel free to merge.

PR #1407, opened 2019-08-16
Fix inverted evaluation on show conditions

avanwinkle:
This PR fixes a bug introduced by #1397 where a `not` was accidentally dropped. The result of the bug is that show_player will play shows with conditions that evaluate False, but won't play shows with conditions that evaluate True.

https://github.com/missionpinball/mpf/pull/1397/files#diff-7a858c376f8cc5cf9b8fe88a515bc99cR38

This PR just adds the `not` back in so the show_player behaves as expected.


jabdoa2:
Uhh. Thanks. Do you have an example for a unit test for that? Otherwise this regress again later.

avanwinkle:
@jabdoa2 I wrote some tests when I originally made this change, but I couldn't get any of them to pass (even though the change is verified correct in real-world applications). I've pushed the new tests to this branch, perhaps you could take a look and advise where my tests are incorrect. Thanks!

jabdoa2:
Thanks. I will have a look.

avanwinkle (commented):
Awesome, thanks for the fix!

PR #1429, opened 2019-10-04
refactor selections

jabdoa2:


avanwinkle:
Calling out as a breaking change for when 0.53 is released.

No impact on yaml files, but custom code files may be impacted if they reference achievement state via player variable.

Previously:
```
if self.machine.game.player.achievements["mastersword"] == "completed":
```

Now, player achievements must be accessed either via index...
```
if self.machine.game.player.achievements["mastersword"][0] == "completed":
```
...or (preferably?) through the device manager and the achievement's `state` property...
```
if self.machine.device_manager.collections["achievements"].mastersword.state == "completed":
```

jabdoa2:
I guess via the the device is much nicer in code. Thanks for the reminder!

PR #1448, opened 2019-10-13
Correct a couple __valid_in__ values in config_spec

kevinleedrum:
This is probably inconsequential, but I found a couple sections in `config_spec.yaml` that had their `__valid_in__` values set to `modes`.  I'm pretty sure these should be `mode`, but please correct me if I'm wrong.


jabdoa2:
Good catch. I guess we probably did not validate those in that case. We should have a look why it worked anyways. Tests pass so this is good.

Issue #1278, opened 2018-12-31
Physically compensate balls in multiball_locks on ball start


jabdoa2:
If the device in the lock has a path to the trough compensate from it at player turn start time before ejecting a ball to the playfield.

evilmopey:
I would really like to try and solve this issue. but let me make sure I understand it since I have a very casual knowledge of pinball. Once you hit all of the required ball locks for a multiball all of the paths to the trough will be blocked off for enough time to allow balls to get past the trough and then the paths will open again.

while looking at the multiball_lock.py and doing background research I have a few questions: 
The trough is the the part at the bottom of the machine that leads the ball out of play?
What part of the Pin Ball machine would block the trough? 

Sorry for the long post and thanks for your time

jabdoa2:
Hi evilmopey,

Pinball machines are multiplayer machines. Every player has multiple ball (usually 3) and player take turns (player 1 ball 1, player 2 ball 1, player 1 ball 2 and so on). Consider a machine with a lock (called multiball_lock in MPF) which will give you some multiball. If you lock a ball on your first ball and then your loose your (new) ball before starting the multiball the ball will still be physically there for the next player. There are multiple strategies about what happens when the next player completes the lock and starts the multiball leaving you with an empty lock on your next ball (see http://docs.missionpinball.org/en/dev/game_logic/multiballs/multiball_locks.html for the four strategies in MPF). If you selected the "virtual_only" strategy MPF will give you a multiball nontheless and compensate the missing balls from the trough (where the balls which are not in play are stored below the apron: http://docs.missionpinball.org/en/dev/mechs/troughs/index.html). In some machines it is physically possible to "feed" a ball from the trough (via the launcher http://docs.missionpinball.org/en/dev/mechs/plungers/index.html) directly into the lock (often via a controllable diverter http://docs.missionpinball.org/en/dev/mechs/diverters/index.html). MPF knows how to route those balls (if there is a route). The lock has to request a ball in this case (via a ball_device.request_ball() call to the lock ball device). Makes sense? Let me know if anything is unclear.

evilmopey:
Thank you for the detailed response it really helped me gain clarity. The one area I am still confused about is how min_virtual_physical differs from physical?

P.S. I am working on the solution and will submit a pull request this weekend.😄

jabdoa2:
[...]

PR #1449, opened 2019-10-23
Issue #1278

evilmopey:
Created a compensate_lock function that make sure the right amount of balls are in the lock. If the numbers don't match up the machine will compensate by removing or adding balls.


avanwinkle (commented):
How will this affect machines that don't have a path directly to the lock?

mpf/devices/multiball_lock.py
```diff
@@ -305,3 +305,26 @@ def _post_events(self, device, **kwargs):
...
         """Request new ball to playfield."""
         self.source_playfield.add_ball(balls=balls)
+    
+    def compensate_lock(self):
+        # evilmopey solution for issue 1278
+        if self.config['locked_ball_counting_strategy'] == "no_virtual":
+            self._locked_balls = 0
+            # remove balls that were in lock
+            balance = (self._locked_balls - (-1 * self._physically_locked_balls))  # -1 to make the number positive
+            for i in range(1, balance):
```
jabdoa2: I don't think we need to remove balls. That would give the current player more balls.

mpf/devices/multiball_lock.py
```diff
@@ -305,3 +305,26 @@ def _post_events(self, device, **kwargs):
     def _request_new_balls(self, balls):
         """Request new ball to playfield."""
         self.source_playfield.add_ball(balls=balls)
+    
+    def compensate_lock(self):
+        # evilmopey solution for issue 1278
+        if self.config['locked_ball_counting_strategy'] == "no_virtual":
```
avanwinkle: If this approach is only applicable to the `"no_virtual"` approach, could you wrap the calling function above instead of having so many conditionals?

```
if self.config['locked_ball_counting_strategy'] == "no_virtual":
  self.compensate_lock()
```

mpf/devices/multiball_lock.py
```diff
@@ -305,3 +305,26 @@ def _post_events(self, device, **kwargs):
...
+            self._locked_balls = self.locked_balls(self)
+        if self._locked_balls > self._physically_locked_balls:
+             # This means we need to add more balls to the physical lock
+            balance = self._locked_balls - self._physically_locked_balls
+            self._request_new_balls(balance)
+        else:
+            # This means we need to remove balls from the physical lock
+            balance = self._physically_locked_balls - self._locked_balls
+            for i in range(1, balance):
+                self.lock_devices[0].eject()
```
avanwinkle: From my understanding, with no virtual lock counting then any balls locked by player 1 will be ejected before player 2 starts. If that's the case, you'll want to block the queue for these balls to drain so they're not on the playfield when player 2's turn starts. See above for an example

mpf/devices/multiball_lock.py
```diff
@@ -305,3 +305,26 @@ def _post_events(self, device, **kwargs):
...
+        if self.config['locked_ball_counting_strategy'] == "physical_only":
+            self._locked_balls = self.locked_balls(self)
+        if self.config['locked_ball_counting_strategy'] == "virtual_only" or "min_virtual_physical":
+            self._locked_balls = self.locked_balls(self)
+        if self._locked_balls > self._physically_locked_balls:
+             # This means we need to add more balls to the physical lock
+            balance = self._locked_balls - self._physically_locked_balls
+            self._request_new_balls(balance)
```
jabdoa2: The condition looks right. However, request_new_balls is not the correct function because it will request the ball to the playfield. Instead you would call request_ball on self.lock_devices[x].

mpf/devices/multiball_lock.py
jabdoa2: Removing balls is not needed I guess. This would require some more logic.

jabdoa2 (changes requested):
The approach is good. Some small comments below.

I think it would help to add a test case to see if things work as expected.

With the test we will also need to check if there actually a path to the trough. Otherwise some other tests will fail or the code will crash.

mpf/devices/multiball_lock.py
evilmopey: Thanks for the feedback everyone I will edit my code with your input in mind.

mpf/devices/multiball_lock.py
evilmopey: So I took out the if statements for both no_virtual and physical_only leaving only the if statement that pertains to virtual_only or min_virtual_physical. This if statement only adds balls not takes them away. My only problem now is that on line 9 MYPY is false, this blocks the class BALLDevice from being imported and being able to use the request_ball function. I added the import above it so I could use the request_ball function.

mpf/devices/multiball_lock.py
```diff
@@ -305,3 +306,12 @@ def _post_events(self, device, **kwargs):
...
         """Request new ball to playfield."""
         self.source_playfield.add_ball(balls=balls)
+    
+    def compensate_lock(self):
+        # evilmopey solution for issue 1278
+        if self.config['locked_ball_counting_strategy'] == "virtual_only" or "min_virtual_physical":
+            if self._locked_balls > self._physically_locked_balls:
+                # This means we need to add more balls to the physical lock
+                ball_deficit = self._locked_balls - self._physically_locked_balls
+                BallDevice.request_ball(self.lock_devices[0], ball_deficit)
```
avanwinkle, line 316: How does `request_ball` work for locks that don't have a route from the trough? Will it be ignored, or wait until a playfield ball lands in it?

evilmopey: So when request_ball is called it starts a for loop that iterates as many times as the value balls were passed in. Each iteration calls setup_queue_eject_to_target, if the ball cannot be routed to the target a tuple entry will then be made into the _ball_requests_ deque. So to my knowledge the request is not being ignored.

evilmopey:
So what would a test case look like on this project?

jabdoa2:
> So what would a test case look like on this project?

You could extend this test: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_MultiballLock.py. The corresponding config would be here: https://github.com/missionpinball/mpf/tree/dev/mpf/tests/machine_files/multiball_locks. You can use self.start_game()+self.add_player() to start a two player game. Then you can use self.drain_one_ball() to end the ball.

evilmopey:
I am trying to test the issue but keep getting errors, can someone point out what I am doing wrong?
https://github.com/evilmopey/issue-1278

PR #1458, opened 2019-11-21
Support arbitrary starting states for StateMachine

avanwinkle:
While `StateMachine` devices can support arbitrary state names for their progression, they currently must use `start` as their initial state.

This PR allows State Machines to define any of their states to use as the initial state, with `start` as the default value. As a result, there should be no breaking change for existing machines that do not specify `starting_state:`.

Tests included!


jabdoa2:
Very well done!

PR #1461, opened 2019-11-24
Fixed issue with setting number of LEDs per node

densminger:
The comment says to set LEDs to 96 and switches to 60.  This wasn't what was happening in the code. As a result, some of the LEDs on some nodes weren't working if their indices were too large. Should mpf auto-detect the number of LEDs/switches based on what's in the config file and set this accordingly?


jabdoa2:
Autodetecting this is tricky in MPF since init of the platform happens before init of all the switches and lights. We could reconfigure this on the fly when configuring switches/lights. Or we can add it to some node config.

PR #1462, opened 2019-11-24
CoilSetPriority causing desyncs

densminger:
The CoilSetPriority command is causing a desync condition on Node 9 and above.  If I recall right, this was added to troubleshoot hardware rules and isn't strictly required by spike, so I removed it.


jabdoa2:
I guess we could also make this configurable per node board.

PR #1463, opened 2019-11-25
Ball save dynamic values

avanwinkle:
This PR extends the behavior of `BallSave` to accept templates as the active time value. The default behavior is unchanged, so this change shouldn't break any existing code.

This change allows the use case where a player may earn a reward that increases their ball save time. The existing ball save(s) can be used and the active time changed on-the-fly.

```
ball_saves:
  default:
    active_time: 30 if current_player.extended_save else 10
  dynamic:
    active_time: 20 * (1 + current_player.ball_save_bonus_pct)
```

For readability, the `active_time` integer value is assumed to be _secs_ rather than _ms_. There is no functional difference, but it "feels" more natural to say `active_time: 30` rather than `active_time: 30000`.

Tests included!

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -199,7 +199,7 @@ ball_saves:
     __valid_in__: machine, mode
     __type__: device
     source_playfield: single|machine(ball_devices)|playfield
-    active_time: single|template_secs|0
+    active_time: single|template_ms|0
     eject_delay: single|ms|0
     only_last_ball: single|bool|False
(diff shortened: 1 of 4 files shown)


jabdoa2:
Nice. Good addition. I will tag it as breaking change because any ball_saves which use an active_time without unit such as "30" instead of "30s" will break.

avanwinkle:
Good callout. All of my code (and all of the tests, apparently) use time strings with an "s"/"ms" suffix so none failed on the active_time change. Thanks for the check!

PR #1464, opened 2019-11-25
Auto-detect FAST device ports

avanwinkle:
This PR enables the auto-detection of FAST hardware on serial ports. It's a crude, first-draft implementation, but it seems to work well on my hardware.

This change is to support people who use multiple machines to power their hardware, and/or have multiple USB ports that assign different `tty` ports to the same device. With the `autodetect` feature, these people don't have to change their hardware config each time they switch computers or USB ports.

With the `ports: autodetect` feature, MPF will attempt to find four tty devices with identical, sequential names ending in `0-4` or `A-D`. If it finds those four, it assumes that the middle two are the FAST Net and RGB devices.

It may not be the most performant approach, but this routine is used only once at bootup so it shouldn't be too bad...

### DIFF
--- a/mpf/platforms/fast/fast.py
+++ b/mpf/platforms/fast/fast.py
@@ -7,7 +7,7 @@ boards.
 import os
 from copy import deepcopy
 from distutils.version import StrictVersion
-
+import serial.tools.list_ports
 from typing import Dict, Set
 
 from mpf.platforms.fast.fast_io_board import FastIoBoard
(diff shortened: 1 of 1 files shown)


jabdoa2:
Good enough most setups. In more complex scenarios we would have to ID all connected ports to find NET and RGB. For instance my machine has 12 ttyUSB (Nano and two DMDs). But this probably works fine for anybody else.

PR #1479, opened 2019-12-23
Support 'priority' for ball_hold and ball_lock devices

avanwinkle:
This PR adjusts the behavior for `ball_holds` and `ball_locks` to support relative priority.

Currently, when a ball lock creates its _ball_enter_ handler, it uses the priority of the mode it's in. When a ball hold creates its _ball_enter_ handler, it uses no priority.

This change configures both holds and locks to use the mode priority *plus* an optional `priority:` config value. As a result, a ball hold and ball lock in the same mode can have a predictable order of event handling.

### DIFF
--- a/mpf/devices/multiball_lock.py
+++ b/mpf/devices/multiball_lock.py
@@ -144,7 +144,8 @@ class MultiballLock(EnableDisableMixin, ModeDevice):
                 self.config['locked_ball_counting_strategy']))
 
     def _register_handlers(self):
-        priority = self.mode.priority + self.config['priority']
+        priority = (self.mode and self.mode.priority or 0) +
+            self.config['priority']
         # register on ball_enter of lock_devices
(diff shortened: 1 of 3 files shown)


jabdoa2:
Nice!

jabdoa2 (approved):

mpf/devices/multiball_lock.py
```diff
@@ -144,14 +144,16 @@ def locked_balls(self, value):
                 self.config['locked_ball_counting_strategy']))
 
     def _register_handlers(self):
+        priority = (self.mode and self.mode.priority or 0) + \
```
jabdoa2: Nit: While `(self.mode and self.mode.priority or 0)` is correct I would use: `(self.mode.priority if self.mode else 0) ` to make it more clear what is happening here

PR #1480, opened 2019-12-23
Include 'triggering_switch' in combo switch events

avanwinkle:
This PR adds a new kwarg, `triggering_switch`, to events posted by a ComboSwitch device.

When the player enters a combination switch, e.g. hold one flipper and press the other, it can be useful to know _which_ switch was hit when. Currently this requires quite a bit of logic to implement.

With this PR, the combo events include a kwarg `triggering_switch` that declares the `tag:` group number of the switch that caused the event.

For example:
```
combo_switches:
  both_flippers:
    tag_1: left_flipper
    tag_2: right_flipper
    events_when_both: flipper_cancel
```

Player holds left flipper, presses right flipper:
```
Event: ======'both_flippers'====== Args={'triggering_switch': 2}
```
Player releases left flipper, still holding right:
```
Event: ======'both_flippers_one'====== Args={'triggering_switch': 1}
```

This change makes it easy for game logic to perform actions based on combo switches, for example rotating selections left or right by holding one flipper and pressing the other, or triggering event based on which direction an orbit shot was completed in.

### DIFF
--- a/mpf/devices/combo_switch.py
+++ b/mpf/devices/combo_switch.py
@@ -1,4 +1,6 @@
 """Contains the Combo Switch device class."""
+from functools import partial
+
 from mpf.core.delays import DelayManager
 from mpf.core.device_monitor import DeviceMonitor
 from mpf.core.mode import Mode
@@ -119,7 +121,7 @@ class ComboSwitch(SystemWideDevice, ModeDevice):
             self._activate_switches_1(switch_name)
         else:
             self.delay.add_if_doesnt_exist(self.config['hold_time'],
-                                           self._activate_switches_1,
+                                           partial(self._activate_switches_1, switch_name),
                                            'switch_1_active')
 
     def _switch_2_went_active(self, switch_name, **kwargs):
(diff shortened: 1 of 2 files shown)


jabdoa2:
Good one!

mpf/devices/combo_switch.py
```diff
@@ -247,7 +247,7 @@ def _switch_state(self, state):
         self.debug_log("New State: %s", state)
 
         for event in self.config['events_when_{}'.format(state)]:
-            self.machine.events.post(event)
+            self.machine.events.post(event, triggering_switch=triggering_switch)
```
jabdoa2: We could also include the switch name here.

jabdoa2 (approved):

mpf/devices/combo_switch.py
avanwinkle: Updated! Now includes `triggering_group` with the group number and `triggering_switch` with the switch name.

avanwinkle:
Thanks for the improvements!

PR #1482, opened 2019-12-29
Allow Fast RGB crashes to be ignored

avanwinkle:
### Summary
There is an instability in the Fast RGB processor that causes it to crash, typically the first time a game is played after bootup. While unfortunate, the crash does not _have_ to be fatal.

### Details
This PR adds an option to the `fast:` config settings: `ignore_rgb_crash`. If `True`, then RGB crashes will be logged but MPF will not exit. All light state will be lost, but any active light shows will immediately reappear when MPF sends the next step of the show. Any lights that were on static will be turned off.

When the RGB reboot occurs, a new event is posted: _fast_rgb_rebooted_. This event allows the game designer to provide emergency fallback lighting to ensure that the playfield is properly lit and any required lights are re-established. 

### Debug Log
```
INFO : FAST : Connected! Processor: RGB, Board Type: FP-CPU-002-2, Firmware: 01.00
INFO : EventManager : Event: ======'machine_var_fast_rgb_firmware'====== Args={'value': '01.00', 'prev_value': None, 'change': True}
INFO : EventManager : Event: ======'machine_var_fast_rgb_model'====== Args={'value': 'FP-CPU-002-2', 'prev_value': None, 'change': True}
ERROR : FAST : The FAST RGB processor rebooted. Unfortunately, that means that it lost all it's state (such as hardware rules or switch configs). This is likely cause by an unstable power supply but it might as well be a firmware bug. Ignoring RGB crash and continuing play.
INFO : EventManager : Event: ======'fast_rgb_rebooted'====== Args={'msg': '00'}
WARNING : FAST : Received malformed message: <0x11><0x11>XX:F from RGB
WARNING : FAST : Port RGB received more messages than were sent! Resetting!
ERROR : FAST : The FAST RGB processor rebooted. Unfortunately, that means that it lost all it's state (such as hardware rules or switch configs). This is likely cause by an unstable power supply but it might as well be a firmware bug. Ignoring RGB crash and continuing play.
INFO : EventManager : Event: ======'fast_rgb_rebooted'====== Args={'msg': '02'}
```

### DIFF
--- a/mpf/platforms/fast/fast.py
+++ b/mpf/platforms/fast/fast.py
@@ -863,10 +863,9 @@ class FastHardwarePlatform(ServoPlatform, LightsPlatform, DmdPlatform,
             remote_processor == self.rgb_connection.remote_processor
         if msg in ('00', '02'):
             action = "Ignoring RGB crash and continuing play." if ignore_rgb else "MPF will exit now."
-            self.error_log("The FAST {} processor rebooted. Unfortunately, that means that it lost all it's state "
-                           "(such as hardware rules or switch configs). This is likely cause by an unstable power "
-                           "supply but it might as well be a firmware bug. {}"
-                           .format(remote_processor, action))
+            self.error_log("The FAST %s processor rebooted. Unfortunately, that means that it lost all its state "
+                           "(such as hardware rules or switch configs). This is likely caused by an unstable "
+                           "power supply but it might also be a firmware bug. %s", remote_processor, action)
             if ignore_rgb:
                 self.machine.events.post("fast_rgb_rebooted", msg=msg)
                 return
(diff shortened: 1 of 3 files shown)


mpf/platforms/fast/fast.py
```diff
@@ -859,8 +859,15 @@ def clear_hw_rule(self, switch, coil):
     def receive_bootloader(self, msg, remote_processor):
         """Process bootloader message."""
         self.debug_log("Got Bootloader message: %s from", msg, remote_processor)
+        ignore_rgb = self.config['ignore_rgb_crash'] and \
+            remote_processor == self.rgb_connection.remote_processor
         if msg in ('00', '02'):
-            self.error_log("The FAST %s processor rebooted. Unfortunately, that means that it lost all it's state "
+            self.error_log("The FAST {} processor rebooted. Unfortunately, that means that it lost all it's state "
```
jabdoa2: I think the convention for log lines is still the old percentage syntax.

avanwinkle: I saw that, but wasn't sure if that was by design or just legacy code. I'm in favor up updating old syntax as files are updated, but I'm happy to stick with the old style if that's desirable.

jabdoa2: The reason for the older syntax here is that it will can be lazy evaluated. If you use format we will always replace those strings even when logging is disabled. With percentage syntax replacement will happen only if the string is displayed. The linter even complains about this when you use logger.debug but since we wrapped it this is not happening here.

Issue #1494, opened 2020-02-02
Prevent adding players during high scores on one ball game


jabdoa2:
Currently, you can add players when in high score mode in a one ball game. We only check for "ball==1" which still holds true at that point. Also check "game.ending" or something similar to prevent this. Also add a unit test.

Discussion: https://groups.google.com/forum/#!topic/mpf-users/vQiJRQNLRpQ

seanirby:
Taking a look.

PR #1509, opened 2020-03-21
Add game ending check before adding another player

seanirby:
I'm not sure what you were expecting for the unit test so I chose to put it in the existing high score mode test class simply because it was convenient.  If you want it elsewhere please let me know.


jabdoa2:
Looks very good. Thanks!

jabdoa2:
fixes #1494

Issue #1504, opened 2020-02-29
Support placeholders in keep_multiplier in bonus mode


jabdoa2:
Support placeholders in `keep_multiplier` (and maybe some more settings) in bonus mode:
```
##! mode: bonus

mode_settings:
    display_delay_ms: 500
    hurry_up_delay_ms: 500
    hurry_up_event: flipper_cancel
    # end_bonus_event: player_score
    #keep_multiplier: true #false            # automatically resets the bonus_multiplier var - this works
    keep_multiplier: true if {current_player.bonus_multiplier<=4} else false    # doesn't work
    #keep_multiplier: true if "{player_bonus_multiplier{value<5}" else false    # doesn't work
    bonus_entries:
        - event: eob_bonus
          score: 1000
          player_score_entry: bonus
          reset_player_score_entry: true
          skip_if_zero: true
```

See: https://groups.google.com/forum/#!topic/mpf-users/hWbm1qx3kKc

seanirby:
Taking a look

PR #1510, opened 2020-03-23
Add dynamic value support for 'keep_multiplier' setting

seanirby:
Let me know what you think.  I'd like to test this on my machine before merging but submitting this now in case you have any comments.

### DIFF
--- a/mpf/modes/bonus/code/bonus.py
+++ b/mpf/modes/bonus/code/bonus.py
@@ -1,6 +1,6 @@
 """Bonus mode for MPF."""
 from mpf.core.mode import Mode
-
+from mpf.core.placeholder_manager import NativeTypeTemplate, BoolTemplate
 
 class Bonus(Mode):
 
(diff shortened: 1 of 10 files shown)


mpf/modes/bonus/code/bonus.py
```diff
@@ -165,6 +165,12 @@ def _total_bonus(self):
     def _end_bonus(self, **kwargs):
         del kwargs
         self.debug_log("Bonus done")
-        if not self.settings['keep_multiplier']:
+        keep_multiplier = self.settings['keep_multiplier']
+
+        keep_multiplier_bool = isinstance(keep_multiplier, bool) and keep_multiplier
```
jabdoa2: Just calling evaluate should be enough. It should be always a Placeholder instance.

seanirby: I think I see what's wrong.  In one of the existing tests the setting is being overridden with a bool value mid-test.  

`        
        self.machine.modes["bonus"].config['mode_settings']['keep_multiplier'] = False
`

I don't think this can happen now that we've changed the config type to a template so theres no need to account for the `keep_multiplier_bool` case.  If I update the test with the following it will pass:

`self.machine.modes["bonus"].config['mode_settings']['keep_multiplier'] = self.machine.placeholder_manager.build_bool_template("False")`

jabdoa2:
fixes #1504

Issue #1495, opened 2020-02-02
Fix encoding error in BCP


jabdoa2:
It seems as "#" in a string currently breaks bcp. Properly encode and decode that.

```
2020-01-26 09:51:54,111 : MPF-MC BCP Server : DECODE BCP ERROR. Message: trigger?json={"name": "slides_play", "settings": {"high_score_enter_initials": {"target": null, "background_color": [0.0, 0.0, 0.0, 1.0], "priority": null, "show": true, "force": false, "expire": null, "slide": null, "tokens": {}, "action": "play"}}, "context": "high_score", "calling_context": "high_score_enter_initials", "priority": 1600, "award": "HIGH SCORE #2", "player_num": 1, "value": 81120}
2020-01-26 09:51:54,561 : mpfmc : Shutting down due to child thread crash
2020-01-26 09:51:54,561 : mpfmc : Crash details: Traceback (most recent call last):
  File "c:\users\andrew\appdata\local\programs\python\python36\lib\site-packages\mpfmc\core\bcp_server.py", line 156, in run
    self._process_receives_messages(commands)
  File "c:\users\andrew\appdata\local\programs\python\python36\lib\site-packages\mpfmc\core\bcp_server.py", line 186, in _process_receives_messages
    self.process_received_message(decoded_cmd)
  File "c:\users\andrew\appdata\local\programs\python\python36\lib\site-packages\mpfmc\core\bcp_server.py", line 247, in process_received_message
    cmd, kwargs = bcp.decode_command_string(message)
  File "c:\users\andrew\appdata\local\programs\python\python36\lib\site-packages\mpf\core\bcp\bcp_socket_client.py", line 45, in decode_command_string
    kwargs = json.loads(bcp_command.query[5:])
  File "c:\users\andrew\appdata\local\programs\python\python36\lib\json\__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "c:\users\andrew\appdata\local\programs\python\python36\lib\json\decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "c:\users\andrew\appdata\local\programs\python\python36\lib\json\decoder.py", line 355, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 339 (char 338)
```

seanirby:
Taking a look

PR #1512, opened 2020-04-11
Fix to allow '#' characters in config values

seanirby:
So I think the issue is that MPF uses URL parsing libs for parsing the BCP query.   The value gets encoded correctly but when decoding using `urlsplit` the part of the query after the `#` is treated as a URL fragment so it is removed when decoding.  If we call `urlsplit` with `allow_fragments=False` then this appears to fix the issue.   From the python docs:

> If the allow_fragments argument is false, fragment identifiers are not recognized. Instead, they are parsed as part of the path, parameters or query component, and fragment is set to the empty string in the return value.

https://docs.python.org/3/library/urllib.parse.html

### DIFF
--- a/mpf/core/bcp/bcp_socket_client.py
+++ b/mpf/core/bcp/bcp_socket_client.py
@@ -39,7 +39,7 @@ def decode_command_string(bcp_string) -> Tuple[str, dict]:
     will be preserved.
 
     """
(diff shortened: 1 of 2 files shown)


jabdoa2:
Good find. Fixes #1495. Thanks!

PR #1515, opened 2020-04-20
Bugfix for variable shows

seanirby:
Found a bug while trying to play a show that updates variables from a standalone show file.  The calling_context arg is passed as an integer [here](https://github.com/missionpinball/mpf/blob/dev/mpf/assets/show.py#L669) but `VariablePlayer.play` expects it to be a string.  I'm guessing this is typically meant to be equal to a mode name.


jabdoa2:
Good one. Thanks for testing and fixing!

PR #1516, opened 2020-04-23
Fix bug for non-connected switches

seanirby:


jabdoa2:
Oh I broke that. Thanks for fixing!

PR #1520, opened 2020-06-22
Added pinproc compiled libraries for Windows (Python 3.7, 3.8)

qcapen:


jabdoa2:
Awesome! Would it be hard to recompile for 3.5 and 3.6 as well? That would bring our proc improvements to windows as well.

qcapen:
I will do that. Shouldn't take me too long. Hopefully these work as I don't
have hardware to test. I was able to load the modules in Python on the
command line, but that is as much as I can test.

Quinn

PR #1522, opened 2020-06-23
Rebuild pypinproc for windows platforms

qcapen:
My last attempt to rebuild all the pypinproc libraries for windows ended up with a dll conflict in the installation wheel (can't have a python lib .pyd and a .dll in the same folder with the same base file name).  This new solution combines libpinproc and pypinproc into a single extension module for Windows which should simplify things. Users still need to have the correct ftx2xx.dll (32bit/64bit) installed in their path.


jabdoa2:
Thanks for caring Quinn!

PR #1524, opened 2020-07-11
Show "start_running" option

avanwinkle:
This PR adds a new config setting for Shows called `start_running`. It accepts a boolean value (default `True`).

If `False`, the show will start on step zero and immediately pause. The show will then hold until it is manually resumed.

Run!

Also some other fixes, noted in the comments.


mpf/config_players/device_config_player.py
```diff
@@ -105,8 +105,8 @@ def _expand_device(self, device):
                 if not self.__class__.allow_placeholders_in_keys or "(" not in device_name:
                     # no placeholders
                     return self.raise_config_error(
-                        "Could not find a {} device with name or tag {}.".format(
-                            self.device_collection.name, device_name),
+                        "Could not find a {} device with name or tag {}, from list {}".format(
```
avanwinkle, line 108: When the device list has an empty option, the `device_name` is empty and this error log isn't very helpful. Adding `device_or_tag_names` helps to identify the error.

mpf/core/config_processor.py
```diff
@@ -179,7 +179,7 @@ def _load_config_file_and_return_loaded_files(
 
         self.log.info('Loading config: %s', filename)
 
-        if config_type in ("machine", "mode"):
+        if config_type in ("machine", "mode") and config_spec:
```
avanwinkle: This is for my own personal use. I have an automation script that imports and parses the config files, but a recent refactor of _config_processor_ changed the behavior and expects that the _config_spec_ has been loaded into memory. My automation script doesn't do this, so I need to ignore config spec parsing in that case.

mpf/assets/show.py
```diff
@@ -574,7 +575,7 @@ def _remove_delay_handler(self):
     def pause(self):
         """Pause show."""
         self._remove_delay_handler()
-        if self.show_config.events_when_stopped:
+        if self.show_config.events_when_paused:
```
avanwinkle, line 582: I believe this was just an oversight in the original implementation, or maybe `events_when_paused` was never set up.

mpf/core/config_processor.py
jabdoa2: I see that. I recently added a config reference parser to expose a stable interface for config reference. It is already used in mpf language server and mpf docs. Guess that might make sense for you too.

Out of curiosity: What does your script do?

avanwinkle: A config parser is exactly what I'm after! Could you point to a reference of it in use so I know where to start?

My script is for audio management, since I have a very large library of sfx to pull from and I constantly add sounds, remove sounds, and move sounds from one mode to another. My script traverses the file trees (once for my source library, once for my machine folder) and traverses the config files, then moves all the sound assets into the appropriate folders.

avanwinkle: It will also do a bulk export of all audio files for the machine, which is helpful since I don't track those in git. When I setup MPF on a different machine, this script's import/export comes in real handy :)

https://github.com/avanwinkle/masseffect2/blob/powers-refactor/mesound.py

jabdoa2: Sounds interesting. Might be interesting for other users as well. We could turn this into MPF commands if you want.

Just checked the code. I only moved the event parser so far. Here is the config parsing from the mpf language server: https://github.com/missionpinball/mpf-ls/blob/dev/mpfls/workspace.py

mpf/devices/state_machine.py
```diff
@@ -52,7 +52,7 @@ def validate_and_parse_config(self, config: dict, is_mode_config: bool, debug_pr
     @property
     def state(self):
         """Return the current state."""
-        if self.config['persist_state']:
```
avanwinkle, line 55: This corrects a behavior in MPF Monitor, which attempts to initialize all state_machines on bootup (for displaying in the device list). Without a player, Monitor's attempt to get the state of the machine causes a crash.

jabdoa2: Good catch. We should probably add some integration test for this so that this will not happen again for other devices. We already had that a few times.

avanwinkle:
@jabdoa2 I've refactored my asset management script into a standalone pip module, in case others would find it useful. Right now it lives at https://github.com/avanwinkle/mpf-asset-manager

I'd be happy to migrate it over to a repo owned by `missionpinball` and let you set it up in pypi using MPF's account. That way others can improve it as they see fit!

jabdoa2:
> @jabdoa2 I've refactored my asset management script into a standalone pip module, in case others would find it useful. Right now it lives at https://github.com/avanwinkle/mpf-asset-manager
> 
> I'd be happy to migrate it over to a repo owned by `missionpinball` and let you set it up in pypi using MPF's account. That way others can improve it as they see fit!

Wow that sounds awesome. I would love that.

jabdoa2:
> @jabdoa2 I've refactored my asset management script into a standalone pip module, in case others would find it useful. Right now it lives at https://github.com/avanwinkle/mpf-asset-manager
> 
> I'd be happy to migrate it over to a repo owned by `missionpinball` and let you set it up in pypi using MPF's account. That way others can improve it as they see fit!

Wow that sounds awesome. I would love that.

jabdoa2 (approved):

PR #1525, opened 2020-07-19
Count based eject timings

jabdoa2:


avanwinkle:
Ah, this looks familiar. Very familiar ;) https://github.com/missionpinball/mpf/pull/1108/files

From what I can see you're extending the multiple eject times from *enable_coil_ejector* to work on all types of ejectors?

jabdoa2:
> Ah, this looks familiar. Very familiar ;) https://github.com/missionpinball/mpf/pull/1108/files
> 
> From what I can see you're extending the multiple eject times from _enable_coil_ejector_ to work on all types of ejectors?

Same mechanism but slightly different implementation (as it also has to work for jam and retry pulse). I might refactor those into one way of doing stuff. Will look at that again tomorrow. Maybe also refactor hold ejector.

PR #1530, opened 2020-09-02
Twitch Chat Events

mwseiden:
This is a mostly direct drop in from my Twitch IRC client code. Some issues with this code that need to be resolved:

No tests. I haven't even tested this as an MPF plugin manually. (in my code it's a scriptlet)
The IRC thread may need to be started later in the machine startup
The IRC thread should be able to be configured with machine variables. Currently it's config section based. Checking in passwords is bad.
The IRC thread should be stopped and restarted when those variables change.
The IRC thread is never shut down gracefully on exit
Is this the proper way to do threading in MPF?
It requires the irc library
Password token needs to be generated here: https://twitchapps.com/tmi/  the OAuth login should probably be integrated into MPF somehow

What it should do:
1. Log on to Twitch chat
2. Monitor chat for messages and post the `twitch_chat_message` event
3. Monitor chat for bit donations and post the `twitch_bit_donation` event
4. Monitor chat for subscriptions and post the `twitch_subscription` event

### DIFF
--- a/mpf/plugins/twitch/twitch_client.py
+++ b/mpf/plugins/twitch/twitch_client.py
@@ -26,7 +26,7 @@ class TwitchClient(irc.bot.SingleServerIRCBot):
         server = 'irc.chat.twitch.tv'
         port = 6667
         self.log.info('Connecting to ' + server + ' on port ' + str(port) + '...')
-        irc.bot.SingleServerIRCBot.__init__(self, [(server, port, 'oauth:' + password)], username, username)
+        super().__init__(self, [(server, port, 'oauth:' + password)], username, username)
         # self.connection.add_global_handler("all_events", self.on_all_events, -100)
 
     def on_welcome(self, c, e):
@@ -114,7 +114,7 @@ class TwitchClient(irc.bot.SingleServerIRCBot):
 
     def post_event_in_mpf(self, event, *args, **kwargs):
         """Post event in MPF via async loop to prevent race conditions."""
-        self.loop.call_soon_threadsafe(partial(self.machine.events.post, *args, **kwargs))
+        self.loop.call_soon_threadsafe(partial(self.machine.events.post, event, *args, **kwargs))
 
     def set_machine_variable_in_mpf(self, name, value):
         """Set machine var in MPF via async loop to prevent race conditions."""
--- a/mpf/plugins/twitch/twitch_client.py
+++ b/mpf/plugins/twitch/twitch_client.py
@@ -1,19 +1,28 @@
 """IRC Chat Bot for monitoring a Twitch chatroom."""
 from functools import partial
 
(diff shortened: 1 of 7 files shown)


jabdoa2:
I added a basic test and config loading. Please have a look.

I guess we could port this to AioSimpleIRCClient (https://github.com/jaraco/irc/blob/master/irc/client_aio.py). Sending events and setting variables from another thread is not super safe. Should work most of the time but bad things may also happen. With an asyncio IRC client the thread would not be needed. Otherwise we need to change the calls slightly later.

mwseiden:
I can confirm that I get incredibly rare crashes on chat events. I know very little about asyncio, I don't have a quick answer on porting it.

jabdoa2:
Pushed a fix with separate threads. This is slightly slower but there shouldn't be too many events anyway. If you got time feel free to extend the test a bit. Currently it is doing almost nothing except making sure that the plugin loads.

mwseiden:
Enabled the plugin in mpfconfig.yaml, fixed some code errors, added a couple more tests. Chat messages were tested by actually connecting to Twitch IRC.

mwseiden:
Documentation PR: https://github.com/missionpinball/mpf-docs/pull/326

mpf/plugins/twitch/twitch_client.py
```diff
@@ -26,7 +26,10 @@ def __init__(self, machine, username, password, channel, loop):
         server = 'irc.chat.twitch.tv'
         port = 6667
         self.log.info('Connecting to ' + server + ' on port ' + str(port) + '...')
-        super().__init__([(server, port, 'oauth:' + password)], username, username)
+        super().__init__([(
+            server, port,
+            (password if password.lower().startswith('oauth:') else ('oauth:' + password))
```
jabdoa2, line 43: we could also have two parameters in config if that helps to make this more clear

mwseiden: Most people are going to have the `oauth:` prefix on their passwords since that is what the token generator produces. I handle both because I did not copy the prefix and other people are going to make the same mistake.

jabdoa2: ok makes sense

jabdoa2:
1. Do you want to merge it as it is or do you want to convert it to machine variables for credentials first?
2. Shutting down the thread gracefully seems to be missing in the irc library for some reason. This will probably prevent MPF from properly shutting down, right?

mwseiden:
1. Added an option to use machine vars instead of hard coding. I would love to get this merged in as soon as possible.
2. It shuts down just fine. I have not noticed any bad effects from not stopping the thread first.

jabdoa2:
Awesome. Thanks!

PR #1531, opened 2020-09-08
Fix subscription event. Add gift subs, raid, and command events.

mwseiden:
Twitch Integration v2:

General code improvements, shorter methods
Use usernotice command for subscriptions
Support gifted subscriptions
Add channel raid event
Add command event (line starts with ! or ?)

### DIFF
--- a/mpf/plugins/twitch/twitch_client.py
+++ b/mpf/plugins/twitch/twitch_client.py
@@ -170,7 +170,7 @@ class TwitchClient(SingleServerIRCBot):
         raid_user = tags.get('msg-param-displayName', user)
         raid_count = tags.get('msg-param-viewerCount', 0)
         self.set_machine_variable_in_mpf('twitch_last_raid_user', raid_user)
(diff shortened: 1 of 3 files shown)


mwseiden:
Documentation PR: https://github.com/missionpinball/mpf-docs/pull/328

mpf/plugins/twitch/twitch_client.py
```diff
@@ -152,13 +82,15 @@ def on_all_events(self, c, e):
     def do_command(self, e, cmd):
         """Handle a chat command (starts with ? or !)."""
         user = e.source.split('!')[0]
-        self.log.info('Received command: [' + user + '] ' + cmd)
+        self.log.info('Received command: [' + user + '] ' + cmd.lower())
```
jabdoa2, line 85: This would ideally be `self.log.info('Received command: [%s] %s', user, cmd.lower())`. This will safe CPU cycles when logs are not generated as string processing is rather expensive

jabdoa2:
+1 for the test. Thanks!

PR #1532, opened 2020-09-20
Allow shot.jump() calls to force a show refresh

avanwinkle:
This PR adds a new parameter to `shot.jump()` that allows the call to force an update of the shot profile show. This fixes a behavior that occurs when a shot's profile is programmatically changed and jumped to a new state while the shot is running. 

In most cases this works fine, but there's a gap because the shot uses the index of the profile's state to determine whether to call `self._update_show()`. In the case where the desired state of the new profile is the same as the state of the old profile, logic will prevent the show from updating and the previous profile's show will remain running.

This PR adds an override parameter `force_show` that instructs the Shot to call `self._update_show()` even if the previous profile state is the same as the new profile state.

### DIFF
--- a/mpf/devices/shot.py
+++ b/mpf/devices/shot.py
@@ -403,7 +403,7 @@ class Shot(EnableDisableMixin, ModeDevice):
     def _release_delay(self, switch):
         self.active_delays.remove(switch)
 
-    def jump(self, state, force=True):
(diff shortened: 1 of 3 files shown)


jabdoa2:
Looks fine to me. Can we have a simple test for it? Otherwise a future refactoring might remove it again.

PR #1535, opened 2020-10-10
Support block_on_flipper_cancel for Carousels

avanwinkle:
### Summary

This PR adds a new mode setting to carousel modes, called `block_events`.

If provided, this setting will prevent the carousel from moving to next/previous items after a blocking event is posted. Next/previous will only unblock after a release event is posted.

```
mode_settings:
  select_item_events: flipper_cancel
  next_item_events: s_flipper_right_inactive
  previous_item_events: s_flipper_left_inactive
  block_events: flipper_cancel
  release_events: both_flippers_inactive
```

### Use Cases
* When using the carousel to pick an item with sound or slide effects on change, selecting an item will trigger the next/prev sound after the selection because the mode is still active when the flippers are released.
* When nesting one carousel inside another (i.e. a submenu), selecting an item in the submenu will trigger a next/prev in the main menu, because the release event occurs after the submenu closes

In both of these cases, setting `block_events` on _flipper_cancel_ will prevent the carousel from rotating after the selection is made. Only after both flippers are released will the block be released, and a subsequent flipper press+release will trigger the carousel.

**Also** this PR adds config_spec support for implementing the same functionality on the TextInput widget in MC.

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -2005,6 +2005,7 @@ widgets:
         select_event: single|event_handler|sw_start
         abort_event: single|event_handler|sw_esc
         force_complete_event: single|event_handler|None
+        block_on_flipper_cancel: single|bool|false
         bold: single|bool_or_token|false
         italic: single|bool_or_token|false
         number_grouping: single|bool_or_token|false
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -2005,7 +2005,8 @@ widgets:
         select_event: single|event_handler|sw_start
         abort_event: single|event_handler|sw_esc
         force_complete_event: single|event_handler|None
-        block_on_flipper_cancel: single|bool|false
+        block_events: list|event_handler|None
+        release_events: list|event_handler|None
         bold: single|bool_or_token|false
         italic: single|bool_or_token|false
         number_grouping: single|bool_or_token|false
--- a/mpf/modes/carousel/code/carousel.py
+++ b/mpf/modes/carousel/code/carousel.py
@@ -80,7 +80,7 @@ class Carousel(Mode):
             self._register_handlers(
                 ["both_flippers_inactive", "flipper_cradle_release"],
                 self._block_disable)
-            
+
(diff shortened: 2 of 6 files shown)


mpf/modes/carousel/code/carousel.py
```diff
@@ -67,6 +72,15 @@ def mode_start(self, **kwargs):
         self._highlighted_item_index = 0
         self._update_highlighted_item(None)
 
+        # If set to block next/prev on flipper cancel, set those event handlers
+        if self._block_on_flipper_cancel:
+            self._register_handlers(
+                ["flipper_cradle", "flipper_cancel"],
```
jabdoa2: Those could also become an option

mpf/config_spec.yaml
```diff
@@ -2005,6 +2005,7 @@ widgets:
         select_event: single|event_handler|sw_start
         abort_event: single|event_handler|sw_esc
         force_complete_event: single|event_handler|None
+        block_on_flipper_cancel: single|bool|false
```
jabdoa2: I don't understand this line. How is text_input related to carousel?

mpf/modes/carousel/code/carousel.py
avanwinkle: Good call. It'd have to be a finite set of options, since the cradle/cancel have different cancellation events, but good plan.

mpf/config_spec.yaml
avanwinkle: Nothing :)

I have a similar implementation in MC for the text input to block on cancel, but MC uses the config_spec from MPF so it will require a change to this file. Thought I might as well take care of it in this PR.

avanwinkle:
Updated to fix a bug where restarting the mode wouldn't clear the blocked state, and added a test to prevent regressions.

Also made a small tweak to _FlasherPlayer_ that flashes in the lights with the `"on"` color rather than explicitly `"white"`.

jabdoa2:
Looks good to me. Feel free to merge it yourself when you are ready.

PR #1536, opened 2020-10-24
Support conditional events and fallback for random_event_player

avanwinkle:
This PR adds support for conditional events in `random_event_player` event lists.

For starters, the _Randomizer_ class now accepts `machine` and `template_type` as constructor arguments, and if provided, will generate conditional templates for the items in its randomization list. When fetching the next random item, if the items are conditional template instances, they will all be evaluated and only those that are true will be returned.

For flexibility, the _Randomizer_ class also gets a new config setting: `fallback_value`. If all the conditions failed and no items will be returned, the _Randomizer_ will return its fallback value if one is defined. If not, it will return `None`.

The _RandomEventPlayer_ is likewise updated to pass in the new constructor arguments and set a fallback event.

Tests forthcoming!

### DIFF
--- a/mpf/config_players/random_event_player.py
+++ b/mpf/config_players/random_event_player.py
@@ -46,12 +46,13 @@ class RandomEventPlayer(ConfigPlayer):
                 if settings['disable_random']:
                     self.machine.game.player[key].disable_random = True
 
-                self.machine.game.player[key].fallback_event = settings.get('fallback_event')
+                self.machine.game.player[key].fallback_value = settings.get('fallback_event')
 
             return self.machine.game.player[key]
 
         if key not in self._machine_wide_dict:
-            self._machine_wide_dict[key] = Randomizer(settings['events'], self.machine, template_type="event")
+            self._machine_wide_dict[key] = Randomizer(
+                settings['events'], self.machine, template_type="event")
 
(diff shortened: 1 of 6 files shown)


avanwinkle:
@jabdoa2 Curious about your thoughts on performance. I had the idea to, after the items list is built, check all the conditional template instances to see if they have conditions. If all conditions are `None`, i.e. none of the events are conditional, replace all the templates with the original strings and remove the `self._template_type` value.

In such a situation, we'd save the iteration to check for conditions when the random events played, at the cost of doing another iteration on construction. Not sure if that's worth it?

jabdoa2:
Our template code will only construct templates if there are conditions. Otherwise, it will emit Native types which do not perform any evaluation.

In general, I like to heavy lifting to startup time. That pays off most of the time. Also provides earlier error which is generally nice. Guess we could write a benchmark to see if this is worth the effort or not. There are some benchmarks for events already. If you want give them a try. Otherwise I would not worry too much. We can tune that once it becomes an issue.

avanwinkle:
Started down the path of reverting conditions if no events have them, but then decided it's probably not worth it. Random_event_players are most likely not spamming tons of events constantly and the conditional lookup for `None` is real fast. Seems good as it is.

PR #1539, opened 2020-10-26
Store and wrap exception message, add exception for show error

avanwinkle:
**Same as https://github.com/missionpinball/mpf-mc/pull/408**

I've been trying to improve the exception handling of MPF/MC as I work—whenever I get a crash that requires me to scroll through the callstack or look in the logs, I update the Exception chain to try and get the meaningful error into the final callstack message.

This PR also stores the initial message in the exception, so other wrappers (especially in config file and MC) can prepend their errors to the original and give a chain to identify the file >> block >> value that caused the error.

I think we should do a 50.4 release soon, so making a PR for the changes I've got so far.

### DIFF
--- a/mpf/assets/show.py
+++ b/mpf/assets/show.py
@@ -167,27 +167,24 @@ class Show:
                 total_step_time = -1
 
             # Now process show step actions
-            try:
-                self._process_step_actions(step, actions)
-            except Exception as e:
-                raise type(e)("An error occurred while processing show {}: '{}'".format(self.name, e)) from e
-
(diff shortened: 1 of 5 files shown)


avanwinkle (commented):
Test failure showed me that I was inadvertently wrapping Show-generated errors while trying to bubble up deeper config errors. Refactored _show.py_ to wrap and bubble deeper errors but maintain existing generated errors.

Did a little bit of touchup for more consistent formatting of the error strings too.

jabdoa2:
Awesome! Thanks !

PR #1541, opened 2020-11-16
Config player delay

avanwinkle:
This PR adds a new settings option to all ConfigPlayer-based classes: `delay`.

When provided, the `delay` option will schedule the play event to be posted some time in the future. The implementation does not currently support cancelling delayed events.

**Note:** the delay is not automatic for all ConfigPlayers. Instead, a ConfigPlayer class may choose to call `self.check_delayed_play()` as part of its play handler, as is done here https://github.com/missionpinball/mpf-mc/pull/409 for SoundPlayer. Other players may follow suit, as needed.

This PR also has some minor bugfixes and tweaks:
* FlasherPlayer supports a `color` setting to allow RGB flashers to flash in color
* Utility functions dictionary merge fixes a (presumed) typo where index access was attempted on a dict
* Shot profiles can handle an exception when changing states away from an index that no longer exists.

This last one is an edge case, but one I've been fighting. When a shot changes state, it posts the name of the new state and the old state. If custom python code has changed the profile of the shot and the new profile doesn't have as many states, *and* the shot was previously in a state higher than the number in the new profile, the attempt to fetch the old name would throw an exception. This PR gracefully handles that exception and just posts "unknown" as the old state name.


mpf/core/utility_functions.py
```diff
@@ -242,7 +242,7 @@ def dict_merge(a, b, combine_lists=True) -> dict:
             elif k in result and isinstance(result[k], dict):
                 result[k] = Util.dict_merge(result[k], v, combine_lists)
             elif k in result and isinstance(result[k], list):
-                if isinstance(v, dict) and v[0] == dict(_overwrite=True):
+                if isinstance(v, list) and v[0] == dict(_overwrite=True):
```
avanwinkle: I don't remember what situation I was in that kept crashing on this error, but attempting to access `v[0]` when `v` is a dict... seems like just a typo. If I'm overlooking something here, please let me know!

avanwinkle:
I reverted my change to `utility_functions.py` because it broke a lot of tests. When I rediscover the issue I had around that, I'll do a more thorough investigation and a separate fix.

PR #1542, opened 2020-11-25
Add 'hits' to counter kwargs, fix 'remaining' kwarg for downward counts

avanwinkle:
This PR adds a new kwarg `hits` to the LogicBlock.Counters hit events, and changes the calculation of the `remaining` kwarg.

Currently, the `remaining` value is always calculated as the `complete_value - self.value`, which works great for counting _up_ but not for counting _down_. This PR changes the logic such that downward counters use `self.value - complete_value` for the remaining amount.

The new `hits` value is the balance between the total value and the remaining. For upward counters it's `self.value - start_value` and for downward counters it's `start_value - self.value`.

### DIFF
--- a/mpf/devices/logic_blocks.py
+++ b/mpf/devices/logic_blocks.py
@@ -460,7 +460,12 @@ class Counter(LogicBlock):
                 "count": self.value
             }
             if count_complete_value is not None:
-                args['remaining'] = count_complete_value - self.value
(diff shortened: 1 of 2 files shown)


avanwinkle:
Builds failed on Python 3.5 platforms with the expected `Yield inside async function`, but passed on all other platforms.

PR #1545, opened 2020-12-07
DigitalScoreReel Device

avanwinkle:
The **DigitalScoreReel** device attaches to an event that posts a value, and then re-posts an event with the value broken up into the data necessary for a score reel. The configuration looks like this:

```
digital_score_reels:
  player_score:
    reel_count: 4
    include_player_number: true
    frames:
      - character: 1
        frame: 9
      - character: 2
        frame: 18
      - character: 3
        frame: 27
      - character: 4
        frame: 36
      - character: 5
        frame: 45
      - character: 6
        frame: 54
      - character: 7
        frame: 63
      - character: 8
        frame: 72
      - character: 9
        frame: 81
      - character: 0
        frame: 90
```

In this example, the DigitalScoreReel will listen for _player_score_ events and then parse the `value` keyword arg from the event. For each reel (in this case, four reels), the DigitalScoreReel will post a new event with the corresponding frame that the reel image should move to.

```
Event: ======'player_score'====== Args={'value': 5280, 'prev_value': 4280, 'change': 1000, 'player_num': 1}
Event: ======'score_reel_player_score_player1'====== Args={'1': '90', '10': '72', '100': '18', '1k': '45'}
```

In the above example, the thousands character "5" corresponds to the image frame 45, so the event includes `'1k': '45'`, and the hundreds character "2" corresponds to the image frame 18, so the event includes `'100': '18'`. All the values are provided as strings, for compatibility with existing patterns (https://github.com/missionpinball/mpf-mc/blob/dev/mpfmc/uix/widget.py#L461)

* The `reel_count` config option ensures that the DigitalScoreReel will zero-pad reels larger than the player's score.
* The `include_player_number` config option, if true, will append `_player(num)` to the posted event.

This implementation is designed to be flexible. Reels may include any characters (not just numbers), and may be used for events besides the score (e.g. ball count, jackpot value, et cetera). 

This PR also includes the _config_spec.yaml_ updates necessary to support image end_frame, frame_skips, and image_templates and provided in https://github.com/missionpinball/mpf-mc/pull/411


jabdoa2:
Does this visualize score reels (i.e. a real score reel with virtual coils) or is this independent? I guess the latter right?

avanwinkle:
@jabdoa2 It does not virtualize real reels, just creates events for a media controller to drive image-based reels. So maybe "virtual" isn't the best name for it. ImageScoreReel? DigitalScoreReel?

jabdoa2:
I like digital_score_reels. Might make it a bit more clear. Nice work!

avanwinkle:
Just pushed a change to make the score reel more flexible. Instead of keying the kwargs on the score character (`100`, `1k`, `100m`), it instead uses the left-to-right position (starting at `1`). This way using non-score implementations don't have to mentally convert the reel positions to score values, and it's more intuitive.

Also added a new config option `start_value`, which is the character that the reel will pad with. Defaults to zero, which was the hard-coded behavior before.

jabdoa2:
LGTM. Feel free to merge when you are ready.

PR #1546, opened 2020-12-18
Dev 7

avanwinkle:
This PR increments the MPF dev version to 7, for the prerelease of the digital score reels.

Also has a log reduction for assets and a config spec fix.


jabdoa2:
👍

PR #1549, opened 2021-01-02
Call mode_stop() before removing devices from mode

avanwinkle:
This PR makes a minor change in the Mode shutdown behavior.

It moves the user-overridable class method `mode_stop()` to be called *before* the event handlers and devices are removed from the mode, instead of after.

With this change, the `mode_stop()` method can access devices in the mode including shots, counters, and timers, and trigger behavior based on their state. Before this change, those devices are unavailable.

### DIFF
--- a/mpf/core/mode.py
+++ b/mpf/core/mode.py
@@ -366,12 +366,14 @@ class Mode(LogMixin):
 
     def _mode_stopped_callback(self, **kwargs) -> None:
         del kwargs
+
+        # Call the mode_stop() method before removing the devices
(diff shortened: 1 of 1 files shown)


jabdoa2:
Maybe we should also improve the naming? mode_will_stop and mode_stopped. That would make it more clear when this method is called

avanwinkle:
Poked around with migrating `mode_start()` to `mode_started()` and `mode_stop()` to `mode_will_stop()` (and also a new `mode_stopped()`), but the rabbit hole is very deep. Probably worth doing, but there are a lot of dependencies to track down.

PR #1550, opened 2021-01-03
Initial attempt at FAST wiring.

MarkRdgOx:


mpf/commands/game.py
```diff
@@ -147,6 +147,7 @@ def __init__(self, mpf_path, machine_path, args):
                             action="store_true", dest="no_sound", default=False)
 
         self.args = parser.parse_args(args)
+        print(self.args, flush=True)
```
jabdoa2: Can you remove this again?

jabdoa2:
LGTM. Do you want to merge this as it is? Or do you want to continue working inside this branch?

MarkRdgOx:
This is up to you - you mentioned that you would like to have some starting point and this is that. I have further ideas to develop but my ability to work further may slow down at this point since work is restarting and especially because I will be involved in managing the consequences of the UK's Shelter in Place order.

jabdoa2:
Good luck with that! I will merge this and we can iterate based on that. Thanks for the effort!

PR #1558, opened 2021-03-26
Fix error related to show pools

atummons:
Set the param when def show pools, and returning show pools to include
start_running=True, and start_running value returned.  This allows the
show pool to run, as it previously threw an error that the param was
missing.  Now if the show_player sets start_running to false, it
behaves.  But defaults, like show_player, to start running.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -27,6 +27,7 @@ their spare time, unpaid, for the love of pinball!
  * Sean Irby <sean.t.irby@gmail.com>
  * Mark Seiden <thearrrrrcade@gmail.com>
  * Andrew Burch <andrewb@enteryourinitials.com>
+ * Avery Tummons
 
 MPF is inspired by pyprocgame and skeletongame which were written by:
 
--- a/mpf/assets/show.py
+++ b/mpf/assets/show.py
@@ -32,10 +32,10 @@ class ShowPool(AssetPool):
         return '<ShowPool: {}>'.format(self.name)
 
     # pylint: disable-msg=too-many-arguments
-    def play_with_config(self, show_config: ShowConfig, start_time=None, start_callback=None, stop_callback=None,
+    def play_with_config(self, show_config: ShowConfig, start_time=None, start_running=True, start_callback=None, stop_callback=None,
                          start_step=None) -> "RunningShow":
         """Play asset from pool with config."""
-        return self.asset.play_with_config(show_config, start_time, start_callback, stop_callback, start_step)
+        return self.asset.play_with_config(show_config, start_time, start_running, start_callback, stop_callback, start_step)
 
     # pylint: disable-msg=too-many-arguments
     # pylint: disable-msg=too-many-locals
--- a/mpf/assets/show.py
+++ b/mpf/assets/show.py
@@ -32,10 +32,11 @@ class ShowPool(AssetPool):
         return '<ShowPool: {}>'.format(self.name)
 
     # pylint: disable-msg=too-many-arguments
-    def play_with_config(self, show_config: ShowConfig, start_time=None, start_running=True, start_callback=None, stop_callback=None,
-                         start_step=None) -> "RunningShow":
+    def play_with_config(self, show_config: ShowConfig, start_time=None, start_running=True, start_callback=None,
+                         stop_callback=None, start_step=None) -> "RunningShow":
         """Play asset from pool with config."""
-        return self.asset.play_with_config(show_config, start_time, start_running, start_callback, stop_callback, start_step)
+        return self.asset.play_with_config(show_config, start_time, start_running, start_callback, stop_callback,
+                                           start_step)
 
     # pylint: disable-msg=too-many-arguments
     # pylint: disable-msg=too-many-locals
--- a/mpf/tests/machine_files/shows/config/test_show_pools.yaml
+++ b/mpf/tests/machine_files/shows/config/test_show_pools.yaml
@@ -166,6 +166,8 @@ show_player:
         color: blue
         color1: green
         color2: yellow
+  stop_pool_random:
+    pool_random: stop
   play_pool_sequence:
     pool_sequence:
       show_tokens:
--- a/mpf/tests/machine_files/shows/config/test_show_pools.yaml
+++ b/mpf/tests/machine_files/shows/config/test_show_pools.yaml
@@ -1,0 +1,192 @@
+#config_version=5
+
+lights:
+    led_01:
+        number: 0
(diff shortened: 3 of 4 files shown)


jabdoa2:
Thanks for the change. I guess we also need a test for this. Do you want to add that as well? I can also do that but it might take me a while.

atummons:
I have no idea how to write a test!  Absolutely willing to do so, but do you any pointers?  There is a test that exist already on document loads of shows, and includes show pools.  
mpf/mpf/tests/machine_files/asset_manager/config/test_asset_loading.yaml 

So that uses show_pools there to show they load...I think.  But clearly it wasn't checking as they didn't really work.

Would I just extend the test_show.py file?  Just go to the end, and call upon some shows in a show group, and run the similar checks to step through and ensure the light is the color it should be?  Might take a review or two to get it clean, as I have never done this.  But I am willing to learn!

jabdoa2:
Config looks good. I guess you could copy the show test: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_Shows.py. Adjust it accordingly (reference your config). You can either start your show via show_player by using `self.post_event("your_event")` or using `self.machine.shows["your_show“]`. The former is more representive so I would go with that. Than assert that the show is running or that one of the side effects (LEDs turning on) is observable. Makes sense? Feel free to ask anything or write me directly.

atummons:
That part makes sense. How do I detect which show is running?  I look at random event, but that seemed different, as they didn’t care if it played. Is there a specific event I can query?

So I play pool A. It picks show 3. I need a way of saying if show 1... if show 2...  then I can adapt the existing stuff to ensure it ran 1 correct by pointing to that def, and then verifying a pass before running pool A again.

Does that make sense? Or am I going about it the wrong way?

jabdoa2:
> That part makes sense. How do I detect which show is running? I look at random event, but that seemed different, as they didn’t care if it played. Is there a specific event I can query? So I play pool A. It picks show 3. I need a way of saying if show 1... if show 2... then I can adapt the existing stuff to ensure it ran 1 correct by pointing to that def, and then verifying a pass before running pool A again. Does that make sense? Or am I going about it the wrong way?

Yeah makes sense. There are multiple ways:
- Check that the side effect of one of the three shows is visible. This could be one out of three LEDs turning on or that one out of three events is posted
- Alternatively, we could mock the random number generator (like here: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_MatchMode.py#L46). That way random will become deterministic in tests.
- We could also check the running shows directly: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_Shows.py#L120

atummons:
Jan,

I am getting frustrated here because I just don't get this.  So basically I took what Anthony had in RandomEventPlayer, and ran with that.  But then his MPFGameTestCase class didn't have the ability to AssertLightColor (as that is MPFTestCase).  But when I tried to call that class I got assertion errors that it didn't have the info for that variable.

I'm not sure how to reinvent the wheel to make it run the full test cases that he had going on the test_shows.py file.  I need it to generate 20 instances of a given event: play_pool_random for example.  Then I need to check for each one, which light is lit up red.  Then I can post the event led_01, led_02, etc.  That is the event that I then check against the 20 and the method run.  Did it go through all before repeating, did it select a unique one, did they go in order, etc.

Not asking you to do it by any stretch, as I want to learn.  But not sure how to get it to understand what functions I want to run.  The assertLightColor is needed for, as there isn't an event (that I know of) that shows this is the show that is running.  Looking at shows L120, it seems that they are telling it what it is.  But I don't know what it is.

I'm starting to feel that the code fix was easy, but the test case is something far more than I should have bit off given that I have never written a test before!  But trial by fire, I guess.

[test_ShowPools.txt](https://github.com/missionpinball/mpf/files/6226097/test_ShowPools.txt)

jabdoa2:
I rebased and added a test for pool_random. Mocked the random generator and asserted that the show is actually running.

> I am getting frustrated here because I just don't get this. So basically I took what Anthony had in RandomEventPlayer, and ran with that. But then his MPFGameTestCase class didn't have the ability to AssertLightColor (as that is MPFTestCase). But when I tried to call that class I got assertion errors that it didn't have the info for that variable.

The method `self.assertLightColor(light_name, color)` exists in any mpf test case and seems to work here. Not sure why it does not work for you. Please post the full error. In general, an IDE such as PyCharm will help with code completion here.

atummons:
Thanks for writing the test for me.  I appreciate it!  So here is the code chunk I have (ripped off from RandomEventPlayer (see attachment).
[test_ShowPools.txt](https://github.com/missionpinball/mpf/files/6230708/test_ShowPools.txt)

And here is the error:
======================================================================
ERROR: test_player_tests (mpf.tests.test_ShowPools.TestShowPoolsGame)
----------------------------------------------------------------------
  File "C:\mpf\mpf\tests\test_ShowPools.py", line 71, in test_player_tests
    tester.run_all_tests()
  File "C:\mpf\mpf\tests\test_ShowPools.py", line 87, in run_all_tests
    test()
  File "C:\mpf\mpf\tests\test_ShowPools.py", line 105, in _test_show_pool_rfn
    if self.assertLightColor("led_01", "red"):
AttributeError: 'TestShowPoolBase' object has no attribute 'assertLightColor'
----------------------------------------------------------------------

I can't figure out how to make that class have the assertLightColor feature.  Because I am doing that to check the show.  Is led_01 red, if so you picked show 1.  Post that event, then I can track the events per the existing RandomEventPlayer code, and verify that all features work as expected.  I have tried passing it into the object, calling from within the class back to the other class, etc. and can't seem to figure it out.

I am just using Notepad++.  Is PyCharm much better?  I don't know how to debug this, or even just run a single test.  After every change, I am running the full test, which takes about a minute.  So I am sure I'm not doing this efficiently either!

Thanks!

jabdoa2:
> I can't figure out how to make that class have the assertLightColor feature. Because I am doing that to check the show. Is led_01 red, if so you picked show 1. Post that event, then I can track the events per the existing RandomEventPlayer code, and verify that all features work as expected. I have tried passing it into the object, calling from within the class back to the other class, etc. and can't seem to figure it out.
> 
> I am just using Notepad++. Is PyCharm much better? I don't know how to debug this, or even just run a single test. After every change, I am running the full test, which takes about a minute. So I am sure I'm not doing this efficiently either!

I had a look at you code. It looks like you are instantiating a second class `TestShowPoolBase` which does not inherit from the MPF test case. This class does not have all the attribute or methods from the usual test classes. I guess you probably just want to use a method instead.

PR #1559, opened 2021-04-03
Support FAST RGB Segment Displays

avanwinkle:
This PR adds support for the FAST colored segment displays.

fast-segment

### ColorSegmentDisplay Device

This new device extends the `SegmentDisplayPlatformInterface` to a new `ColorSegmentDisplayPlatformInterface` which supports color setting. 

Note: the segment display _must_ be run at baud rate 230400.

```
hardware:
  platform: fast
  driverboards: fast
  segment_displays: fast

fast:
    driverboards: fast
    ports: /dev/tty.usbserial-14100
    baud: 230400
```

### Segment Display Player

fast-segment-multicolor

A `color:` payload can be submitted with any `segment_display_player:` command, or the `action: set_color` payload will update the color without affecting the text.

A single color value can be provided for all seven characters, or a list of colors for each character.

```
segment_display_player:
    timer_seven_seg_display_timer_tick.1{ticks%3==0}:
        7_seg_display_1:
            text: "{machine.high_score:>7}"
            color: 000033, 003300, 330000, 330033, 003333, 333300, 333333
    timer_seven_seg_display_timer_tick.2{ticks%3==1}:
        7_seg_display_1:
            color: 3f0099
            action: set_color
    timer_seven_seg_display_timer_tick.3{ticks%3==2}:
        7_seg_display_1:
            text: "ABCDEFG"
            color: f12711, f23e12, f25414, f36b15, f48216, f49818, f5af19
```

### Other Adjustments

* Not all serial devices support low latency mode, this PR adds an error catch so these platforms won't crash
* The FAST segment displays require a specific baud rate. This PR adds support for per-serial-bus baud rates on FAST hardware
* This PR sets explicit serial communication terms `EIGHTBITS, PARITY_NONE, STOPBITS_ONE`. If this is overly broad and should be configurable per-connection, please let me know.

Huge thanks to @ecurtz for the foundational work on this one, I'm just the touchup guy here :)

### DIFF
--- a/mpf/config_players/segment_display_player.py
+++ b/mpf/config_players/segment_display_player.py
@@ -61,8 +61,14 @@ class SegmentDisplayPlayer(DeviceConfigPlayer):
                 display.set_flashing(FlashingType.FLASH_MATCH)
             elif action == "no_flash":
                 display.set_flashing(FlashingType.NO_FLASH)
+            elif action == "set_color":
+                # Setting a color makes no other changes to the display
+                pass
             else:
                 raise AssertionError("Invalid action {}".format(action))
+
+            if s['color']:
+                display.set_color(s['color'])
 
     def _remove(self, instance_dict, key, display):
         if key in instance_dict[display]:
--- a/mpf/platforms/base_serial_communicator.py
+++ b/mpf/platforms/base_serial_communicator.py
@@ -68,7 +68,7 @@ class BaseSerialCommunicator:
             try:
                 serial.set_low_latency_mode(True)
             except NotImplementedError:
-                self.log.debug("Could not set %s to low latency mode.", port)
+                self.log.info("Could not set %s to low latency mode.", port)
 
         # defaults are slightly high for our usecase
         self.writer.transport.set_write_buffer_limits(2048, 1024)
--- a/mpf/platforms/base_serial_communicator.py
+++ b/mpf/platforms/base_serial_communicator.py
@@ -2,7 +2,7 @@
 from typing import Optional
 
 import asyncio
-from serial import SerialException
+from serial import SerialException, EIGHTBITS, PARITY_NONE, STOPBITS_ONE
 
 from mpf.core.utility_functions import Util
 
@@ -49,7 +49,8 @@ class BaseSerialCommunicator:
         while True:
             try:
                 connector = self.machine.clock.open_serial_connection(
-                    url=port, baudrate=baud, limit=0, xonxoff=xonxoff)
+                    url=port, baudrate=baud, limit=0, xonxoff=xonxoff,
+                    bytesize=EIGHTBITS, parity=PARITY_NONE, stopbits=STOPBITS_ONE)
                 self.reader, self.writer = await connector
             except SerialException:
                 if not self.machine.options["production"]:
@@ -64,7 +65,10 @@ class BaseSerialCommunicator:
 
         serial = self.writer.transport.serial
         if hasattr(serial, "set_low_latency_mode"):
-            serial.set_low_latency_mode(True)
+            try:
+                serial.set_low_latency_mode(True)
+            except NotImplementedError:
+                self.log.debug("Could not set %s to low latency mode.", port)
 
         # defaults are slightly high for our usecase
         self.writer.transport.set_write_buffer_limits(2048, 1024)
--- a/mpf/platforms/fast/fast.py
+++ b/mpf/platforms/fast/fast.py
@@ -38,7 +38,7 @@ class FastHardwarePlatform(ServoPlatform, LightsPlatform, DmdPlatform,
 
     """Platform class for the FAST hardware controller."""
 
-    __slots__ = ["dmd_connection", "net_connection", "rgb_connection", "leg_connection", "seg_connection",
+    __slots__ = ["dmd_connection", "net_connection", "rgb_connection", "seg_connection",
                  "serial_connections", "fast_leds", "fast_commands", "config", "machine_type", "hw_switch_data",
                  "io_boards", "flag_led_tick_registered", "_watchdog_task", "_led_task"]
 
@@ -70,7 +70,6 @@ class FastHardwarePlatform(ServoPlatform, LightsPlatform, DmdPlatform,
         self.dmd_connection = None
         self.net_connection = None
         self.rgb_connection = None
-        self.leg_connection = None
         self.seg_connection = None
         self._watchdog_task = None
         self._led_task = None
@@ -123,13 +122,6 @@ class FastHardwarePlatform(ServoPlatform, LightsPlatform, DmdPlatform,
                 self.dmd_connection.remote_processor,
                 self.dmd_connection.remote_model,
                 self.dmd_connection.remote_firmware)
-        if not self.leg_connection:
-            infos += "No connection to the Legacy Controller.\n"
-        else:
-            infos += "Legacy Controller: {} {} {}\n".format(
-                self.leg_connection.remote_processor,
-                self.leg_connection.remote_model,
-                self.leg_connection.remote_firmware)
         if not self.seg_connection:
             infos += "No connection to the Segment Controller.\n"
         else:
@@ -195,8 +187,6 @@ class FastHardwarePlatform(ServoPlatform, LightsPlatform, DmdPlatform,
             self.rgb_connection.writer.write(b'BL:AA55\r')  # reset CPU using bootloader
         if self.dmd_connection:
             self.dmd_connection.writer.write(b'BL:AA55\r')  # reset CPU using bootloader
-        if self.leg_connection:
-            self.leg_connection.writer.write(b'BL:AA55\r')  # reset CPU using bootloader
         # if self.seg_connection:
         #     self.seg_connection.writer.write(b'BL:AA55\r')  # reset CPU using bootloader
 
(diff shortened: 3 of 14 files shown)


mpf/platforms/base_serial_communicator.py
```diff
@@ -64,7 +65,10 @@ async def _connect_to_hardware(self, port, baud, xonxoff=False):
 
         serial = self.writer.transport.serial
         if hasattr(serial, "set_low_latency_mode"):
-            serial.set_low_latency_mode(True)
+            try:
+                serial.set_low_latency_mode(True)
+            except NotImplementedError:
+                self.log.debug("Could not set %s to low latency mode.", port)
```
jabdoa2: I would go with info or warning here. This should work for FTDI at least. Might be a permission issue (i.e. only works as root)

mpf/platforms/fast/fast.py
```diff
@@ -177,6 +187,8 @@ def stop(self):
             self.rgb_connection.writer.write(b'BL:AA55\r')  # reset CPU using bootloader
         if self.dmd_connection:
             self.dmd_connection.writer.write(b'BL:AA55\r')  # reset CPU using bootloader
+        # if self.seg_connection:
+        #     self.seg_connection.writer.write(b'BL:AA55\r')  # reset CPU using bootloader
```
jabdoa2: Please remove commented code. We got version control for that ;-)

mpf/platforms/fast/fast.py
```diff
@@ -735,6 +759,18 @@ def configure_dmd(self):
 
         return FASTDMD(self.machine, self.dmd_connection.send)
 
+    async def configure_segment_display(self, number: str, platform_settings) -> ColorSegmentDisplayPlatformInterface:
```
jabdoa2: Guess the signature should be `-> FASTSegmentDisplay`

mpf/platforms/interfaces/segment_display_platform_interface.py
```diff
@@ -79,3 +79,16 @@ def set_text(self, text: str, flashing: FlashingType) -> None:
     def _set_text(self, text: str) -> None:
         """Set a text to the display."""
         raise NotImplementedError
+
+
+class ColorSegmentDisplayPlatformInterface(SegmentDisplayPlatformInterface):
```
jabdoa2: Not sure if we want this as segment_display_player calls this unconditionally. What do you think? Either we add this to all segment displays (which probably would not hurt) or we need to check in the player

mpf/platforms/virtual.py
```diff
@@ -324,6 +324,10 @@ def set_text(self, text: str, flashing: FlashingType):
         self.text = text
         self.flashing = flashing
 
+    def set_color(self, colors: any) -> None:
```
jabdoa2, line 328: According to the base class this method should not be needed.

mpf/config_players/segment_display_player.py
```diff
@@ -61,9 +61,15 @@ def play(self, settings, context, calling_context, priority=0, **kwargs):
                 display.set_flashing(FlashingType.FLASH_MATCH)
             elif action == "no_flash":
                 display.set_flashing(FlashingType.NO_FLASH)
+            elif action == "set_color":
```
jabdoa2, line 64: I first was a bit confused here. So we always set the color? What is the purpose of the command then? Do we want to move `display.set_color(s['color'])` into this method?

mpf/platforms/fast/fast_segment_display.py
```diff
@@ -0,0 +1,49 @@
...
+        self.serial.send(('PC:{},{}').format(
+            self.hex_id, colors))
+
+    def _delayed_write(self, text_cmd):
+        """Debugging method for reduced serial communication speed."""
+        self.serial.platform.debug_log(text_cmd)
+        while (len(text_cmd) > 0):
+            self.serial.writer.write(text_cmd[0:1].encode())
+            text_cmd = text_cmd[1:]
+            sleep(0.001)
```
jabdoa2: Sleep in async code is dangerous.

mpf/platforms/fast/fast_segment_display.py
```diff
@@ -0,0 +1,49 @@
...
+        """Set display color."""
+        self.serial.platform.info_log("Color: {}".format(colors))
+        if len(colors) == 1:
+            colors = (RGBColor(colors[0]).hex + ',') * 7
+        else:
+            colors = ','.join([RGBColor(color).hex for color in colors])
+        self.serial.send(('PC:{},{}').format(
+            self.hex_id, colors))
+
+    def _delayed_write(self, text_cmd):
```
jabdoa2: If this is debug code only please remove it.

mpf/platforms/fast/fast_serial_communicator.py
```diff
@@ -121,7 +122,7 @@ async def _identify_connection(self):
             self.remote_processor, self.remote_model, self.remote_firmware = (
                 msg[3:].split())
         except ValueError:
```
jabdoa2: Can we have an unit test for this part? I would like a test where we use the fallback path.

jabdoa2:
Thanks for the PR guys! Some small nags but generally looks good. Guess we want some unit tests to prevent stuff from breaking over time. As you know if it ain't tested consider it broken ;-).

mpf/config_players/segment_display_player.py
avanwinkle: Originally the `set_color()` method was called here, but that required *two* player calls to set both the text and the color. I imagine that most use cases will want color and text to be set at the same time, so I moved the `set_color()` out below.

But if somebody wants to change the color without affecting what's displayed, the `set_color` action will allow that.

mpf/platforms/fast/fast_serial_communicator.py
avanwinkle: I would even more like a firmware update so the segment display returns the same formatted response as the other platforms :) 

But yes, I can add one

ecurtz: The segment controller should be temporarily returning "ID:SEG FP-DSP-2050-1 00.50\r" which I believe matches the existing FAST stuff, but if not I can certainly modify it.

mpf/platforms/base_serial_communicator.py
ecurtz: I was getting an exception on my M1 Mac for this, which was the only reason I added the handler. The serial descriptions says it is supported but it fails when actually set.

mpf/platforms/fast/fast_segment_display.py
ecurtz: _delayed_write is debug only although currently Anthony requires it for his HW, should probably just stash it locally and put the regular send code in for the PR.

mpf/config_players/segment_display_player.py
jabdoa2: Maybe we could call set color on `set_color` and `add`? That would be consistent.

jabdoa2:
I will implement this for light_segment_displays as well. The interface fits nicely for those as well. Just waiting for you to finish. Will create a PR afterwards.

mpf/platforms/base_serial_communicator.py
jabdoa2: I think somebody from the OPP folks saw that error on Linux as well. They said it worked when they ran MPF with sudo but not as user. So it is probably a good idea not to crash here but show a notice. I think they got some OSError "Not supported" or something totally confusing.

mpf/platforms/fast/fast_serial_communicator.py
avanwinkle: Yeah, looks like that fallback is not needed. I've removed it :)

mpf/platforms/base_serial_communicator.py
ecurtz: I thought that's what it was doing, but I'm not familiar with Python exception handling. Do you mean it should be using a different message than the log.debug() ? with the except: block it runs ok on my machine where the exception was halting it previously.

ecurtz:
Any unclear stuff or problems with the return types or overrides is 100% me being unfamiliar with Python and copy/pasting from the wrong example, rather than being intentional so please go ahead and fix them, no need to check with me.

avanwinkle:
@jabdoa2 To make prospector happy, I implemented empty `set_color()` methods on the P-Roc and LISY devices, so color is now possible to configure on any segment display. I got rid of the `ColorSegmentDisplay` device, so that's one less thing to worry about.

Thanks for the great feedback!

mpf/config_players/segment_display_player.py
avanwinkle: I foresee a situation like this:
```
segment_display_player:
  mode_hurryup_started:
    seg_display_0:
      action: set_color
      color: orange
  hurryup_almost_finished:
      seg_display_0:
        action: flash
        color: red
  mode_hurryup_stopped:
      seg_display_0:
        text: "Time Out"
        color: white
```

My goal was to abstract the color from any other behavior, so any action the user wants to make can change the color at the same time without an extra command. 

But I've also never used a segment display before, so I'm just making things up as I go 😛

mpf/platforms/virtual.py
avanwinkle: I've updated the base class to be an abstract method and consolidated the ColorSegmentDisplay into the SegmentDisplay. Just did a check and now the virtual mode is happy with this declaration.

PR #1562, opened 2021-04-15
Segment display emulator

qcapen:
PR for a new virtual segment display emulator (also related PRs for MPF-MC and docs). Uses virtual hardware platform segment display to update new segment_display_emulator widget in the MC. Supports the new color feature added for the Fast segment displays. The segment display widgets can be updated using the segment_display_player.

Still have some more advanced features to add (transition effects, individual character colors), but wanted to get this pulled into the code base first.


mpf/platforms/virtual.py
```diff
@@ -311,23 +320,41 @@ class VirtualSegmentDisplay(SegmentDisplayPlatformInterface):
...
+    def __init__(self, number, platform_options, machine) -> None:
         """Initialise virtual segment display."""
         super().__init__(number)
+        self.machine = machine
         self.text = ''
         self.flashing = FlashingType.NO_FLASH
-        self.colors = 'FFFFFF'
+        self.colors = [RGBColor('FFFFFF')]
+        self.platform_options = platform_options
+        if self.platform_options and 'post_update_events' in self.platform_options:
```
jabdoa2: Why do we want to post events when updating a segment_display? That sounds like a job for event_player to me.

mpf/platforms/virtual.py
```diff
@@ -311,23 +320,41 @@ class VirtualSegmentDisplay(SegmentDisplayPlatformInterface):
...
     """Virtual segment display."""
 
-    __slots__ = ["text", "flashing", "colors"]
+    __slots__ = ["text", "flashing", "platform_options", "colors", "machine", "post_update_events"]
 
-    def __init__(self, number) -> None:
+    def __init__(self, number, platform_options, machine) -> None:
         """Initialise virtual segment display."""
         super().__init__(number)
+        self.machine = machine
```
jabdoa2, line 323: I assume this only works when you use the virtual platform, right?

What about this: A plugin (or device) just registers a handler on the segment_display device. This way we could mirror any segment display to MC. That way it would also work with real hardware attached.

mpf/config_spec.yaml
```diff
@@ -1294,15 +1296,21 @@ segment_displays:
...
     __valid_in__: machine, mode, show
     __type__: config_player
     priority: single|int_or_token|0
     text: single|str|None
-    color: list|color|ffffff
+    color: list|color|None
     action: single|enum(add,remove,flash,no_flash,flash_match,set_color)|add
     key: single|str|None
     expire: single|ms_or_token|None
+    platform_options: single|dict|None
```
jabdoa2: Why do we need this in the player? Could we move platform_options (or platform_setting how we call it in most places I think) to the segment_display (simiar to coils/switches)?

mpf/config_spec.yaml
```diff
@@ -954,6 +951,11 @@ mpf:
 mpf-mc:
     __valid_in__: machine                           # todo add to validator
     __type__: config
+mpfmc_segment_display_emulator:
+    __valid_in__: machine
+    __type__: config
+    bcp_connection: single|str|local_display
+    debug: single|bool|false
```
jabdoa2: I would expect a `source_display: single|machine(segment_displays)|` here

mpf/platforms/virtual.py
qcapen: This was my not so elegant solution to communicating with the MC. My stumbling point was that BCP was not setup and connected when the hardware was being setup so this was an easy way to solve that timing issue. I'm happy to get rid of this and send the trigger using bcp directly bypassing the need for any event generation.

mpf/config_spec.yaml
qcapen: The thought was to send addition commands to the display at runtime (such as transition effects). Would it be better to add those to the segment_display_player spec directly and then allow other platform implementations to ignore them if not supported (like color)? I'm happy to handle it that way instead. Setting them in the segment_display won't work since it is a runtime option related to the new text.

mpf/platforms/virtual.py
qcapen: I like that idea, although I could use a little help in how this would work. What would trigger the plugin getting loaded (I suppose some config option relating to the segment display)? The plugin would need to know the bcp connection name and then generate the proper triggers when the segment display text or other parameters were updated (hook into the segment_display set_text and set_color methods?).

mpf/config_spec.yaml
qcapen: This section is not used (I forgot to remove it from one of my previous attempts at a dedicated virtual segment display platform). I will fix.

mpf/config_spec.yaml
qcapen: I also could add the transition effects directly to the segment display player since it would then be available to all hardware displays as well...  This is probably a better idea anyway. I will remove the platform_options.

mpf/platforms/virtual.py
qcapen: I think I have an idea of how to implement this after studying the other mpf plugins. I'll give it a shot. Thanks.

mpf/platforms/virtual.py
```diff
@@ -311,21 +315,23 @@ class VirtualSegmentDisplay(SegmentDisplayPlatformInterface):
 
     """Virtual segment display."""
 
-    __slots__ = ["text", "flashing", "colors"]
+    __slots__ = ["text", "flashing", "platform_options", "colors", "machine", "post_update_events"]
 
-    def __init__(self, number) -> None:
+    def __init__(self, number, platform_options, machine) -> None:
```
jabdoa2: Guess we should remove this again right?

mpf/config_spec.yaml
```diff
@@ -1292,17 +1289,21 @@ segment_displays:
...
     __valid_in__: machine, mode, show
     __type__: config_player
     priority: single|int_or_token|0
     text: single|str|None
-    color: list|color|ffffff
+    color: list|color|None
     action: single|enum(add,remove,flash,no_flash,flash_match,set_color)|add
     key: single|str|None
     expire: single|ms_or_token|None
+    platform_options: single|dict|None
```
jabdoa2: This is probably still a leftover, right?

jabdoa2 (approved):
LGTM (with the two cleanups)

mpf/platforms/virtual.py
qcapen: Yes. I missed that one.

mpf/config_spec.yaml
qcapen: Correct. Another one I missed.

PR #1564, opened 2021-04-21
Logic block timer

### LINKED ISSUE #608: Add timeouts to logic blocks
Add a timeout to logic blocks such as counters, sequences and sequels. It should reset the block if it expires. This should only cover this very basic usecase and everything advanced can use an external timer.

atummons:
Jan,

Here is the setup and test file for the time on logic blocks.  For some reason my show file is still in here, and I can't figure out how to get rid of it.  You already merged that into dev, so not sure why its a change, or why its showing in this branch.  So feel free to delete that, or leave it, as I can't seem to pull it out.

Fixes #608

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -835,6 +835,7 @@ logic_blocks_common:
     start_enabled: single|bool|None
     events_when_complete: list|event_posted|None
     events_when_hit: list|event_posted|None
+    logic_block_timeout: single|ms|0
 accruals:
     __valid_in__: machine, mode
     __type__: device
--- a/mpf/devices/logic_blocks.py
+++ b/mpf/devices/logic_blocks.py
@@ -1,4 +1,6 @@
 """Logic Blocks devices."""
+import uuid
+
 from random import shuffle
 
 from typing import Any, List, Optional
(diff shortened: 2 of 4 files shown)


jabdoa2:
Looks good on the first glance. Thanks for also extending the tests. I will rebase it later (seems to be based on an older fork - dont worry I can fix that).

atummons:
Sounds good. Yeah I forked for my other fix, and then thought I got back to this one.  Had to google how to wipe it.  Its saying I am 57 commits behind.  I tried to bring it back to fresh before I started, but clearly didn't do it right.

How do I get my fork to be back to up even with dev?

atummons:
I clicked around and did a PR from MPF/Dev back to my fork Dev, and now it is saying I am one ahead (this PR).  Was that the right way of doing it?  And does that mean I have the correct value online and on my local?  Or are there other steps?

jabdoa2:
I rebased and pushed a small cleanup. Few things I noticed:
- The timeout starts when the logic block is enabled. It always fires. Even when the block completed or gets disabled. I guess we should call `self.delay.remove("timeout")` during disable to prevent unexpected behavior here.
- When the timeout fires the block is reset. That will not disable it and restart the timeout. Is that expected behavior? Should there be an option to choose what it should do? `disable` or `restart`? What do you think?

Jan

atummons:
Thanks!  Yes, I would say a remove of the timeout would make sense during the completed or disabled functions to ensure that it only fires if the logic block is still going.  Do you want me to look into that one?

As for the reset, I read this as just to drop you back to square one whenever the feature was timed out.  I believe I have it setup to where it posts an event when that happens(?).  If I did that correctly, then you could always use that event to post a disable of that logic block.  I am not sure restart matters, as reset does the same, since it is created upon enable.  If we add the remove you described above, then it would only be active if enabled so reset=restart.

atummons:
I think I broke something.... I was trying to get your web changes back to my local, and now I don't have the changes on my web or local...  Hopefully there is an undo button here?

jabdoa2:
I pushed my branch again. Guess you could delete your local branch and check it out again. Merging with your local branch somehow broke it.

atummons:
I added the two remove statements with the latest commit.  So hopefully I got it all cleaned up and didn't break anything.  The compare looks good at this point.

jabdoa2:
Looks very good. All tests also passed. Thanks you!

PR #1565, opened 2021-04-23
Segment Display Transitions

qcapen:
This PR builds on the recent segment display improvements in MPF and MPF-MC. This PR adds new text transition effects to all segment display devices (similar to slide transitions in the MC). These transitions will work with all platforms. The transitions include full color support as well as integrated dot/period and comma support. Transitions include push, cover, uncover, wipe, and split and most allow additional transition text (with independent color settings) paired with the effect when switching between the current and new text.

The segment display text stack has been extended to optionally store color, flashing, and transition information. This allows the entire segment display state to be stored in the stack and restored when the stack entry becomes active again. Parameters with a None value leave the current state unchanged when becoming active. This allows a great deal of control when creating segment display shows. The add action in the segment_display_player can now configure text, and optionally flashing, colors, and transitions.

In addition, a new mask flashing mode was added that allows flashing control over each individual character in the display.

Configs should be backwardly compatible with existing segment display usage with many additional settings now available.

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -1301,6 +1301,7 @@ segment_displays:
     size: single|int|7
     integrated_dots: single|bool|false
     integrated_commas: single|bool|false
+    default_transition_update_hz: single|float_or_token|30
     platform_settings: single|dict|None
     platform: single|str|None
 light_segment_displays_device:
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -1301,6 +1301,7 @@ segment_displays:
     size: single|int|7
     integrated_dots: single|bool|false
     integrated_commas: single|bool|false
+    initial_color: list|color|white
     default_transition_update_hz: single|float_or_token|30
     platform_settings: single|dict|None
     platform: single|str|None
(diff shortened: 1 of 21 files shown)


mpf/config_spec.yaml
```diff
@@ -1309,9 +1314,40 @@ segment_display_player:
     priority: single|int_or_token|0
     text: single|str|None
     color: list|color|None
-    action: single|enum(add,remove,flash,no_flash,flash_match,set_color)|add
+    action: single|enum(add,remove,flash,no_flash,flash_match,flash_mask,set_color)|add
+    transition: ignore
```
qcapen, line 1318: Is there some way to use a subconfig here instead of doing it in code? The segment_display_transitions section contains the legal transition config values. My implementation was based on the MC slide transition config validation. It would be nice if the config could be fully evaluated on load instead of at runtime when a transition is created.

jabdoa2: Currently we do not have that unfortunately. This would be cool. I think this is called tagged union or something.

jabdoa2 (approved):
LGTM

Issue #1402, opened 2019-08-06
Time of day as placeholder


jabdoa2:
Implement seconds, minutes, hours, day, month and year as placeholder. This can be useful in conditions (i.e. midnight multiball).

jabdoa2:
https://groups.google.com/forum/m/#!topic/mpf-users/opVE1U62ggU

pmansukhani:
I'd like to take this up, but I need some more clarity on what we're trying to achieve. Or another example that achieves something similar. Thanks.

jabdoa2:
Hi @pmansukhani sorry I missed your comment. You sure can. The idea is to add a placeholder into machine like `machine.date.minute` or `machine.date.hour` which return the minutes or hour of the day. Those placeholder should then work like any other placeholders (http://docs.missionpinball.org/en/dev/config/instructions/dynamic_values.html). 

For instance this should work:

```
event_player:
  game_started{machine.date.minute == 0 and machine.date.hour=0}: start_midnight_multiball
```

When a game is started at midnight `start_midnight_multiball` should be posted which then can trigger an easteregg multiball.

filipe-gomes:
I would like to take a crack at this, if it's still up for grabs.

jabdoa2:
Go for it!

atummons:
@filipe-gomes Are you doing anything with this one anymore?  

@jabdoa2 Does this need an interface in settings to allow the user to enter the time, and then the system takes over counting?  Or does this just need to read the system clock, and convert if from date time into machine variables for YYYY-MM-DD HH:MM:SS (each as its own machine variable)?

filipe-gomes:
@atummons you can take it. I'd completely forgotten about this, and really just don't have the time anymore

jabdoa2:
> @jabdoa2 Does this need an interface in settings to allow the user to enter the time, and then the system takes over counting? Or does this just need to read the system clock, and convert if from date time into machine variables for YYYY-MM-DD HH:MM:SS (each as its own machine variable)?

This would use the current time of day of the machine. Guess we need to add an option to change that in service mode later or it will certainly be incorrect in most machines.

atummons:
@jabdoa2 How do you best want this to work?  As of now, I have the machine vars created for Year, Month, Day, Hour, Minute, Second.  And test to verify it matches the machine time.  This is working.

[...]

PR #1566, opened 2021-04-25
Added Machine Vars for Date and Time

atummons:
Added machine vars for year, month, day, hour, minute, and second.
New async function was created to update this time every second, so when
you call it in MPF, it will always know the current time when the action
is taken.  Tests were updated to ensure that the machine var time
matches the actual time of the machine.

In the future, we will need to give a setting to write to the system
clock.  But for now, the user would need to set the system clock though
the OS/System, and MPF will read that value into a machine var.

This is related to issue #1402.

### DIFF
--- a/mpf/config_players/event_player.py
+++ b/mpf/config_players/event_player.py
@@ -38,6 +38,32 @@ class EventPlayer(FlatConfigPlayer):
             else:
                 self._post_event(event_dict.name, s, **kwargs)
 
+    # pylint: disable-msg=too-many-arguments
+    def handle_subscription_change(self, value, settings, priority, context, key):
+        """Handle subscriptions."""
+        instance_dict = self._get_instance_dict(context)
+        previous_value = instance_dict.get(key, None)
+        if previous_value == value:
+            return
+        instance_dict[key] = value
+
+        if not value:
+            return
+
+        for event, s in settings.items():
+            s = deepcopy(s)
+            event_dict = self.machine.placeholder_manager.parse_conditional_template(event)
+
+            if event_dict.condition and not event_dict.condition.evaluate({}):
+                continue
+
+            if event_dict.number:
+                delay = Util.string_to_ms(event_dict.number)
+                self.delay.add(callback=self._post_event, ms=delay,
+                               event=event_dict.name, s=s)
+            else:
+                self._post_event(event_dict.name, s)
+
     def _post_event(self, event, s, **kwargs):
         event_name_placeholder = TextTemplate(self.machine, event.replace("(", "{").replace(")", "}"))
         for key, param in s.items():
--- a/mpf/core/clock.py
+++ b/mpf/core/clock.py
@@ -1,5 +1,6 @@
 """MPF clock and main loop."""
 import asyncio
+import datetime
 from typing import Tuple
 from serial_asyncio import create_serial_connection
 from mpf.core.logging import LogMixin
@@ -87,6 +88,11 @@ class ClockBase(LogMixin):
     def get_time(self):
         """Get the last tick made by the clock."""
         return self.loop.time()
+
+    @staticmethod
+    def get_datetime():
+        """Get current datetime."""
+        return datetime.datetime.now()
 
     def start_server(self, client_connected_cb, host=None, port=None, **kwd):
         """Start a server."""
--- a/mpf/core/machine_vars.py
+++ b/mpf/core/machine_vars.py
@@ -1,5 +1,8 @@
 """Contains the MachineVariables class."""
 
+import asyncio
+from time import time, sleep
+import datetime
 import copy
 from platform import platform, python_version, system, release, version, system_alias, machine as platform_machine
 from typing import Any, Dict, Optional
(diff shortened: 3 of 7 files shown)


atummons:
Not sure if we want to add a datetime machine var as well.  So right now they are 6 unique variables.  We could add three more, date (yyyy-mm-dd) and time (hh:mm:ss) and then datetime (yyyy-mm-dd-hh:mm:ss) or whatever format you want it in.  Then you could do everything you want with one check.

Otherwise if you wanted midnight madness, for example, you would have to do {machine|hours = 0 and machine|minutes = 0 and machine|seconds = 0}.  So doable as is, but more config code to write.

Thoughts?

jabdoa2:
This definitely will work. However, it adds some load for updating the variables all the time. We could also hook it into the placeholders here: https://github.com/missionpinball/mpf/blob/ac8d6393c01fb49389d91d8e50ac063b37589e8d/mpf/core/placeholder_manager.py#L539. That would only evaluate them as needed. It would still be a placeholder machine.time.hour or something. So it would work on conditions. However, it would not currently work on slides (as they only directly support machine/player variables but we will change that soon).

atummons:
Jan,

Thanks for looking this over so quickly and providing feedback!  I will look into that a bit.  So basically what you are saying is that what I have updates every second (even when not needed), which was my fear.  This other way with placeholder would evaluate it only when it was called in logic.  So the time would be off, but if I then call it to say run this is its 11am, it would update those values at that time (and only at that time).  Is that correct?

If so, how would a midnight multiball work?  Would the mode start condition be time = midnight?  And if so, would that force your machine to update and evaluate constantly?  That would prevent everyone else's machine from doing that, but if you use the feature it would do that?  Or is there some other way to be more cleaver around how that works?

jabdoa2:
> Thanks for looking this over so quickly and providing feedback! I will look into that a bit. So basically what you are saying is that what I have updates every second (even when not needed), which was my fear. This other way with placeholder would evaluate it only when it was called in logic. So the time would be off, but if I then call it to say run this is its 11am, it would update those values at that time (and only at that time). Is that correct?

Yeah. Basically it would not be a variable at all. It would only be computed once it is accessed.
 
> If so, how would a midnight multiball work? Would the mode start condition be time = midnight? And if so, would that force your machine to update and evaluate constantly? That would prevent everyone else's machine from doing that, but if you use the feature it would do that? Or is there some other way to be more cleaver around how that works?

Good question. For midnight multiball I would expect that this is triggered by some event (i.e. you hit a certain target). Then you would add:

`start_events: my_target_hit{machine.time.hour == 24 and machine.time.minute == 0}`

That would evaluate the time when this event is posted. Theoretically, we could also use subscription syntax: https://docs.missionpinball.org/en/dev/config_players/index.html#subscription-syntax but that would require us to implement https://github.com/missionpinball/mpf/blob/ac8d6393c01fb49389d91d8e50ac063b37589e8d/mpf/core/placeholder_manager.py#L535 for the variable (which would be possible but not needed otherwise).

atummons:
So with Stern games that have Midnight Madness (like my Game of Thrones), it starts right at midnight.  So it doesn't matter what is happening, when Midnight hits, lights, audio and DMD all just launch into the multiball.  When you drain back to single ball it picks up where you left off.

If that is beyond the scope that we want to accomplish here that is fine, and you can use custom code to determine that, or use scheduler to kick off an event in hours:minutes:seconds = 00:00:00 - current time.

How would you like it implemented for now?  The way in which you described above?  Or like Stern?  Or do you have other thoughts?

jabdoa2:
I guess this is fine for now. It will definitely work. One event per second will spam your log quite a bit though. Performance probably wont matter for real machines. One compromise would be to drop seconds. That way we would only get one event per minute. What do think?

With placeholders and subscriptions this autostart would also be possible but slightly more complicated to implement.

atummons:
If you think you could walk through the process for me on how you would do that with placeholders and subscriptions, then I could write up documentation on how to do it.  Then if people want to complain its hard, they are welcome to build it through efficient python code.

Yeah, I don't want to spam logs!  Is it possible to omit this from logs?  Or is that just bad practice and counter to everything else?  Even once a minute in the logs seems a bit excessive, and I don't want to frustrate people since its a default, not optional machine_var.  If we decide to go this route for now, I think we can certainly drop seconds and then set the sleep to 60 seconds.  If we want to do the other, I can dig in and maybe hit you up for questions if I get confused.

atummons:
@jabdoa2 Checking back in on this one.  No rush.  The other thing, I think would be nice to do with this, is incorporate "High Score Today" like JJP has on their games.  This would need a future extension to the High Score mode (or custom setup) to check the date when the game is completed, and compare to that table of high scores for the day.

I think both of these, as you stated above can be done with subscriptions.  Is there documentation, another bug, that you can point me to that shows how the subscription model actually works.  I get the process from a code standpoint, but how does it then check that value.

Alternatively, there is obviously a system clock for the log.  Is there anyway to tap into that value vs having to do machine_vars?  I wouldn't think so, but wanted to ask.

Thanks!

jabdoa2:
You can trigger players based on subscriptions: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/machine_files/shows/modes/mode1/config/mode1.yaml#L18. That also works for other player such as light_player.

Implementation would be here: https://github.com/missionpinball/mpf/blob/ac8d6393c01fb49389d91d8e50ac063b37589e8d/mpf/core/placeholder_manager.py#L539

Maybe we can cordinate a hacking/coding session together over the weekend. I like daily high score as well. That would be a nice addition to high score mode.


Jan

atummons:
Jan,

That part makes sense to me that you can subscribe, and when the value is set trigger an event, and how you would get it.  But I'm still struggling with how we set that value without having a trigger to update the time periodically.  I am sure I am missing something obvious though.

As for coding together, that sounds good.  I am in US central time, so I believe 7 hours behind you.  I am generally not up and moving until about 9am CST, but let me know if you have some time.  Or we can coordinate over on the Slack channel (User: atum).

PR #1569, opened 2021-05-06
Refactor segment displays and add vpe

jabdoa2:
- Rename `initial_color` to `default_color`
- Ensure segment display site in device (right aligns currently - might become an option later)
- Set colors on fadeout transitions
- Add Visual Pinball Engine support


mpf/devices/segment_display/segment_display.py
```diff
@@ -251,31 +261,93 @@ def _update_stack(self) -> None:
...
+            return
+
+        self._current_state = new_state
+        text = new_state.text
+        colors = new_state.colors
+        flashing = new_state.flashing
+        flash_mask = new_state.flash_mask
+
+        # make sure text is the same length as the display
+        if len(text) > self.size:
```
qcapen: I don't this this will work when using integrated dots/periods or commas as they are collapsed when displaying them shortening the actual length of the text. For example, with displays that support integrated commas, the following text is legal to be sent to a display of size 7: "1,000,000" even though len(text) is 9 in this case.

qcapen: In the text transition calculations, I created lists of characters that each had additional attributes (color, integrated dot, integrated comma). It was the easiest way to perform operations on the text while preserving colors and integrated dots/commas. Perhaps we might want to do something similar here or even create a SegmentDisplayCharacter class to use throughout all segment display-related classes (including the platforms).

mpf/devices/segment_display/segment_display.py
```diff
@@ -20,6 +20,29 @@
     from mpf.core.platform import SegmentDisplayPlatform    # pylint: disable-msg=cyclic-import,unused-import; # noqa
 
 
+class SegmentDisplayState:
```
qcapen, line 23: I like the addition of a state class

mpf/devices/segment_display/segment_display.py
jabdoa2: Makes sense. Thanks! We definitely need a test for that.

mpf/devices/segment_display/segment_display.py
```diff
@@ -20,6 +20,29 @@
...
+                 flashing: FlashingType, flash_mask: Optional[str] = None):
+        """Class initializer."""
+        self.text = text
+        self.colors = colors
+        self.flashing = flashing
+        self.flash_mask = flash_mask
+
+    def __repr__(self):
+        """Return str representation."""
+        return '<TextStackEntry: {} (colors: {}, flashing: {} flashing_mask: {}) >'.format(
```
qcapen: Should we change 'TextStackEntry' to 'SegmentDisplayState'?

mpf/devices/segment_display/segment_display.py
```diff
@@ -124,51 +142,38 @@ def add_text(self, text: str, priority: int = 0, key: str = None) -> None:
...
+            self._update_stack()
 
     # pylint: disable=too-many-arguments
     def _start_transition(self, transition: TransitionBase, current_text: str, new_text: str,
-                          current_colors: Optional[List[RGBColor]] = None, new_colors: Optional[List[RGBColor]] = None,
-                          update_hz: float = 30.0):
+                          current_colors: List[RGBColor], new_colors: List[RGBColor],
+                          update_hz: float):
         """Start the specified transition."""
+        current_colors = self._expand_colors(current_colors, len(current_text))
```
qcapen, line 159: Same potential problem with the text length here (integrated commas/dots)

mpf/devices/segment_display/segment_display.py
jabdoa2: Guess the stack contains slightly more. But the idea sounds good

mpf/devices/segment_display/segment_display.py
jabdoa2: Yeah we probably have to interleave color + text in the beginning. Or do we? I guess colors are just for the character positions and interleaved dots just get the same color. I will have a look tomorrow

PR #1570, opened 2021-05-06
Reset timer on show advance/step_back

mike-mcgann:
Resets the show timer when advancing or stepping back.

I am setting up an attract mode that advances or steps back when the flipper buttons are pressed. Each slide changes after 4 seconds. When rapidly pressing the flipper button, say 5 times, the slide that it stops on doesn't change for 20 seconds or so as if the times are being accumulated. The expected behavior is to change after 4 seconds. 

This change resets the next step time to the current time when advancing or stepping back.

Let me know if you have any questions.

### DIFF
--- a/mpf/assets/show.py
+++ b/mpf/assets/show.py
@@ -614,6 +614,7 @@ class RunningShow:
     def advance(self, steps=1, show_step=None):
         """Manually advance this show to the next step."""
         self._remove_delay_handler()
+        self.next_step_time = self.machine.clock.get_time()
 
         if steps != 1:
             self.next_step_index += steps - 1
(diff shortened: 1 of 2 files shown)


mike-mcgann:
I've added the following machine configuration for testing:

https://github.com/drop-target-pinball/mpf-sandbox/tree/main/pr-1570

Start MPF for that machine configuration and use the left and right arrow keys to move through the "attract" mode. Then hit the right arrow key quickly five times and notice how it takes about 20 seconds to advance the slide once done. The change in this branch corrects this problem.

jabdoa2:
I did not try it yet but it looks very good. Thanks for adding a test.

jabdoa2:
Thanks!

PR #1571, opened 2021-06-04
Penny K Pinball PKONE Hardware Platform

qcapen:
Adding hardware platform for the new Penny K Pinball PKONE hardware platform (hardware will be released any day now).


mpf/core/platform_batch_light_system.py
```diff
@@ -38,7 +38,7 @@ def get_fade_and_brightness(self, current_time):
         max_fade_ms = self.get_max_fade_ms()
         start_brightness, start_time, target_brightness, target_time = self._current_fade
         fade_ms = int(round((target_time - current_time) * 1000.0))
-        if fade_ms > max_fade_ms > 0:
+        if fade_ms > max_fade_ms >= 0:
```
jabdoa2, line 41: Sure about this one? If max_fade_ms is 0 there should not be any fade. Right?

mpf/mpfconfig.yaml
```diff
@@ -245,6 +246,10 @@ fadecandy:
     keyframe_interpolation: True
     dithering: True
 
+pkone:
+    port: com3
```
jabdoa2: I guess we should remove this block. Config validation already contains the baud rate and the port will be different for most people.

jabdoa2:
Two small nits above. Other than that LGTM! Very well done. I like the way you implemented it.

jabdoa2 (approved):
LGTM!

PR #1576, opened 2021-06-11
fix embedded dots in light_segment_displays

jabdoa2:


qcapen:
Looks good. Do you also want to add the optional embedded commas as well while you're at it? I know is is not as common, but most of the System 11 displays had both embedded commas and dots (independent from each other). Very nice for scores (although dots work for that in some locales).

jabdoa2:
> Looks good. Do you also want to add the optional embedded commas as well while you're at it? I know is is not as common, but most of the System 11 displays had both embedded commas and dots (independent from each other). Very nice for scores (although dots work for that in some locales).

Good idea. Will do that as well.

PR #1578, opened 2021-06-19
Crash reporter

jabdoa2:
Add MPF Crash Reporter

Corresponding backend is here: https://github.com/missionpinball/mpf-crash-reporter. Already running as well. There is no report rendering or processing yet. We may want to add that later. Users will get back a report url which we can use in the support forum (but we need to get the report manually).

### DIFF
--- a/mpf/commands/game.py
+++ b/mpf/commands/game.py
@@ -12,10 +12,12 @@ from logging.handlers import QueueHandler, SysLogHandler
 
 from queue import Queue
 
+from mpf.core.crash_reporter import report_crash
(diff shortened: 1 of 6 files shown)


avanwinkle (commented):
Very cool! I'll have to make some crashes to check it out! :D

jabdoa2:
> Very cool! I'll have to make some crashes to check it out! :D

Go for it. Backend is already up. It currently does not let you download the report. Not sure if we want some auth there or not. So far it dumps the report on disk on my kubernetes cluster. We either need some batch processing or some database later on to index it.

PR #1579, opened 2021-06-27
Option to override switch in starting game from s_start

cfbenn:
Added ability to pass in the start switch name if you want to start a test game that does not have a switch s_start.  If it is not specified, it will default to s_start.


jabdoa2:
Awesome. Thanks!

PR #1584, opened 2021-07-12
Detect cyclical light chains

avanwinkle:
I don't want to admit how much time I spent pounding my head against the keyboard trying to figure out why my game kept hanging on bootup without any logs or errors.

Suffice it to say, here's a PR that detects when a light chain of `previous:` declarations creates an infinite loop and lets the user know where the infinite loop is :)


jabdoa2:
LGTM!

PR #1585, opened 2021-07-12
Modified and extended 16segment "font"

cobra18t:


jabdoa2:
I would prefer to replace the existing 16segment mappings. Yours seem to be strickl better to me. Do you see any reason to keep the old ones? I think the lower case letters were extrapolated from other mappings (with less segments).

cobra18t:
I am fine with replacing it. I was not sure if people were already using the existing one and preferred it. I you do not think so, I can just replace the existing.

jabdoa2:
> I am fine with replacing it. I was not sure if people were already using the existing one and preferred it. I you do not think so, I can just replace the existing.

Those are relatively new and yours are better. There isnt much hardware for it out yet with MPF support.

cobra18t:
I updated everything to overwrite the existing font and eventually got it to pass the automated tests.

jabdoa2:
Awesome. Thanks!

Issue #766, opened 2017-02-14
Create a blinkenlights player


toomanybrians:
For scenarios where you want lights to alternate between one (or more) colors, we should create a blinkenlights config player which can have all the specific properties, timing, and sync we'd want for that.

Details will need to be worked out. This is more of a placeholder for now.

RoseAndres:
Hi there. As far as working out the details for this, who/what needs to be involved in working them out?

jabdoa2:
Brian would be the right person but he is currently very busy in his private life (will probably be back in a few month). The idea here is to develop a feature to leverage RGB inserts on lane arrows inserts on modern machines. Let's say you got a blue, a red and green achievement in your game. Player completed blue and red. Now he should shoot for the scoop (or any shot) to collect the reward. Should the arrow be blue or red?

The idea here is to stack awards on the scoop by making it flash for 100ms in red and then 100ms in blue. Then 300ms off. Of couse this must stay in sync with all other inserts (even if those flash in only one or three colors). Sort order should be always the same. I'm not 100% sure if any machines actually do this currently.

Brian wanted to have it in MPF but we decided not to include this inside the light stack which only displays the show with the highest priority. Therefore this is a new feature which gathers all the colors for one light, creates (or parameters) a show and puts the show on the light (with sync_ms to keep it in sync with other shows).

Since it is nothing which is requests by any user you are free to make modifications/omit parts for now or create a minimal example. We always can extend it when people start using it.

densminger:
I thought about some config options for how this might work a while back:

https://groups.google.com/g/mpf-users/c/tcU-dRa1zdQ/m/2dAS0_mABgAJ

jabdoa2:
Let me just copy your suggestion over here:

> I considered something like what you suggested, but the thing with Game of Thrones is, there are 7 modes, any 2 of which can run concurrently.  So there would be 7 "slots" for the lights and that would mean the lights were in the "off" state most of the time.  I saw your link for the blinkenlights discussion and I think this would be a solution to this problem.  I started thinking about this some.  I was thinking to have a "blinkenlights" section of the config, along with a "blinkenlights_player".
> 
> Example:

```
blinkenlights:
  stark:
    each_duration: 1 # How long each color will be on per cycle
    off_duration: 1 # How long the light is off per cycle
[...]

PR #1591, opened 2021-08-08
Blinkenlights

densminger:
Blinkenlights feature, as discussed in issue #766


mpf/config_players/blinkenlight_player.py
```diff
@@ -0,0 +1,61 @@
...
+class BlinkenlightPlayer(DeviceConfigPlayer):
+
+    """Blinks lights different colors based on config."""
+
+    config_file_section = 'blinkenlight_player'
+    show_section = 'blinkenlights'
+    machine_collection_name = 'blinkenlights'
+    allow_placeholders_in_keys = True
+
+    def __init__(self, machine):
```
jabdoa2: This method is the same as the default. Could be removed.

mpf/config_players/blinkenlight_player.py
```diff
@@ -0,0 +1,61 @@
...
+                self._remove_all_colors(blinkenlight)
+            blinkenlight.num_colors = len(blinkenlight.colors)
+            blinkenlight._restart()
+
+    def _add_color(self, blinkenlight, color, key):
+        if blinkenlight is None:
+            return
+        # only add this color if the key does not already exist
+        if len([x for x in blinkenlight.colors if x[1] == key]) < 1:
+            blinkenlight.colors.append((color, key))
```
jabdoa2: I would turn this into a method in the blinkenlight device.

mpf/devices/blinkenlight.py
```diff
@@ -0,0 +1,80 @@
...
+        # self.each_duration = 0
+        # self.off_duration = 0
+        # self.off_when_multiple = False
+        # self.light = None   # type: Light
+        self._colors = []
+        self.delay = DelayManager(machine)
+        self._color_i = 0
+        self._num_colors = 0
+        # Add this device to the blinkenlight section
+        self.machine.blinkenlights[name] = self
```
jabdoa2: This should not be needed here.

mpf/devices/blinkenlight.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @property
+    def colors(self):
+        return self._colors
+
+    @property
+    def num_colors(self):
+        # self.info_log('num_colors queried - returning {}'.format(len(self._colors)))
```
jabdoa2: Please remove commented out code.

mpf/config_players/blinkenlight_player.py
```diff
@@ -0,0 +1,61 @@
...
+        for blinkenlight, s in settings.items():
+            action = s['action']
+            if action == 'add':
+                self._add_color(blinkenlight, s['color'], s['key'])
+            elif action == 'remove':
+                self._remove_color(blinkenlight, s['key'])
+            elif action == 'removeall':
+                self._remove_all_colors(blinkenlight)
+            blinkenlight.num_colors = len(blinkenlight.colors)
+            blinkenlight._restart()
```
jabdoa2: Not sure if restarting this here is wise. It would destroy sync and might cause flickering on mode start/stop.

mpf/devices/blinkenlight.py
```diff
@@ -0,0 +1,80 @@
...
+        return self.config['light']
+
+    def _restart(self):
+        self._color_i = 0
+        self.delay.clear()
+        self.config['light'].color(RGBColor("off"))
+        self._perform_step()
+
+    def _perform_step(self):
+        if len(self._colors) == 0:
```
jabdoa2: I would turn this into a coroutine. We just need to figure out how we synchonize all blinkenlights. Maybe always start at sync_ms xx. Would that work? That would slightly change the logic. I looked at some sterns and it seems that they got up to 2 slots (i.e. on star trek). If both are filled the light will toggle between them. If one is filled it will turn off. Lights seem to be mostly in sync.

jabdoa2 (commented):
I like it. Some small nits and some thoughts about sync.

mpf/devices/blinkenlight.py
densminger: I'll have to think about this.  Are you suggesting that each color change is in sync (so that each blinkenlight will change from one color to the next in sync with the others), or that each blinkenlight starts/finishes its entire cycle in sync will all other blinkenlights?  In the second case, I don't know how it would work if one blinkenlight had 2 colors and another has 6, since it would take the 6 color blinkenlight three times as long to cycle through its colors as the 2 color blinkenlight, so they couldn't stay in sync.

jabdoa2: I would imagine to sync the cycles. One idea would be to run them at a fixed duration and fill up the remaining slots with black (or transparency). This seems to be the case for sterns.

I see that this would be an issue with variable length cycles. Alternatively, we could only sync same length cycles (by using sync_ms = cycle length). That might still work well for the common case: green + off and green + red would sync but green + red + yellow would not sync. Guess that would be ok.

What do you think?

densminger: I’m thinking I could make both options happen. Either you could specify a cycle_duration=1 for example, so that each cycle is 1 second regardless of how many colors there are, or color_duration=1 would specify that each color lasts 1 second. And then the sync_ms would sync accordingly (either on a cycle or color basis). Would that work?

atummons: So with Stern, Game of Thrones has 3 colors that can be on a single light at a time.  I have a slow mo video, if you want.  I could upload and share the link.

The way it works is all are the same duration.  Lets say 1 for example.  Then each light has its cycle of what it can do.  LED1 has 1 color (R), LED2 has 2 colors (R,B), and LED3 has 3 colors (R,B,G).  They all loop through their colors and go to off at the end before restarting.  So here is a time breakdown.

Time 0: LED1 = R, LED2 = R, LED3 = R
Time 1: LED1 = Off, LED2 = B, LED3 = B
Time 2: LED1 = R, LED2 = Off, LED3 = G
Time 3: LED1= Off, LED2 = R, LED3 = Off

This would repeat.  So they aren't all showing red at once, or anything like that.  But that all shift at the exact same time.  So we would need a sync_ms for all blinkenlights (not each individual light), and a fixed duration for all lights (not each individual light).  This would mimic Stern to make it seamless to the user, as you have things all shifting simultaneously, instead of randomly.

I don't believe we need to enforce all of those rules, but rather expose the variables.  So I can set all of mine to duration 1 and fade_ms of 100.  Then mine will sync up.  But I would not set each to have 4 slots always, that are either filled or blank.  I would make them dynamic, so that they fill the slots that they have.  I had started code for this as well.  So I might play a bit more to see if it even works.  Then maybe we can compare and cherry pick best of class for inclusion in MPF.

jabdoa2: Thanks for checking that. We should support it this way.

Can you check one more thing for me: If you got two lights with R+B+off are those always in sync? I mean does both show red at the same time? I would guess so.

Basically, the above is what we should get with sync ms based on cycle length. Maybe, we can hack this together in the next few days.

atummons: Yes, all binkenlights are exactly on the same sync.  So they all hit color 1 at the exact same time, color 2, etc.  So if two lights have the same values (R, B, Off), they would be perfectly in sync.  Then when you hit the R on shot 1, that gets removed from the loop for that light.  So then they aren't showing the same colors at the same time (1 showing B, Off and the other showing R, B, Off), but the sync_ms for transition is identical.  Here is a link to a video showing this on Game of Thrones: https://pindev.slack.com/files/U01GQ857T4L/F02A78ME7KR/img_8126.mov

densminger: Thanks. I believe the way I've outlined would work then. What you're describing for how GoT works would be just like setting color_duration, so that each color on a particular light lasts for a given amount of time. The sync issue still might need some thought, since a sync_ms on each blinkenlight wouldn't ensure that they all started their cycle at the same time. For example, if I have 2 blinkenlights and they're each set with color_duration=500ms, and sync_ms=1s, this would ensure they started at the same time if they each had 2 colors. But if they each had 4 colors, they could be a half-cycle off from each other.

atummons: That's a good point.  So might need to be at a higher level.  let me do some "research" on Game of Thrones.  I can get a mode started, and then add shots by progressing the modes.  I will see if it then syncs up with all shots or not.  I have a suspicious the answer will be yes.

But regardless of how Stern implemented, it would be nice to have them all start from the same place.  But with long duration, that becomes a bit of lag between when the event occurs and the lights add that color.

densminger: How about this:

Blibkenlight_player figures out the sync_ms automatically, based on the current cycle length of the blinkenlight (I.e., the number of colors the light currently has if color_duration is set, or else the cycle_duration value if it is set). This is probably what Jan was suggesting earlier. Based on this, the blinkenlight_player could figure out which color should be on at any given time so that it wouldn't have to wait till the beginning of its cycle if the sync_ms happens to be very large when it starts.

In this case, there would be no user-configurable sync_ms option for blinkenlights, since it would be calculated and synced automatically.

However this does bring up another issue. This would require that all colors are added to the blinkenlight in the same order, or else some other ordering would be required. If one blinkenlight had its colors added R then G but reversed for the other blinkenlight, then these would flash alternating colors. 

(Also, this discussion should probably be moved to the issue #766 tracker instead of here on the pull request)

densminger: In the latest commit, I addressed all these.  I added different parameters:

```
blinkenlights:
  my_blinkenlight:
    color_duration: 1s
    off_when_multiple: false
    light: l_stark_sigil
    priority: 1000
  my_blinkenlight2:
    cycle_duration: 1s
    off_when_multiple: false
    light: l_martell_sigil
    priority: 1000

```
Now you can either specify color_duration or cycle_duration (you have to specify one or the other, but not both - this is validated in code).  Sync across multiple blinkenlights is achieved based on the logic I outlined above (i.e., the cycle duration is automatically calculated and a "virtual" sync_ms is calculated based on the length of the cycle).  When the blinkenlight starts, it doesn't have to wait till the beginning of the next cycle (as it would with a typical sync_ms), but will instead figure out which color should be on at that point and go from there.

Also I added a priority.  This defaults to 0 if not specified.

mpf/config_players/blinkenlight_player.py
densminger: I think this is ok, since now the blinkenlight figures out where it should be in the cycle, so it shouldn't interfere with the sync or cause flickering.

mpf/devices/blinkenlight.py
jabdoa2: That sounds awesome!

> Sync across multiple blinkenlights is achieved based on the logic I outlined above (i.e., the cycle duration is automatically calculated and a "virtual" sync_ms is calculated based on the length of the cycle). When the blinkenlight starts, it doesn't have to wait till the beginning of the

We have to check how that looks. Might give us weird effects im some cases. One slightly more complex alternative would be to always finish the current cycle and only considering changes when the loop starts. Guess that would be something we might add later.

mpf/devices/blinkenlight.py
```diff
@@ -0,0 +1,131 @@
...
+        self._cycle_duration = self.config['cycle_duration']
+        if (self._color_duration is None and self._cycle_duration is None) or (self._color_duration is not None and self._cycle_duration is not None):
+            self._blinkenlight_validation_error("Either color_duration or cycle_duration must be specified, but not both.", 1)
+
+    def _blinkenlight_validation_error(self, msg, error_code) -> "NoReturn":  # pragma: no cover
+        raise ConfigFileError('"{}" >> {}'.format(self.name, msg), error_code, "blinkenlight", self.name)
+
+    @property
+    def num_colors(self):
+        # I would prefer this to be a calculated value (i.e., just return len(self._colors)), but then there's no setter involved so things like mpf monitor won't know when this value gets updated.
```
jabdoa2: There are some magic methods on DeviceMonitor to notify it about changes.

mpf/devices/blinkenlight.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @property
+    def light(self):
+        return self.config['light']
+
+    @property
+    def num_colors_in_cycle(self):
+        # similar to num_colors, but adds 1 for the "off" color if there is one between cycles
+        return self.num_colors + (1 if self._off_between_cycles() else 0)
+
```
jabdoa2, line 53: This additional newline probably makes the linter unhappy.

jabdoa2:
Guess we are good to land this. Some small coding style changes to fix and we can merge this:

```
Messages

========

mpf/config_players/blinkenlight_player.py

  Line: 2

    pylint: unused-import / Unused deepcopy imported from copy

  Line: 4

    pylint: unused-import / Unused RGBColor imported from mpf.core.rgb_color

  Line: 5

    pylint: unused-import / Unused Util imported from mpf.core.utility_functions

  Line: 29

    pylint: protected-access / Access to a protected member _restart of a client class (col 12)

  Line: 31

    pylint: no-self-use / Method could be a function (col 4)

  Line: 36

    pylint: no-self-use / Method could be a function (col 4)

  Line: 41

    pylint: no-self-use / Method could be a function (col 4)

mpf/devices/blinkenlight.py

  Line: 3

    pylint: unused-import / Unused partial imported from functools

  Line: 7

    pylint: unused-import / Unused ColorException imported from mpf.core.rgb_color

  Line: 13

    pep8: E302 / expected 2 blank lines, found 1 (col 1)

    pep8: E302 / expected 2 blank lines, found 1 (col 1)

  Line: 30

    pep257: D102 / Missing docstring in public method

  Line: 33

    pylint: attribute-defined-outside-init / Attribute '_color_duration' defined outside __init__ (col 8)

  Line: 34

    pylint: attribute-defined-outside-init / Attribute '_cycle_duration' defined outside __init__ (col 8)

  Line: 35

    pep8: E501 / line too long (150 > 120 characters) (col 121)

  Line: 36

    pep8: E501 / line too long (126 > 120 characters) (col 121)

  Line: 38

    pyflakes: F821 / undefined name 'NoReturn' (col 66)

  Line: 42

    pylint: missing-function-docstring / Missing function or method docstring (col 4)

    pep257: D102 / Missing docstring in public method

  Line: 43

    pep8: E501 / line too long (197 > 120 characters) (col 121)

  Line: 47

    pep8: N803 / argument name 'newValue' should be lowercase (col 27)

    pep8: N803 / argument name 'newValue' should be lowercase (col 27)

    pylint: invalid-name / Argument name "newValue" doesn't conform to '^[a-z_][a-z0-9]*((_[a-z0-9]+)*)?$' pattern (col 4)

  Line: 54

    pylint: missing-function-docstring / Missing function or method docstring (col 4)

    pep257: D102 / Missing docstring in public method

  Line: 58

    pylint: missing-function-docstring / Missing function or method docstring (col 4)

    pep257: D102 / Missing docstring in public method

  Line: 63

    pep8: E303 / too many blank lines (2) (col 5)

    pep8: E303 / too many blank lines (2) (col 5)

  Line: 72

    pylint: missing-function-docstring / Missing function or method docstring (col 4)

    pep257: D102 / Missing docstring in public method

  Line: 79

    pylint: missing-function-docstring / Missing function or method docstring (col 4)

    pep257: D102 / Missing docstring in public method

  Line: 84

    pylint: missing-function-docstring / Missing function or method docstring (col 4)

    pep257: D102 / Missing docstring in public method
```

mpf/config_players/blinkenlight_player.py
```diff
@@ -0,0 +1,48 @@
...
+
+        for blinkenlight, s in settings.items():
+            action = s['action']
+            if action == 'add':
+                self._add_color(blinkenlight, s['color'], s['key'])
+            elif action == 'remove':
+                self._remove_color(blinkenlight, s['key'])
+            elif action == 'removeall':
+                self._remove_all_colors(blinkenlight)
+            blinkenlight._restart()
```
jabdoa2: Guess this could be included in add/remove of the blinkenlight device.

mpf/config_players/blinkenlight_player.py
```diff
@@ -0,0 +1,48 @@
...
+    allow_placeholders_in_keys = True
+
+    def play(self, settings, context, calling_context, priority=0, **kwargs):
+        """Adds/removes colors to blink."""
+        del kwargs
+
+        for blinkenlight, s in settings.items():
+            action = s['action']
+            if action == 'add':
+                self._add_color(blinkenlight, s['color'], s['key'])
```
jabdoa2: We should pass priority here (`priority + s['priority']`).

mpf/devices/blinkenlight.py
```diff
@@ -0,0 +1,131 @@
...
+        # this is the key that's passed to the light as the color's key
+        # so that we can remove this blinkenlight's color from the light stack
+        # when all the blinkenlight's colors are removed.
+        # This prevents the light from staying in the off state if the blinkenlight
+        # was on top of the priority list.
+        return 'blinkenlight_{}'.format(self.name)
+
+    def add_color(self, color, key):
+        # only add this color if the key does not already exist
+        if len([x for x in self._colors if x[1] == key]) < 1:
```
jabdoa2: Guess we should just remove the entry if it is already in. This allows updates.

mpf/devices/blinkenlight.py
```diff
@@ -0,0 +1,131 @@
...
+            self._colors.append((color, key))
+            self.num_colors += 1
+            self.info_log('Color {} with key {} added'.format(color, key))
+
```
jabdoa2, line 61: I would order the list by priority here

atummons:
Looking good.  Well I haven't had time to look at mine any further, and you have already solved the problems, so I'll bow out here on fixes.  I think it is looking great with Jan's comments there.

Jan, how do we pull this down to run against our local machine?  I'd love to pull it down and test it out with Monitor to see how its works.  Or do we have to have it merged before that can be done?

jabdoa2:
> Jan, how do we pull this down to run against our local machine? I'd love to pull it down and test it out with Monitor to see how its works. Or do we have to have it merged before that can be done?

I roughly explained this in my video on the IDE setup. Short:

- git clone https://github.com/densminger/mpf.git (clone Daves repo)
- cd mpf (change into new dir)
- git checkout blinkenlights (change to daves branch)
- pip3 install -e . (install in dev mode)

There are other ways with multiple remotes in one checkout but those are a bit tricky to use unless you are very familiar with the inner workings of git.

atummons:
> 
> 
> 
> I roughly explained this in my video on the IDE setup. Short:
> 
>     * git clone https://github.com/densminger/mpf.git (clone Daves repo)
> 
>     * cd mpf (change into new dir)
> 
>     * git checkout blinkenlights (change to daves branch)
> 
>     * pip3 install -e . (install in dev mode)
> 
> 
> There are other ways with multiple remotes in one checkout but those are a bit tricky to use unless you are very familiar with the inner workings of git.

Haven't watched that one yet (in my saved for later on YouTube), and I love the content!!!  I will check it out to make sure I understand the ins and outs.  But basically same as getting my local, just swapping the steps for Dave, and then do mine to switch back.  Makes perfect sense!

mpf/devices/blinkenlight.py
densminger: Yes this could be weird in some cases but it works in a predictable manner. For example, if a blibkenlight has 2 colors (let's say red and green), so it blinks red, green, red, green. If green is removed by a mode while the green light is lit, then for the rest of green's "slot" time, it would the light would be off since now the blinkenlight should flash red, off, red, off. This could cause it to look like it flickers but it's intended behavior in this case.

densminger:
So here's an example of some of the "weirdness" that might happen:

Let's say the blinkenlight has `cycle_duration=1s` (that is, each cycle lasts 1 second, regardless of how many colors the blinkenlight has).  The blinkenlight only has one color (red). In this case, the blinkenlight's cycle would look like this:

```
|       1s      |       1s      |       1s      |       1s      |
|-------|-------|-------|-------|-------|-------|-------|-------|
|  red  |  off  |  red  |  off  |  red  |  off  |  red  |  off  |
```

If green color is added to the blinkenlight (this assumes `off_when_multiple=false`), the cycle would look like this:

```
|       1s      |       1s      |       1s      |       1s      |
|-------|-------|-------|-------|-------|-------|-------|-------|
|  red  | green |  red  | green |  red  | green |  red  | green |
```

Then a third color (blue) is added:

```
|       1s      |       1s      |       1s      |       1s      |
|  r |  g  |  b |  r |  g  |  b |  r |  g  |  b |  r |  g  |  b |
```

Note that each color now is only 1/3 of a second long, since there are three of them per cycle now.

Now, let's say blue is removed from the blinkenlight, while the blinkenlight is currently showing a blue color:

```
|       1s      |       1s      |       1s      |       1s      |
|  r |  g  |  b |  r |  g  | b|g|  red  | green |  red  | green |
        blue is removed here ^
```

Notice how blue is displayed when the color is removed, and the light immediately switches to green, since green should be displayed at that point in time now that the blinkenlight only has 2 colors.  So the end result is green "flashes" very briefly before red is displayed again and the red/green cycle starts.

I hope this makes sense.  It's predictable, and I think if I was playing the game and saw this happen I would understand what happened and why the color change occurred like that.

jabdoa2:
> So here's an example of some of the "weirdness" that might happen:
> 
> Let's say the blinkenlight has `cycle_duration=1s` (that is, each cycle lasts 1 second, regardless of how many colors the blinkenlight has). The blinkenlight only has one color (red). In this case, the blinkenlight's cycle would look like this:
> 
> ```
> |       1s      |       1s      |       1s      |       1s      |
> |-------|-------|-------|-------|-------|-------|-------|-------|
> |  red  |  off  |  red  |  off  |  red  |  off  |  red  |  off  |
> ```
> 
> If green color is added to the blinkenlight (this assumes `off_when_multiple=false`), the cycle would look like this:
> 
> ```
> |       1s      |       1s      |       1s      |       1s      |
> |-------|-------|-------|-------|-------|-------|-------|-------|
> |  red  | green |  red  | green |  red  | green |  red  | green |
> ```
> 
> Then a third color (blue) is added:
> 
> ```
> |       1s      |       1s      |       1s      |       1s      |
> |  r |  g  |  b |  r |  g  |  b |  r |  g  |  b |  r |  g  |  b |
> ```
> 
> Note that each color now is only 1/3 of a second long, since there are three of them per cycle now.
> 
> Now, let's say blue is removed from the blinkenlight, while the blinkenlight is currently showing a blue color:
> 
> ```
> |       1s      |       1s      |       1s      |       1s      |
> |  r |  g  |  b |  r |  g  | b|g|  red  | green |  red  | green |
>         blue is removed here ^
> ```
> 
> Notice how blue is displayed when the color is removed, and the light immediately switches to green, since green should be displayed at that point in time now that the blinkenlight only has 2 colors. So the end result is green "flashes" very briefly before red is displayed again and the red/green cycle starts.
> 
> I hope this makes sense. It's predictable, and I think if I was playing the game and saw this happen I would understand what happened and why the color change occurred like that.

Agreed. It definitely works this way :-). I guess we should add the ordering by priority in `blinkenlights_player` and with that we should be good to go.

mpf/devices/blinkenlight.py
densminger: I tried to find the magic methods you mentioned but I don't see any on DeviceMonitor aside from `__init__` and `__call__`.  I know `__setattr__` is used to notify subscribers when a variable is changed, but how can I notifiy subscribers without creating a fake variable like this?

mpf/config_players/blinkenlight_player.py
densminger: This makes me think. I believe that including priority here would include the mode's priority with the color.  This would mean that each color of a blinkenlight could have a different priority.  This seems strange to me.  The way I have it now is, the blinkenlight itself has a priority, and colors added to it all get the same priority regardless of the mode that added it.  That way, all of the blinkenlight's colors will either get shown or not shown as a group, instead of some colors being shown and others not shown due to the priority of the mode that added the color.

Another thing to think about is, what about the "off" color between cycles?  This isn't explicitly added by any mode, so what should its priority be?  If it's 0, then the "off" color might not get displayed at all which might look weird.

I'm not sure how this should be handled truthfully.

jabdoa2: Naming is a bit confusing here. `priority` on the light stack is one thing (which you already implemented). Another thing would be mode or stack priority or order. That would determine if red or blue comes first in your blinkenlight. This way the order should become deterministic (and based on mode priority). Makes sense?

densminger: Got it. Thanks.

mpf/devices/blinkenlight.py
densminger: I order the list in _restart, since add_color, remove_color, and remove_all_colors calls _restart.

densminger:
I think this is ready to go.

atummons:
@densminger Looks like it needs a couple of cleanup items to pass the Travis CI build test.  All looks like general formatting and consistency stuff.  Once that's cleaned up, it should pass all three tests and get the green check mark.  Great work!
```
mpf/config_players/blinkenlight_player.py
  Line: 28
    pylint: missing-function-docstring / Missing function or method docstring (col 4)
    pep257: D102 / Missing docstring in public method
  Line: 34
    pylint: missing-function-docstring / Missing function or method docstring (col 4)
    pep257: D102 / Missing docstring in public method
  Line: 40
    pylint: missing-function-docstring / Missing function or method docstring (col 4)
    pep257: D102 / Missing docstring in public method

mpf/devices/blinkenlight.py
  Line: 15
    pylint: ungrouped-imports / Imports from package typing are not grouped (col 4)
  Line: 45
    pep8: E502 / the backslash is redundant between brackets (col 50)
    pep8: E502 / the backslash is redundant between brackets (col 50)
  Line: 53
    pep257: D401 / First line should be in imperative mood; try rephrasing (found 'Number')
  Line: 65
    pep257: D401 / First line should be in imperative mood; try rephrasing (found 'The')
  Line: 97
    pep257: D300 / Use """triple double quotes""" (found "-quotes)
  Line: 104
    pep257: D401 / First line should be in imperative mood (perhaps 'Remove', not 'Removes')
```

densminger:
Green check!

atummons:
Sweet!!!

densminger:
I've been playing with this, and I have a question.  Right now, a `blinkenlight` is a `SystemWideDevice` object.  As such, it exists at a machine level, and the `blinkenlight_player` has no concept of a game or player state.  So, when a mode adds a color to a blinkenlight, and the player's ball drains, that blinkenlight will continue to flash unless the mode specifically tells it to remove its colors when the mode ends (and it would have to "remember" which colors were active when it restarts).  Same thing if the game ends. My question is, is this ok, or should I change the `blinkenlight` to be a `LogicBlock` instead so it can use the `persist_state` functionality of logic blocks?  It's not exactly a logic block since it interfaces with hardware (lights) so I don't know if this is a good fit or not.

Edit: I just realized that the `ConfigPlayer` class has a `mode_stop` method.  I'm guessing I could override that method and unload the blinkenlight when modes end.  I would just need to keep track of which mode added which colors. So there's likely enough functionality already in `ConfigPlayer` to handle this correctly.

Edit2: I've marked this PR as draft.  I want to get this working before it gets merged.

atummons:
I was looking at the same thing when I saw your edit. Agreed, just need to use that to cancel when mode ends there and it clears the stack for it's called keys.

As for starting back up, is there a scenario where it would need to do that? I would assume my shot would be in a state that it's profile calls a specific show. So if the shot is in that state when the mode starts again, it should automatically be calling the show, and adding it back to the stack.

densminger:
Good point.  I think it's ok to not remember state.  I have it working so that when mode ends it removes the mode's colors automatically.  I changed the `removeall` action to `remove_all` (to make it consistent with other actions in mpf that use underscores) and I added another action called `remove_mode` that will remove all the colors added by the calling mode (just in case you want to remove all colors at a point other than when they're automatically removed at mode stop).

Thanks for your input, it helps a lot!

I'll push these changes shortly.

jabdoa2:
> Edit: I just realized that the `ConfigPlayer` class has a `mode_stop` method. I'm guessing I could override that method and unload the blinkenlight when modes end. I would just need to keep track of which mode added which colors. So there's likely enough functionality already in `ConfigPlayer` to handle this correctly.

We already got a way to handle this. It is called "clear_context“. That is called when modes are unloaded, shows are stopped and in any other case. You can have a look at LightPlayer. It is similar to your approach but covers aboslutely everything in MPF. Also you probably do need the change in Blinkenlight with that.

densminger:
Got it. Thanks, Jan.  I changed it to use `clear_context` instead.  So blinkenlight colors are removed when the associated mode/context ends.  Example, mode1 adds blue and green, mode2 adds red.  When mode2 ends, red will automatically be removed from the blinkenlight, and when mode1 ends, blue and green will be removed.  However, a caveat to this is, the colors are also removed when a show ends.  So, if a show uses `blinkenlights:` to add a color to a blinkenlight, then the color will only flash while the show is active, and as soon as that show ends, the blinkenlight's colors will be removed.  I am guessing this is ok, just needs to be documented?  Let the user know that if they want a blinkenlight's color to remain active for a mode, the mode should add the color to the blinkenlight via a `blinkenlight_player` and not a through a show.  Speaking of which, I'm working on the documentation as well.

I also renamed the `key` parameter in `blinkenlight_player` to `label` instead, because I think that's clearer to the end-user as to its purpose.  Finally, I added an express config for `blinkenlight_player` which does the add action with a uuid/random label.

mpf/config_players/blinkenlight_player.py
```diff
@@ -0,0 +1,67 @@
...
+    config_file_section = 'blinkenlight_player'
+    show_section = 'blinkenlights'
+    machine_collection_name = 'blinkenlights'
+    allow_placeholders_in_keys = True
+
+    def play(self, settings, context, calling_context, priority=0, **kwargs):
+        """Adds/removes colors to blink."""
+        del kwargs
+
+        for blinkenlight, s in settings.items():
```
jabdoa2, line 19: This is kind of different than most of our other players. Not sure if better or worse. One issue might be consistency.

Usually you can add and remove stuff with or without a `key`. Here this is called `label`. However, here you kind of have to use a label or removing your entry will not work later on (because you do not know the uuid). Also what you can do here is to remove `labels` from other modes. `keys` are namespaced in other players so that you cannot have conflicts between modes. Then we got contexts and labels here. In other places we just map remember our keys insides a context dict. Here `blinkenlight` needs to know both `context` and `label`.

Any thoughts on this? I'm always open to suggestions for improvement.

densminger: So is your main concern that a mode can remove a color that was added by another mode if it uses the same label? My thought there is, that's kind of intended. Honestly, I think most of the time the user won't even use a `label` since the color is automatically removed when the mode ends. But it could be that you want to remove the color when a shot is hit or something.

jabdoa2: > So is your main concern that a mode can remove a color that was added by another mode if it uses the same label? My thought there is, that's kind of intended. Honestly, I think most of the time the user won't even use a `label` since the color is automatically removed when the mode ends. But it could be that you want to remove the color when a shot is hit or something.

I see that this might be a feature but it might as well confuse people. Conceptually, I think that it would be bad to design something which is added in one mode and removed in another one. That will lead to confusion and issues down the road.

In my opinion the main usecase would be this (which currently will not work this way):

```
blinkenlight_player:
  mode_my_mode_started:
    my_blinkenlight1: orange
  some_other_event:
    my_blinkenlight1: remove/stop    # some magic color. we use "stop" for lights
  some_other_event2:
    my_blinkenlight1:
      action: remove
```

I would simplify `blinkenlights` a bit to now know about contexts. Also I would use `key` instead of `label` for consistency with other players in MPF. If you want me to make those changes I can do that as well. We can do it on stream on Saturday if you like. Not 100% sure I will make it but I actually might.

atummons: As this is being discussed here, what is the future possibility of integrating this as a show?  I am thinking a shot profile call a show based on state.  Can I have a show for state 1, which adds to a blinkenlight with priority?  Then when it goes to an "off" state, I remove that key from the blinkenlight?

I don't think that would work today, and I don't think its necessary.  But for broader appeal and use, that would be a nice way to be able to call and use this feature.

jabdoa2: That should already work. Your profile will stop and start shows. Stop will call clear_context and remove the previous color. Guess we should add a test with a show but I am quite confident that it will just work.

jabdoa2: Just to make this clear: I am very happy with the code how it is. I would like to change `label` to `key` for consistency and unexpose the context. Thats about it. Config interface is hard to change later. Refactoring the internal code can be done later and nobody will notice.

atummons: Awesome. Well I can play around with it as I'm thinking and add tests as we go.
And as Jan said, awesome job! I think this will be a popular feature!

densminger: Hmm. Something to consider if we go with what you're suggesting. What happens if a mode adds several colors (not sure why a mode would but it's totally possible and is designed to allow this). Then what would happen if:

```
blinkenlight_player:
  some_event:
    my_bl: remove
```

Would this remove all colors from the mode? Just the latest one?

Also, if you want to add or change any of this on stream, please feel free!

densminger: > As this is being discussed here, what is the future possibility of integrating this as a show?  I am thinking a shot profile call a show based on state.  Can I have a show for state 1, which adds to a blinkenlight with priority?  Then when it goes to an "off" state, I remove that key from the blinkenlight?
> 
> 
> 
> I don't think that would work today, and I don't think its necessary.  But for broader appeal and use, that would be a nice way to be able to call and use this feature.

This wouldn't quite work because the blinkenlight will remove the show's colors when the show ends. So if your shot profile called a show which added a color, that color would only exist for the duration of the show.

The way you could do this now though, is to use blinkenlight_player to add a color when the shot is made:

```
blinkenlight_player:
  shot_left_ramp_hit:
    blinkenlight_left_ramp: red
```

atummons: Yeah, that's what I wasn't sure about. If I had a show similar to on, then it would never end. So theoretically it would be in that state until I told it to end that show, which would remove it.

Otherwise that was my though, I add it when mode starts, and then xx_lit_hit would be my trigger to remove it. I'm excited to test this through the scenarios!

densminger: > I add it when mode starts, and then xx_lit_hit would be my trigger to remove it. I'm excited to test this through the scenarios!

Yep. Just keep in mind that (as of now) if you want to remove the color yourself (as opposed to automatically removing the color when the mode ends) then you'll need to specify a `label` value when you add the color, then `remove` using that label later.

atummons: Yep, and that's consistent with keys to remove widgets and other players. So people should be able to follow that logic through to here with a nice write up in the docs. I'm more than willing to assist with that piece for nice we get the code finalized.

densminger: You can check out my PR on the mpf-docs repo. I've already added documentation for blinkenlight in that pull request. I'm getting a Travis CI error though and I'm not sure why.

jabdoa2: > Hmm. Something to consider if we go with what you're suggesting. What happens if a mode adds several colors (not sure why a mode would but it's totally possible and is designed to allow this). Then what would happen if:
> 
> ```
> blinkenlight_player:
>   some_event:
>     my_bl: remove
> ```
> 
> Would this remove all colors from the mode? Just the latest one?

In this case you should use keys. Otherwise the second color should remove the first one (because they got the same key). That is the logic for all other players at least. Except maybe in widget player where you can intentionally leak objects to add one widget multiple times (prone to memory leaks obviously).

> > As this is being discussed here, what is the future possibility of integrating this as a show?  I am thinking a shot profile call a show based on state.  Can I have a show for state 1, which adds to a blinkenlight with priority?  Then when it goes to an "off" state, I remove that key from the blinkenlight?
> > I don't think that would work today, and I don't think its necessary.  But for broader appeal and use, that would be a nice way to be able to call and use this feature.
> 
> This wouldn't quite work because the blinkenlight will remove the show's colors when the show ends. So if your shot profile called a show which added a color, that color would only exist for the duration of the show.

Basically, your color has to be included in all shows where it should exist. Similar to light_player behaviour.

mpf/devices/blinkenlight.py
```diff
@@ -0,0 +1,164 @@
...
+            self.add_color(color, label, priority, context)
+
+    def remove_all_colors(self):
+        """Remove all colors from the blinkenlight."""
+        self._colors.clear()
+        self.num_colors = 0
+        self.info_log('All colors removed')
+        self._restart()
+
+    def remove_color_with_label(self, label):
```
atummons: This needs changed to private _remove_color_with_label (or if it needs to be public, line 92 needs changed to remove the underscore.

mpf/config_players/blinkenlight_player.py
atummons: Played around with this and so far with basic tests its working great.  I tried adding it to shot_profiles, and it worked just as Jan said it would.  When I dug in, I did run into one error though.  Placeholders for blinkenlights work fine, but for color threw an error.
```
shows:
    bl_on:
      - duration: -1
        blinkenlights:
            (light): (color_1)

File "c:\users\atummons\desktop\davebranch\mpf\mpf\core\config_validator.py", line 683, in _validate_type_color return int(color[0]), int(color[1]), int(color[2])
ValueError: invalid literal for int() with base 10: '(color_1)'
```
Not sure why this works for shows with lights, but not blinkenlights.  I looked around, but don't really know what I was looking for.

jabdoa2: Good find. I think we need color_or_token as type to support that.

mpf/config_spec.yaml
```diff
@@ -331,6 +339,12 @@ fast_coils:
     hold_pwm_patter: single|str|None
 pkone_coils:
     recycle_ms: single|ms|None
+blinkenlight_player:
+    __valid_in__: machine, mode, show
+    __type__: config_player
+    action: single|enum(add,remove,remove_mode,remove_all)|add
+    label: single|str|None
+    color: single|color|ffffff
```
atummons: This should be changed to:     color: single|str|white
This will allow tokens to be passed in, in addition to color names of color codes.

mpf/config_players/blinkenlight_player.py
atummons: I made the comment in the config spec.  Didn't need to make it a token.  But in config spec it was a type color, but needs to be a type str.  This is the same as light_player.  I made the change local, and it worked as expected.  So I made a comment on that row for the change.

Not sure what's appropriate and good coding practices.  Should I make a push for that change to someone else's branch?  Or just note it for them?  Not sure what standard practice is, or if its all personal preference for Dave, since this is his branch.

mpf/config_spec.yaml
jabdoa2: No. This would change to `single|color_or_token|white`

jabdoa2: Looks like you are right. This does not yet exist. I guess we should create it (as we have it for floats or ints). That will move str parsing out of runtime in most cases.

mpf/config_players/blinkenlight_player.py
densminger: Go ahead and make the change if you want. I was holding off on making changes in case Jan decided to do a stream today, but generally I have no problems if you want to make changes.

jabdoa2: Maybe we can finish this baby together later on. Should be fun.

atummons: Its looking super awesome.  I put it through a few more paces this morning, and worked just as expected!  So I think a few of those cleanup things from this thread, and if we want to implement the color_or_token method and we should be in business!

densminger: Woohoo!

atummons: I put together a fix for a method call that was wrong, and added color_or_token, but saying I'm getting permission denied to push to Dave's branch.  So is best option to just post changes here and someone with authority can make them?  Or is there a way I can push against it and I'm trying to push to Dave's branch, instead of this PR?

densminger: > I put together a fix for a method call that was wrong, and added color_or_token, but saying I'm getting permission denied to push to Dave's branch.  So is best option to just post changes here and someone with authority can make them?  Or is there a way I can push against it and I'm trying to push to Dave's branch, instead of this PR?

I have it set to allow changes from mpf maintainers (if you have write access to mpf you can change the PR) but I don't think I can allow changes from anyone else? My GitHub knowledge is lacking here lol

atummons: Yeah, I'm not an mpf maintainer, I don't believe.  I have your branch checked out on local, and am pushing to your local.  I think you have to add me to your repo as a collaborator (don't do this, I have limited knowledge as well and am not to be trusted on your branches!).  So I can shoot it over (its only about 10 lines) if we get together to discuss final stuff.

mpf/config_spec.yaml
atummons: Can't publish changes, but here's what they are.

Update 347 to:     color: single|color_or_token|ffffff.

Then in config_validator:
Add line 81:             "color_or_token": self._validate_type_or_token(self._validate_type_color),
Line 669: Swap above the two lines of comments (#) for the following:
    def _validate_type_color(self, item, validation_failure_info, param=None):
        assert not param
        if item is not None:
            try:
                if isinstance(item, tuple):
                    if len(item) != 3:
                        self.validation_error(item, validation_failure_info, "Color needs three components")
                    return item
            except ValueError:
                self.validation_error(item, validation_failure_info, "Cannot convert value to color.", 11)

mpf/devices/blinkenlight.py
atummons: Blinkenlight_player calls this, so it looks like the call on 92 needs to be global.

densminger: Good catch.  That would have been a runtime error.  Made me realize I didn't have a unit test for this.  I've fixed it and added a test.  Will push in next update.

mpf/config_spec.yaml
densminger: What's the purpose of `param=None` and `assert not param`?  These aren't in my version of config_validator and I'm not sure the purpose.  I'm not sure what's going on here in general.

atummons: Jan would have to answer the specifics for you here. But we switched to allowing tokens, and it runs through the token validated to see is it a token on a color, in this case. If it's a token, we know that, so we stop before the validation breaks. If it's not a token, we go through. Without this argument it breaks here as it expects three arguments, but gets 4 due to the token validation.

This is based on the other token_or_xx methods here.

densminger: Ah ok. I'll take a look when I get home and push the changes later tonight.

densminger:
@atummons Ok, both of your fixes have been pushed.  I changed the `_validate_type_color` method a little from what you suggested since it didn't require everything you said.  I added tests for both the bug/typo you found, as well as the tokens, so these things should continue to work in the future.

mpf/config_players/blinkenlight_player.py
densminger: > I would like to change `label` to `key` for consistency and unexpose the context.

I renamed `label` back to `key`.  As far as "unexposed the context" I'm not sure what you would like to do there, exactly.  So I'll let you make those changes, if that's ok.

densminger:
Renamed `label` back to `key` as per Jan's suggestion. @atummons, this will mean you'll need to change your code that you're using for testing as well, just fyi.

atummons:
Awesome work! I'll get latest on your repo and continue to play around with my test machine.

mpf/config_players/blinkenlight_player.py
atummons: Pulled this in and all is working as expected with both of your updates.

densminger (commented):
Looks great! Everything works on my end.  It's ready to go as far as I'm concerned.

Edit: So it looks like the "stop/remove" express config option will only remove colors that were added without a key (that is, colors that were added with the express config method, or colors that were added without specifying a key).  This makes sense, I just need to update the documentation to reflect that.

jabdoa2:
> Looks great! Everything works on my end. It's ready to go as far as I'm concerned.

Thanks for testing and confirming. This has been an awesome collaboration! Thanks @densminger @atummons .



> Edit: So it looks like the "stop/remove" express config option will only remove colors that were added without a key (that is, colors that were added with the express config method, or colors that were added without specifying a key). This makes sense, I just need to update the documentation to reflect that.

Yeah exactly. My theory is that 99% of the users will have maximum one color per mode.

PR #1592, opened 2021-08-10
Adds Multiball Ball Save for Add A Ball

atummons:
This adds a unique ball save feature for add a ball.  This uses the same
event calls, so you can trigger your shows off the same events for
normal ball save and add a ball ball save.  The add a ball ball save
will not run (or add time) if the ball save is already running.  But if
the normal ball save has expired, it will start this ball save at the
add a ball configured times for shoot again, hurry up, and grace period.

### DIFF
--- a/mpf/devices/multiball.py
+++ b/mpf/devices/multiball.py
@@ -163,8 +163,7 @@ class Multiball(EnableDisableMixin, SystemWideDevice, ModeDevice):
         self._start_shoot_again(shoot_again_ms, grace_period_ms, hurry_up_time_ms)
 
     def _start_shoot_again(self, shoot_again_ms, grace_period_ms, hurry_up_time_ms):
-        """Sets the event callbacks for shoot again, grace period, and hurry up,
-           if values above 0 are provided
+        """Set callbacks for shoot again, grace period, and hurry up, if values above 0 are provided.
 
         This is started for both beginning multiball ball save and add a ball ball save
         """
--- a/mpf/devices/multiball.py
+++ b/mpf/devices/multiball.py
@@ -302,8 +302,7 @@ class Multiball(EnableDisableMixin, SystemWideDevice, ModeDevice):
         if self.shoot_again:
             # if main ball save timer is running, don't run this timer
             return
-        else:
-            self.shoot_again = True
+        self.shoot_again = True
 
         shoot_again_ms = self.config['add_a_ball_shoot_again'].evaluate([])
         if not shoot_again_ms:
--- a/mpf/devices/multiball.py
+++ b/mpf/devices/multiball.py
@@ -291,6 +291,55 @@ class Multiball(EnableDisableMixin, SystemWideDevice, ModeDevice):
             self.balls_added_live += 1
             self.machine.game.balls_in_play += 1
             self.source_playfield.add_ball(balls=1)
+            self.add_a_ball_timer_start()
+
+    def add_a_ball_timer_start(self) -> None:
+        """Start the timer for add a ball ball save.
+
+        This is started when multiball add a ball is triggered if configured,
+        and the default timer is not still running.
+        """
+        if self.shoot_again:
+            # if main ball save timer is running, don't run this timer
+            return
+        else:
+            self.shoot_again = True
+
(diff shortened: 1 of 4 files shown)


mpf/devices/multiball.py
```diff
@@ -291,6 +291,54 @@ def add_a_ball(self):
             self.balls_added_live += 1
             self.machine.game.balls_in_play += 1
             self.source_playfield.add_ball(balls=1)
+            self.add_a_ball_timer_start()
+
+    def add_a_ball_timer_start(self) -> None:
```
jabdoa2: guess this should be `_add_a_ball_timer_start` as it should be private and not called from outside this class.

mpf/devices/multiball.py
```diff
@@ -291,6 +291,54 @@ def add_a_ball(self):
...
+        """
+        if self.shoot_again:
+            # if main ball save timer is running, don't run this timer
+            return
+        self.shoot_again = True
+
+        shoot_again_ms = self.config['add_a_ball_shoot_again'].evaluate([])
+        if not shoot_again_ms:
+            # No shoot again. Just stop multiball right away
+            self.stop()
```
jabdoa2, line 317: I guess you also want to add `return` here. Otherwise, we will still post the events below.

mpf/devices/multiball.py
```diff
@@ -291,6 +291,54 @@ def add_a_ball(self):
...
+                                            priority=1000)
+
+        self.machine.events.post('ball_save_{}_add_a_ball_timer_start'.format(self.name))
+        '''event: ball_save_(name)_add_a_ball_timer_start
+        desc: The multiball add a ball ball save called (name) has just start its countdown timer.
+        '''
+
+        grace_period_ms = self.config['add_a_ball_grace_period'].evaluate([])
+        hurry_up_time_ms = self.config['add_a_ball_hurry_up_time'].evaluate([])
+        if shoot_again_ms > 0:
```
jabdoa2: If those are the same handlers as in the normal shoot again case you could refactor this into a separate method with takes `grace_period_ms` and  `hurry_up_time_ms` as methods.

jabdoa2:
Quite a nice addition. I love it! Left some small nits.

mpf/devices/multiball.py
atummons: Agreed, and fixed in my next push.

mpf/devices/multiball.py
atummons: Agreed.  I merged the two methods for normal multiball into one, since there is less to handle, but screwed that up.  So I have added the return statement, and then removed the else, as all things past will run if they don't return.

mpf/devices/multiball.py
atummons: Not 100% sure I understand how to execute the ask here.  They are different config section.  You have grace_period for normal ball save and add_a_ball_grace_period for the add a ball (so that they can be different values).  We then need to calculate them, and add handlers if needed.

If that means we are good as is, then I will get the other stuff pushed.  If not, I think you are meaning create a method that takes a config section and performs the conversion and returns the ms.  So instead of doing this inline, we would call the method to do it?

mpf/devices/multiball.py
atummons: While I am in this file.  The timer_start method for the regular ball save is not private.  Do we want it to be?  If so, I'll make that change as well in my push.

jabdoa2: Good find. I would make that private as well.

mpf/devices/multiball.py
jabdoa2: I propose a method such as:

```
    def _start_shoot_again(self, shoot_again_ms, hurry_up_time_ms):
```
This would then just contain the line below this comment and be called `self._start_shoot_again(shoot_again_ms, hurry_up_time_ms)` from there and from `_timer_start`. Does that make sense?

atummons: Got it, that makes sense.  Should be there now where both places can call the one method, and they evaluate the variables first from the config, but then pass them in for handlers to be setup.  Good idea to reuse code and make it easier to maintain going forward.

PR #1594, opened 2021-08-24
Adds a default show bl_color for blinkenlights

atummons:
The bl_color show takes a token for the light (led, light, blinkenlight,
etc.) as well as the color that is being passed.  It calls to the
blinkenlights player only, but can take any of the standard light
tokens.  It also includes some error handling for blinkenlight_player to
verify its a type, and not one of the blank strings when adding or
removing.

### DIFF
--- a/mpf/config_players/blinkenlight_player.py
+++ b/mpf/config_players/blinkenlight_player.py
@@ -1,6 +1,7 @@
 """Blinkenlight config player."""
 from mpf.config_players.device_config_player import DeviceConfigPlayer
 from mpf.devices.blinkenlight import Blinkenlight
+from typing import Union
 
 
 class BlinkenlightPlayer(DeviceConfigPlayer):
@@ -17,28 +18,29 @@ class BlinkenlightPlayer(DeviceConfigPlayer):
         del kwargs
 
         for blinkenlight, s in settings.items():
-            action = s['action']
-            key = context + s['key'] if s["key"] else context
-            if action == 'add':
-                self._add_color(blinkenlight, s['color'], key, priority + s['priority'])
-                self._get_instance_dict(context)[(key, blinkenlight)] = blinkenlight
-            elif action == 'remove':
-                self._remove_color(blinkenlight, key)
-                try:
-                    del self._get_instance_dict(context)[(key, blinkenlight)]
-                except KeyError:
-                    pass
-            elif action == 'remove_mode':
-                self.clear_context(context)
-            elif action == 'remove_all':
-                self._remove_all_colors(blinkenlight)
-            else:
-                raise AssertionError("Unknown action {}".format(action))
+            if not isinstance(blinkenlight, str):
+                action = s['action']
+                key = context + s['key'] if s["key"] else context
+                if action == 'add':
+                    self._add_color(blinkenlight, s['color'], key, priority + s['priority'])
+                    self._get_instance_dict(context)[(key, blinkenlight)] = blinkenlight
+                elif action == 'remove':
+                    self._remove_color(blinkenlight, key)
+                    try:
+                        del self._get_instance_dict(context)[(key, blinkenlight)]
+                    except KeyError:
+                        pass
+                elif action == 'remove_mode':
+                    self.clear_context(context)
+                elif action == 'remove_all':
+                    self._remove_all_colors(blinkenlight)
+                else:
+                    raise AssertionError("Unknown action {}".format(action))
 
     @staticmethod
-    def _add_color(blinkenlight: Blinkenlight, color, key, priority):
+    def _add_color(blinkenlight: Union[Blinkenlight,str], color, key, priority):
         """Instructs a blinkenlight to add a color to its list of colors."""
-        if blinkenlight is None or isinstance(blinkenlight, str):
+        if blinkenlight is None:
             return
         blinkenlight.add_color(color, key, priority)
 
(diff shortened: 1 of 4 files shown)


densminger:
Nice! I can see where this would be useful

atummons:
Yeah, it allows you to use shot profiles, which takes a ton of logic out. So you just set the show in profile, then set the two tokens (blinker light, and color). Then magic happens (or many things I can't explain!)

mpf/config_players/blinkenlight_player.py
```diff
@@ -38,7 +38,7 @@ def play(self, settings, context, calling_context, priority=0, **kwargs):
     @staticmethod
     def _add_color(blinkenlight: Blinkenlight, color, key, priority):
         """Instructs a blinkenlight to add a color to its list of colors."""
-        if blinkenlight is None:
+        if blinkenlight is None or isinstance(blinkenlight, str):
```
jabdoa2: I think in other places we check that the string starts with `(` and ends with `)` to make sure that we only ignore tokens.

Bonus question: When is this ever None? That case does not make sense to me. Can we remove it?

mpf/config_players/blinkenlight_player.py
```diff
@@ -38,7 +38,7 @@ def play(self, settings, context, calling_context, priority=0, **kwargs):
     @staticmethod
     def _add_color(blinkenlight: Blinkenlight, color, key, priority):
```
jabdoa2, line 40: Signature should become:
`def _add_color(blinkenlight: Union[Blinkenlight,str], color, key, priority):`

mpf/config_players/blinkenlight_player.py
```diff
@@ -59,7 +59,8 @@ def _remove_color(blinkenlight, key):
     def clear_context(self, context):
         """Clear the context. In our case, this means remove the mode colors from all blinkenlights."""
         for (key, _), blinkenlight in self._get_instance_dict(context).items():
-            blinkenlight.remove_color_with_key(key)
+            if not isinstance(blinkenlight, str):
```
jabdoa2: This should not be needed. We should not add stuff to `instance_dict` if we did not add any colors to it.

mpf/config_players/blinkenlight_player.py
jabdoa2: I suggest you move this to line 20 as we do not want to do anything with those kinds of keys.

mpf/config_players/blinkenlight_player.py
atummons: It was needed due to previous error handling, but with your suggestions is no longer needed and has been removed.

mpf/config_players/blinkenlight_player.py
atummons: I moved it to line 20, and that seems to still work.  I looked at how light_player is doing it, and it would be significant refactoring from what we have today, as the rest of this method would have to be moved to a new method, and cleaned up a bit.  I did an ugly solution, which was duplicating all the code to pass a different value.  When I did this, I could not get it to actually hit.  Because its saying the type is a str, which will always fail the next tests anyways.  The feeds in are a bit different, so I think we skip some of those things.  If we wanted to be more robust, we would need to refactor this to follow a similar path of light_player.  I don't know if that's on your list now, or wait to see if there is an edge case and we fix the potential bug then.

As for the bonus question... no idea.  Its in three methods here, and they all still worked when I removed it.  So unless you or @densminger  see a reason to keep them, I can do another push to remove them.

densminger: Nope, it's never None.  I put this in there as a paranoid check, but it's not necessary.

mpf/config_players/blinkenlight_player.py
```diff
@@ -17,26 +18,27 @@ def play(self, settings, context, calling_context, priority=0, **kwargs):
...
+                elif action == 'remove_mode':
+                    self.clear_context(context)
+                elif action == 'remove_all':
+                    self._remove_all_colors(blinkenlight)
+                else:
+                    raise AssertionError("Unknown action {}".format(action))
 
     @staticmethod
-    def _add_color(blinkenlight: Blinkenlight, color, key, priority):
+    def _add_color(blinkenlight: Union[Blinkenlight,str], color, key, priority):
```
densminger: Will `blinkenlight` ever be a `str` based on the check on line 21?

mpf/config_players/blinkenlight_player.py
atummons: Sounds good.  I had to do another push for some Travis CI fixes, so I went ahead and included this update as well.

mpf/config_players/blinkenlight_player.py
atummons: I guess not really going forward.  Unless Jan has thoughts in regard to some refactoring mentioned above, but I can't find a case where that is hit in my whole game for light_player, so it might be unnecessary handling, and thus we could revert.

atummons: I removed it on my local and did not have any issues, as we are weeding out all strings up above, so it has to be type Blinkenlight to get here.  So my vote would be to revert back to just blinkenlight (and drop the import for union).

jabdoa2: @densminger is right. We do not need it when moving the check up. I made this comment before looking at the rest of the code. It is also just a type hint so it does nothing at runtime.

PR #1597, opened 2021-09-12
Service list custom values for list_coils etc

avanwinkle:
### Parameter Values for Service Lists
This PR adds an optional `values` parameter when requesting `list_coils`, `list_lights`, and `list_switches`. The parameter allows a BCP client to specify which device properties it needs included in the list (and in a predictable order), so that Service modes can display a full list of devices without unnecessary detail.

If no `values` kwarg is supplied, the response payload uses the existing values as default, so this change is fully backward compatible.

### DIFF
--- a/mpf/core/bcp/bcp_interface.py
+++ b/mpf/core/bcp/bcp_interface.py
@@ -131,17 +131,19 @@ class BcpInterface(MpfController):
         elif subcommand == "stop":
             await self._service_stop(client)
         elif subcommand == "list_switches":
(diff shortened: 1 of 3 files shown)


avanwinkle:
Added some tests to cover the expected responses. I don't have an example of this working in a real MC (well, not an MC that's ready to be seen yet), but the use case is for displaying the full list of devices in a scrolling view, rather than just the name of the currently selected one.

The default responses don't include the `label` so I initially just added that, but then I wasn't thrilled with the payload being 3-4x larger than it needed to be. Hence this solution.

PR #1598, opened 2021-09-22
Production bundles: new options for MC and target machine path

avanwinkle:
This PR makes three additions to the fabulous production bundle behavior.

### Production bundle building without MC
This PR adds support for the argument `-b` when running `mpf build production_bundle`, which builds an MPF bundle only. The `-b` arg was chosen because that's the argument used to run MPF without connecting to a BCP host—effectively, how you run MPF without an MC.

### Explicit machine paths for production bundles
A side-effect of how the production bundles are built—loading all the config files and pickling them—is that the bundle is saved with the `machine_path` of the machine that generated the bundle. This causes trouble when trying to distribute a bundle to other machines with different file structures. This PR adds a new command argument `--dest-path=<machine_path>` which will generate the config data and then overwrite the `machine_path` with the explicitly-provided path.

The machine path is a private variable, so this PR also adds a method `set_machine_path(pathname)` to the config loader.

### Running production bundles without _/config_
This PR adds support for running an MPF machine from a production bundle without any other file structure requirements. Currently, MPF will always check for a machine folder by looking for the _/config_ folder. When running production bundles this folder is not used, so this PR makes it not required.

### DIFF
--- a/mpf/commands/__init__.py
+++ b/mpf/commands/__init__.py
@@ -7,6 +7,7 @@ from pkg_resources import iter_entry_points
 
 
 import mpf.core
+from mpf.core.config_loader import ProductionConfigLoader
 from mpf._version import version
 
 EXAMPLES_FOLDER = 'examples'
@@ -24,13 +25,14 @@ class MpfCommandLineParser:
         self.mpf_path = os.path.abspath(os.path.join(mpf.core.__path__[0],
                                                      os.pardir))
 
-    def get_machine_path(self, machine_path_hint=None) -> str:
+    def get_machine_path(self, machine_path_hint=None, is_production=False) -> str:
         """Find the full machine path based on the current directory and option hint.
 
         Args:
         ----
             machine_path_hint: Helps MPF locate the machine path. If None,
                 the 'config' folder in the current working directory is used.
+            is_production: True when the game is run from a production bundle.
 
         Returns a string of full path of the machine folder that was located.
         """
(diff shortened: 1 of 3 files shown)


mpf/commands/__init__.py
```diff
@@ -98,9 +111,13 @@ def parse_args(self):
 
         if len(self.argv) > 1 and self.argv[1].startswith('-'):
             self.argv.insert(1, None)
-
         args, remaining_args = parser.parse_known_args(self.argv[1:])
-        machine_path = self.get_machine_path(args.machine_path)
+
+        # If the production arg was pulled here, re-insert it to the remaining args
+        # so that the game parser also knows to run in production mode.
+        if args.production:
```
avanwinkle, line 118: A quirk of the `parse_known_args()` method is that it consumes the known args and only passes unknown args to the next consumer. In this case I need both _init.py_ and _game.py_ to consume the `-P` arg.

I looked through the parser documentation and didn't see any way to keep an arg in both places, so here I re-add it so that _game.py_ can consume it. Totally open to a better approach!

avanwinkle:
@jabdoa2 I'm getting consistent integration test failures on Python 3.8 with unfinished futures. Could this be related to your recent work in the area? (https://github.com/missionpinball/mpf/commit/162f4e484a068b9f15308ff27311b5bc86ec107a)

jabdoa2:
Yeah. I will have a look. I catch some cases in tests where we do not cancel futures properly.

jabdoa2:
I need to look further into this. Feel free to merge this anyway. Something looks slightly different in Python 3.8 on Windows. The warning is harmless on its own.

PR #1599, opened 2021-09-29
Add 'set_tick_interval' control event for Timers, with event kwargs

avanwinkle:
This PR adds a new control event to the Timer device, `set_tick_interval`. It differs from the existing `change_tick_interval` because it uses absolute values instead of relative, and it leverages the existing `def set_tick_interval(self):` method that the Timer has (currently used only by the `reset_tick_interval` event).

The new control event also supports kwarg placeholders, so the triggering event can specify a new absolute tick interval and the timer will switch to that interval. Yay!

### DIFF
--- a/mpf/devices/timer.py
+++ b/mpf/devices/timer.py
@@ -157,9 +157,13 @@ class Timer(ModeDevice):
                 handler = self.change_tick_interval
                 kwargs = {'change': entry['value']}
 
+            elif entry['action'] == 'set_tick_interval':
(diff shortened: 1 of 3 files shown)


jabdoa2:
Good one! This changes placeholder evaluation to runtime. Guess that makes sense here.

avanwinkle:
Yeah, I feel okay about the runtime evaluation because it's only done when the event itself is triggered, which shouldn't be very frequent.

PR #1600, opened 2021-10-01
handle autofire hw rules in a switch matrix

cobra18t:
matrix switch based hw rules need to have the matrix columns mirrored since opp has reverse column order


jabdoa2:
Thanks!

jabdoa2:
Maybe we should add a testcase for those numbers as well. Otherwise we might regress this in the future without noticing.

cobra18t:
If you give those functions a switch_num of 81 (0x51), it should send a command referencing switch 41 (0x29).

PR #1602, opened 2021-10-02
Service mode_setting to prevent device list sorting

avanwinkle:
This PR adds a new `mode_settings:` option to the Service mode to choose between sorting device lists by their hardware number (default, current behavior) or preserving the order listed in the config files.

When the mode setting `sort_devices_by_number` is `false`, devices listed will preserve the config file order. This applies both to the iteration through devices in the utilities menu (coil test and light test) as well as the list returned to BCP clients using the `list_coils`, `list_switches` and `list_lights` commands.

By default the sort setting is true, so this change causes no regressions for existing machines that do not specify a value.

### DIFF
--- a/mpf/core/bcp/bcp_interface.py
+++ b/mpf/core/bcp/bcp_interface.py
@@ -132,19 +132,22 @@ class BcpInterface(MpfController):
             await self._service_stop(client)
         elif subcommand == "list_switches":
             values = kwargs.get("values") and kwargs["values"].split(",")
+            do_sort = kwargs.get("sort")
             self.machine.bcp.transport.send_to_client(client, "list_switches",
                                                       switches=[self._switch_body(s[0], s[1], values) for s in
-                                                                self.machine.service.get_switch_map()])
+                                                                self.machine.service.get_switch_map(do_sort)])
         elif subcommand == "list_coils":
             values = kwargs.get("values") and kwargs["values"].split(",")
+            do_sort = kwargs.get("sort")
             self.machine.bcp.transport.send_to_client(client, "list_coils",
                                                       coils=[self._coil_body(s[0], s[1], values) for s in
-                                                             self.machine.service.get_coil_map()])
+                                                             self.machine.service.get_coil_map(do_sort)])
         elif subcommand == "list_lights":
             values = kwargs.get("values") and kwargs["values"].split(",")
+            do_sort = kwargs.get("sort")
             self.machine.bcp.transport.send_to_client(client, "list_lights",
                                                       lights=[self._light_body(s[0], s[1], values)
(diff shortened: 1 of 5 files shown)


jabdoa2:
What is the use-case for this? I mean we could also add a "service_sort" parameter to devices if we want to control it explicitly.

avanwinkle:
The use case is for the coil and light tests. The numbered order of devices isn't intuitive to the operator, so cycling through the list looking for a specific light or coil—you have no idea when it's going to show up.

With this option to preserve the config order, relevant coils and lights are in sequence (all the drop targets in a row, the pop bumpers in a row, the lane lights in a row). This makes it much easier and faster to find what you're after. And with my previous improvements to the service menu I'm able to list all the coils/lights and scroll through them, so putting similar things together just "looks" better.

I thought about exposing a `sort` value, but couldn't think of a use case. We've got ordered by address (good for tracing issues) and the user-provided order, which seems good to me! 😄

atummons:
The only other option I could see of value would be alphabetical. But I would argue if that's what you want, then organize them that way in your config.

jabdoa2:
Ok if this is what you guys want: Do we also want to use it in MPF service?

jabdoa2:
@avanwinkle saw my last question? I guess you are developing some external (private?) tool. Please don't forget our built-in mechanisms.

avanwinkle:
Yes! I did, but I've never used MPF service before so I need to ramp up on it before I'm able to port this functionality ;)

This isn't for a private tool, but for the Godot MC I've been working on that has enough screen size to display the full list of coils/lights during service mode. The sorting makes it much easier to navigate, per the attached (ugly and very early-draft) screenshots.

Will reach out to you soon about making a new repo for the Godot BCP Server so the functionality can be shipped to the Godot Asset Library, but there's still much work to do!

coil_test_no_sort
_Without sorting, coils are listed in an unintuitive order_

coil_test_with_sort
_With sorting, coils are listed in a logical order_

jabdoa2:
That is quite cool. Ok I see the usecase for that. Do you intent to replace/reimplement the MC monitor features in godot? Guess all those things are a bit fuzzy because depending on the usecase we need similar functionality in multiple places.

avanwinkle:
Right now the Godot setup is very similar to the Unity setup, where the BCP interface creates communication with MPF but all of the MC logic runs like a "game" that's built natively and unique to the machine. I think some UI elements (like the service menu, attract mode, and bonus mode) could have templates to get new users up and running, but the majority of the UI will be built using Godot itself rather than with MC-like configs.

But it's all very new and I'm open to ideas on how to make it more accessible and a great solution for anyone willing to learn Godot!

PR #1603, opened 2021-10-05
Support for FAST pulse-with-hold, set one-shot bit on pulse commands

avanwinkle:
This PR makes improvements to the driver `pulse()` command for FAST hardware. Includes a test!

### Pulse with hold
This PR adds two platform settings to FAST drivers: `pulse_hold_power` and `pulse_hold_ms`. When provided, these parameters are added to the hardware pulse command. FAST hardware supports a two-phase pulse action, with an initial pulse at PWM1 and a hold at PWM2. Using the hardware commands for this behavior is safer and more time-accurate than a pulse-and-enable followed by a software delay and disable.

### One-shot bit on pulse
This PR changes the pulse command from `81` to `89`. In FAST driver commands the `1 << 3` bit indicates to "one-shot" the driver after the configuration is written. This allows the configuration and pulse to be sent in a single command, which eliminates the need for a subsequent `TN:NN,01` command (which adds network traffic and can sometimes be missed by the hardware).

As a result, the pulse method is updated to send *either* the pulse-and-one-shot command *or* the manual trigger command, not both.

Here we go!

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -298,6 +298,8 @@ coils:
     default_pulse_ms: single|template_ms|None
     default_pulse_power: single|float(0,1)|None
     default_hold_power: single|float(0,1)|None
+    pulse_hold_power: single|float|None
+    pulse_hold_ms: single|ms|None
     max_pulse_ms: single|ms|None
     max_pulse_power: single|float(0,1)|1.0
     max_hold_power: single|float(0,1)|None
@@ -342,8 +344,6 @@ opp_coils:
 fast_coils:
     connection: single|enum(network,local,auto)|auto
     recycle_ms: single|ms|None
-    pulse_hold_power: single|float|None
-    pulse_hold_ms: single|ms|None
     hold_pwm_patter: single|str|None
 pkone_coils:
     recycle_ms: single|ms|None
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -342,6 +342,8 @@ opp_coils:
 fast_coils:
     connection: single|enum(network,local,auto)|auto
     recycle_ms: single|ms|None
+    pulse_hold_power: single|float|None
+    pulse_hold_ms: single|ms|None
     hold_pwm_patter: single|str|None
 pkone_coils:
     recycle_ms: single|ms|None
--- a/mpf/core/platform.py
+++ b/mpf/core/platform.py
@@ -500,7 +500,8 @@ class SwitchPlatform(BasePlatform, metaclass=abc.ABCMeta):
 SwitchSettings = namedtuple("SwitchSettings", ["hw_switch", "invert", "debounce"])
 DriverSettings = namedtuple("DriverSettings", ["hw_driver", "pulse_settings", "hold_settings", "recycle"])
 DriverConfig = namedtuple("DriverConfig", ["name", "default_pulse_ms", "default_pulse_power", "default_hold_power",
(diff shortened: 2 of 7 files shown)


jabdoa2:
Spike can do this as well. P/P3-Roc might also be able to do it. Do we have a general use-case for this? I have never seen this used in a commercial game. Where do we want it? Is this "timed enable"? That might be something which we could use in a few places (i.e. eject for older troughs).

avanwinkle:
If multiple platforms can do it, I think it would definitely make sense to have it be general use. 

My current use case is on drop targets, which require quite a lot of time (up to 200ms) to fully restore. I tried using just the pulse, but even at 200ms it took >80% PWM to successfully reset. With this change, I can pulse for 30-40ms and hold at 25% for 120-140ms and they reliably reset.

I had some similar behavior in Mass Effect, but I used software timers to enable and disable the coils. It was functional, but the timing wasn't as precise as I would have liked.

Happy to refactor to be more general use!

jabdoa2:
Sorry my late reply! I guess we could add a generic "enable_timed" command to our drivers. What do you think? We could default implement it using a delay and platforms which support this could use it. Would definitely be useful for drops and ball devices. Probably for more.

mpf/config_spec.yaml
```diff
@@ -299,6 +299,8 @@ coils:
     default_pulse_ms: single|template_ms|None
     default_pulse_power: single|float(0,1)|None
     default_hold_power: single|float(0,1)|None
+    pulse_hold_power: single|float|None
```
jabdoa2, line 302: Not sure if we need those at all anymore. I propose to add `timed_enable` as a new command to coils. We should add that to `Driver` and `coil_player`. Guess we could call it a in few places such as drop targets or ball devices.

avanwinkle: Can you clarify how the `timed_enable` would work? Would it use the `default_hold_power`, assuming that a given coil is not going to want both enable functionality _and_ pulse-with-hold functionality (or that it would use the same power in both situations)? And where to specify how long the enable should last?

jabdoa2: My idea would be:

A new method `timed_enable` similar to `enable` in driver (https://github.com/missionpinball/mpf/blob/dev/mpf/devices/driver.py#L193). 
```
    def timed_enable(self, enable_duration_ms, pulse_ms: int = None, pulse_power: float = None, hold_power: float = None, max_wait_ms: int = None):
```
I would make `enable_duration_ms` a required parameter here. I guess I would also include the pulse inside here (otherwise it should be called `hold_duration_ms`). This would then be called from coil_player or other devices. Not sure if we would need a default for this.

This method would call into the platform and catch `NotImplementedError` or alternatively check in `features` if this is supported. If not we would emulate it using a delay. 

Thoughts? I'm open to other suggestions.

avanwinkle: Interesting approach! Definitely makes sense for coil_player scenarios. For other scenarios where this behavior would be used in place of a default pulse (like a drop target reset or diverter), would we check for `timed_enable()` before using `pulse()`, or configure the coil somehow to use the timed_enable _instead_ of pulse?

jabdoa2: I would make this an explicit option in drop targets. Might actually become the default though in this case.

avanwinkle:
Closing this PR in favor of the abstract approach in https://github.com/missionpinball/mpf/pull/1613.

PR #1606, opened 2021-10-13
Posts update events for counter control_events

atummons:
When an control_event is posted for a counter, it will now post the
logicblock_{name}_update code, so that you can trigger off of that event
happening, similar to if a hit event is registered.


mpf/tests/test_LogicBlocks.py
```diff
@@ -385,35 +387,43 @@ def reset_event_mocks():
...
         # Test subtraction
         reset_event_mocks()
         self.post_event("counter6_count")
         self.assertEventCalledWith("counter6_hit", count=1, hits=1, remaining=9)
         self.post_event("reduce_counter6_5")
+        self.assertEventCalled("logicblock_counter6_updated")
         self.post_event("counter6_count")
         self.assertEventCalledWith("counter6_hit", count=-3, hits=-3, remaining=13)
         self.post_event("reduce_counter6_3")
+        self.assertEventCalled("logicblock_counter6_updated")
```
jabdoa2, line 426: You probably want to call reset_event_mocks after each assertion. Otherwise, this will stay true.

jabdoa2:
Looks good with the test change from above. Code smells are not on you. I will have a look later. Just ignore them for now.

mpf/tests/test_LogicBlocks.py
atummons: Did not know that!  I got them added in, and had some false positives that were corrected to NOTCalled.  I had to remove the check from one place because it was resetting the counter being completed twice, but we have already tested for that case on the update, so shouldn't lead to any breaks.

jabdoa2:
Thanks!

PR #1609, opened 2021-10-17
Add check in RGBColor constructor to account for integer colors

seanirby:
I ran into this issue while trying to add some particular colors to my `named_colors` configuration.  Values such as `990000` will be interpreted by MPFs YamlInterface as an integer rather than a string and will cause the following error when an RGBColor is instantiated with an integer value:

```
  File "/home/sean/dev/mpf/mpf/tests/test_RGBColor.py", line 34, in test_int_color
    color = RGBColor(990000)
  File "/home/sean/dev/mpf/mpf/core/rgb_color.py", line 181, in __init__
    self._color = (color[0], color[1], color[2])
TypeError: 'int' object is not subscriptable
```

Added a check in the constructor of RGBColor to fix this particular issue.

### DIFF
--- a/mpf/core/rgb_color.py
+++ b/mpf/core/rgb_color.py
@@ -169,14 +169,18 @@ class RGBColor:
 
     __slots__ = ["_color"]
 
-    def __init__(self, color: Union["RGBColor", str, int, List[int], Tuple[int, int, int]] = None) -> None:
+    def __init__(self, color: Union["RGBColor", str, List[int], Tuple[int, int, int]] = None) -> None:
         """Initialise color."""
         if isinstance(color, RGBColor):
             self._color = color.rgb
         elif isinstance(color, str):
             self._color = RGBColor.string_to_rgb(color)
         elif isinstance(color, int):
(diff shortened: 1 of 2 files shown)


jabdoa2:
Ideally you quote colors. We had similar issues with octal encoding such as "011111". Guess this is still a stop gap solution for some cases. Maybe we should warn users? Or raise a nice error?

seanirby:
> Ideally you quote colors. We had similar issues with octal encoding such as "011111". Guess this is still a stop gap solution for some cases. Maybe we should warn users? Or raise a nice error?

Ah OK.  The doc section I was referencing wasn't quoting the colors so that threw me off: https://docs.missionpinball.org/en/0.54/config/named_colors.html

I'll update to raise an error and also update that doc section.

jabdoa2:
Thanks for taking care!

seanirby:
Updated to output an error message instead. Sorry for the delay @jabdoa2, forgot about this one.

jabdoa2:
Awesome. Thanks!

PR #1610, opened 2021-10-27
prevent ball search from marking playfield active

cobra18t:
Added ignore_switch_ms to ball search resets to prevent a ball search from cancelling itself because of erroneous switch reads during reset.

### DIFF
--- a/mpf/devices/drop_target.py
+++ b/mpf/devices/drop_target.py
@@ -79,6 +79,7 @@ class DropTarget(SystemWideDevice):
 
     def _ball_search_phase1(self):
         if not self.complete and self.reset_coil:
+            self._ignore_switch_hits_for(ms=self.config['ignore_switch_ms'])
             self.reset_coil.pulse()
             return True
         # if down. knock down again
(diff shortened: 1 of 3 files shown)


jabdoa2:
Makes sense. Got a test case as well?

cobra18t:
I do not have a test case. There is an ignore_switch_ms test in drop_targets already, but I am not sure how to incorporate the ball search aspect.

jabdoa2:
I guess you are solving some issue with this right? Basically, that the switch activation is delayed after the reset, right? We should reproduce the condition in a test to make sure we do not regress this later on.

cobra18t:
Yes, how do I call the _ball_search method from drop_targets.py in a test?

cobra18t:
OK. I added a test. For each ball search phase, it waits for a drop reset to occur, twiddles the drop reset switch within ignore_switch_ms, and confirms that the state of the target has not changed. The new version of drop_targets.py passes the test and the old version fails.

jabdoa2:
Awesome. Thanks!

PR #1612, opened 2021-10-31
Production bundle args: without MC, with custom machine_path

avanwinkle:
This PR adds two new command-line args for building MPF production bundles.

### `-b` for MPF-only bundles
By calling `mpf build production_bundle -b`, the additional `-b` argument instructs MPF to only package the MPF machine configs in the bundle and not to attempt an MC bundle. This is useful for games that don't use MC for their display, but use something else (e.g. Unity, Godot).

### `--dest-path` for explicit machine paths
By calling `mpf build production_bundle --dest-path="/path/on/target/machine"`, the additional `--dest-path` argument bakes an explicit `machine_path` value into the bundle. This value will overwrite whatever path the machine folder uses in the current machine, which is useful for running production bundles on computers with a different directory structure than the development computer.

While the `machine_path` value is saved to the machine config, the current code does not expose a method for setting the machine path with an explicit value. This PR adds a setter method, to enable the `--dest-path` feature to work.

### DIFF
--- a/mpf/commands/__init__.py
+++ b/mpf/commands/__init__.py
@@ -7,6 +7,7 @@ from pkg_resources import iter_entry_points
 
 
 import mpf.core
+from mpf.core.config_loader import ProductionConfigLoader
 from mpf._version import version
 
 EXAMPLES_FOLDER = 'examples'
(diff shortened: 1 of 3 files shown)


jabdoa2:
I just messaged travis to ask for more OSS credits. Unfortunately, this is currently a manual process. The hope to have an automatic process soon. If not we might want to switch the Linux builds over to my kubernetes cluster with a github actions runner or something similar (hope we do not have to).

jabdoa2 (approved):

PR #1613, opened 2021-11-01
New coil option for timed_enable

avanwinkle:
This PR is an alternative implementation of https://github.com/missionpinball/mpf/pull/1603 that implements a new Coil device method, `timed_enable()`.

Posting this as a draft for feedback on the implementation!


avanwinkle:
The main thing I'm not sure of here is how to branch behavior between the existing pulse and enable methods. With the addition of `default_hold_ms`, would that also apply to the `enable()` method for any coil with it set, or is it exclusively for `timed_enable()` actions? And for devices that use `pulse()` by default, do the automatically use `timed_enable()` if there's a `default_hold_ms` set or do we configure them separately to use `timed_enable()` instead of `pulse()`?

mpf/devices/driver.py
```diff
@@ -318,3 +347,33 @@ def pulse(self, pulse_ms: int = None, pulse_power: float = None, max_wait_ms: in
...
+            pulse_ms: The number of seconds the driver should be enabled at pulse_power
+            pulse_power: The pulse power for the initial pulse. A float between 0.0 and 1.0
+            hold_power: The hold power for the enable duration. A float between 0.0. and 1.0
+            max_wait_ms: Maximum time this pulse may be delayed for PSU optimization
+        """
+        pulse_duration = self.get_and_verify_pulse_ms(pulse_ms)
+        pulse_power = self.get_and_verify_pulse_power(pulse_power)
+        hold_duration = self.get_and_verify_hold_ms(hold_ms)
+        hold_power = self.get_and_verify_hold_power(hold_power)
+        wait_ms = self._notify_psu_and_get_wait_ms(pulse_duration, max_wait_ms)
```
jabdoa2, line 382: wait_ms is ignored here

mpf/core/utility_functions.py
```diff
@@ -335,11 +335,18 @@ def event_config_to_dict(config) -> dict:
         return return_dict
 
     @staticmethod
-    def int_to_hex_string(source_int: int) -> str:
-        """Convert an int from 0-255 to a one-byte (2 chars) hex string, with uppercase characters."""
+    def int_to_hex_string(source_int: int, allow_overflow: bool = False) -> str:
+        """Convert an int from 0-255 to a one-byte (2 chars) hex string, with uppercase characters.
```
jabdoa2, line 339: Maybe we should have two functions? This one seems to be named imprecisely. The doc block still states that it only converts to 2 chars. Maybe "int_to_hex_string" for a generic one. And "one_byte_to_hex_string" (or similar)?

mpf/devices/driver.py
```diff
@@ -318,3 +347,33 @@ def pulse(self, pulse_ms: int = None, pulse_power: float = None, max_wait_ms: in
...
+        """Event handler for timed enable events."""
+        del kwargs
+        self.timed_enable(hold_ms, hold_power, pulse_ms, pulse_power, max_wait_ms)
+
+    def timed_enable(self, hold_ms: int = None, hold_power: float = None, pulse_ms: int = None, pulse_power: float = None, max_wait_ms: int = None):
+        """Pulse and enable this driver.
+
+        Args:
+        ----
+            enable_duration_ms: The number of milliseconds the driver should be
```
jabdoa2: Parameters are inconsistent here. enable_duration_ms is not a parameter.

jabdoa2:
> The main thing I'm not sure of here is how to branch behavior between the existing pulse and enable methods. With the addition of `default_hold_ms`, would that also apply to the `enable()` method for any coil with it set, or is it exclusively for `timed_enable()` actions? And for devices that use `pulse()` by default, do the automatically use `timed_enable()` if there's a `default_hold_ms` set or do we configure them separately to use `timed_enable()` instead of `pulse()`?



That is a good question:
I don't think this should change enable. That would be super confusing and break stuff. However, changing the the way pulse works might be an option.

Pro "automatically use timed_enable instead of pulse":
- Enables this feature without further changes in multiple places

Contra "automatically use timed_enable instead of pulse":
- Would be a bit magic
- Might be confusing if this does not work in hardware rules
- We might have to change some devices anyway to allow overwriting `enable_duration`. Also this might be confusing. Previously, we would only allow overwriting `pulse_duration`. 

Pro "explicit feature":
- Would be opt-in and less confusing
- Would be consistent with hardware rules

Contra "explicit feature":
- Would require changes in device
  - DropTargets
  - BallDevices
  - Diverters
  - Anything else?

Generally, I do favor an explicit call. Maybe we can get some hybrid here. Make this a new explicit call but in case you do not set `default_hold_ms` we could fall back to `pulse` in `timed_enabled`. Would that work?

Where does this generally make sense?
- Drop targets gain quite a bit from this. Might become the default there.
- Some diverters might also improve with this. I think we currently time those enables in software. `timed_enable` could simplify the code
- Ball devices for old Gottlieb troughs would benefit from it. That is a separate ejector already. I don't think this would help on modern troughs.

mpf/devices/driver.py
```diff
@@ -318,3 +347,33 @@ def pulse(self, pulse_ms: int = None, pulse_power: float = None, max_wait_ms: in
...
+            hold_power: The hold power for the enable duration. A float between 0.0. and 1.0
+            max_wait_ms: Maximum time this pulse may be delayed for PSU optimization
+        """
+        pulse_duration = self.get_and_verify_pulse_ms(pulse_ms)
+        pulse_power = self.get_and_verify_pulse_power(pulse_power)
+        hold_duration = self.get_and_verify_hold_ms(hold_ms)
+        hold_power = self.get_and_verify_hold_power(hold_power)
+        wait_ms = self._notify_psu_and_get_wait_ms(pulse_duration, max_wait_ms)
+
+        self.hw_driver.timed_enable(PulseSettings(pulse_power, pulse_duration),
```
jabdoa2, line 386: We probably want a try catch here for two cases:
a) Hardware does not support timed_enable
b) Hardware does not support the duration we need (i.e. only up to 500ms for Spike)

In both cases we need to emulate in software.

avanwinkle:
@jabdoa2 Thanks for the feedback. I've been thinking a lot on this and came up with the following proposition:

1. Add (keep) the new Coil config property, whether it's called `default_hold_ms` or `default_timed_enable_ms`. 
1. Add (keep) the new Coil method `timed_enable(hold_ms)` and its corresponding `event_timed_enable(hold_ms)`, which could be triggered by explicit calls and by `timed_enable_events:` in the config.
1. Add a new Coil config property `pulse_with_timed_enable: false`. If set to `true`, the coil's `self.pulse()` method would immediately call `self.timed_enable()` instead of following the pulse code.
1. The base-level platform Driver can implement `timed_enable()` in software, so every platform automatically gets an implementation. Platform-specific Drivers can implement their own code as suitable. I'd base the implementation on the current Diverter code—and possibly refactor Diverters to leverage the `timed_enable()` instead of what they have now.

All these items are new functionality that don't impact existing flows, so there's no compatibility/breakage for existing users. Without making an explicit change to their code, designers won't be affected—which is good.

For point 3, I like the idea of managing the timed-enable "pulse" at the Coil device level, so that it can be used by Drop Targets, Diverters, et cetera without needing to implement behavior changes on _n_ number of file.

mpf/devices/driver.py
avanwinkle: Should we have a log entry to inform the user that their platform doesn't support timed_enable and we are falling back to software?

mpf/devices/driver.py
avanwinkle: Yes, the interchangeability of `hold` and `enable` throws me off sometimes :)

Should we consider migrating the "hold"-based config values to "enabled"? Like `default_enable_power` and `kwargs["enable_ms"]`? With fallback logic to not break existing games, but a log notice that the "hold" params are going to be deprecated?

jabdoa2:
Makes sense.

> 3\. Add a new Coil config property `pulse_with_timed_enable: false`. If set to `true`, the coil's `self.pulse()` method would immediately call `self.timed_enable()` instead of following the pulse code

What will happen if some device explicitly provides "pulse_ms"? A few devices do that. Will the magic disappear? Or will it just change the pulse part? That feels a bit inconsitent for me. For that reason I would prefer to change call sites to timed_enable which would provide a consistent interface.

avanwinkle:
My expectation would be that calling `pulse(pulse_ms=<some_value>)` on a device with `pulse_with_timed_enable: true` would apply the pulse_ms value to the timed pulse, and the same would apply to a pulse_power value.

What scenarios are you thinking in which a device would be tagged with `pulse_with_timed_enable: true` and its `pulse()` method would be called with an override value? I can think of ball search and eject retries, but I don't think either of those situations would suffer from doing a timed_enable with the larger pulse. Or are you thinking that we should offer increased hold_ms and hold_power for timed_enable during ball search?

(Or if there's a scenario I'm missing, happy to take it into consideration. My mind is limited to drop banks and diverters right now :) )

jabdoa2:
You are probably right. Lets try it and refactor it later in case it goes wrong. Since we do not expose this part I guess that is a valid approach.

avanwinkle:
Updated per the above discussion!

mpf/core/utility_functions.py
avanwinkle: Agreed. The function is limited to just FAST platform devices, but there are a lot of instances of its use. I'm happy to file this for refactoring in a separate PR, to keep the changes clean.

mpf/devices/driver.py
avanwinkle: Fixed to return at the end of the block, as the other methods do

mpf/devices/driver.py
jabdoa2: Yes but we need to be careful not to spam too much as this might happen super often. Maybe at startup or somewhere?

mpf/platforms/pkone/pkone_coil.py
```diff
@@ -149,3 +149,7 @@ def pulse(self, pulse_settings: PulseSettings) -> None:
         cmd = "PCP{}{:02}".format(self.number.board_address_id, self.number.coil_number)
         self.log.debug("Sending Pulse coil command: %s", cmd)
         self.send(cmd)
+
+    def timed_enable(self, pulse_settings: PulseSettings, hold_settings: HoldSettings):
```
jabdoa2, line 153: This could be inside the platform interface. I don't think we have to repeat it everywhere.

jabdoa2:
Left a small comment. Besides that LGTM!

mpf/platforms/pkone/pkone_coil.py
avanwinkle: Thanks, I'll make that change. Wasn't sure what the protocol was, and how much logic an "interface" should have. But it definitely makes it nicer :)

avanwinkle: Ah, that's what I was afraid of. Prospector doesn't like the change:
```
pylint: abstract-method / Method 'timed_enable' is abstract in class 'DriverPlatformInterface' but is not overridden
```

Would it make sense to have the platform interface method be regular and not abstract? That will solve the problem, but speaks to my above question about what is "allowed" in an interface file. Your thoughts appreciated, thanks!

jabdoa2: You can create a new partial class or mixing for platforms which need to emulate this "TimedCoilSoftwareEmulation" or something like that.

avanwinkle:
@jabdoa2  I think a Mixin for the software timed enable makes sense.

I've reverted my last change and restored the `NotImplementedError` on each platform's `timed_enable()` method, to get this PR back to green. I'd like to tackle the software timed_enable in a follow-up PR, and I think that would be the correct place to create the Mixin and replace the per-platform methods with the Mixins—so that this PR doesn't get _too_ huge :)

PR #1614, opened 2021-11-04
Update 14 segment for better lowercase and symbols

cobra18t:
I applied similar changes to the 14 segment font as I did to the 16 segment font previously. It is a handful of lowercase letters and a few symbols.


jabdoa2:
Awesome. Thanks for taking this on!

PR #1616, opened 2021-11-07
feature: service mode for segment displays

jabdoa2:


mpf/modes/service_segment_display/config/service_segment_display.yaml
```diff
@@ -0,0 +1,217 @@
...
+  priority: 100001
+  game_mode: False
+  start_events: reset_complete
+  stop_events: shutdown
+  stop_on_ball_end: False
+
+shows:
+  service_door_open_segment_display:
+    - duration: -1
+      segment_displays:
```
qcapen, line 12: Looks good, although I would have to modify for my segment display machine which has 2 16-segment displays (the top one is 16-segment and the bottom one is 8-segment) and not 4 segment displays. I don't think that's too big of a deal as it would be hard to provide configs for every common possible hardware setup. I'm sure 4 segment displays is a pretty common setup.

jabdoa2: Wow you are quick! The idea is to provide default "slides" for dmd, lcd and segment displays. If you got "special" setups you could copy those and unload ours. If 2 16-digit displays is common you can also ship a set of slides for that as well.

I want to extend this to all our built-in modes such as high score, tilt, match, credits and bonus.

PR #1619, opened 2021-11-24
Auditor checks if events exist before setting to 0

atummons:
The _load_defaults was setting all events that were configured to 0 when
this was called (which happens on every _initiatlize.  This changes it
to be the same logic as switches and players.  It checks if the event is
already in the list.  If it is not, it creates it and sets it to 0.  If
it is in the list, it is ignored.  This also works for the reset because
it blanks the entire list out first, then recreates.  So when it does
this, all values would be missing, and thus are created with a value of
0.

### DIFF
--- a/mpf/core/file_interface.py
+++ b/mpf/core/file_interface.py
@@ -1,7 +1,7 @@
 """Interface for config file loaders."""
 import logging
 import os
-from typing import Union, Tuple, Optional
+from typing import Tuple, Optional
 
 MYPY = False
 if MYPY:    # pragma: no cover
@@ -24,7 +24,7 @@ class FileInterface:
     def find_file(self, filename) -> Tuple[Optional[str], Optional[str]]:
         """Test whether the passed file is valid.
 
-        If the file does not have an externsion, this method will test for files with that base name with
+        If the file does not have an extension, this method will test for files with that base name with
         all the extensions it can read.
 
         Args:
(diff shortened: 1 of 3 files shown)


atummons:
@jabdoa2 The one CI error isn't related to what I touched:

build/lib/mpf/core/file_interface.py
  Line: 4
    pylint: unused-import / Unused Union imported from typing

I can go remove Union from the import, but want to verify first since I didn't touch that file.

jabdoa2:
Awesome! Thanks!

avanwinkle (approved):
Neato!

jabdoa2:
> @jabdoa2 The one CI error isn't related to what I touched:
> 
> build/lib/mpf/core/file_interface.py Line: 4 pylint: unused-import / Unused Union imported from typing
> 
> I can go remove Union from the import, but want to verify first since I didn't touch that file.

Feel free to fix it. Either I broke that by accident or maybe the linter got some update which resulted in new findings.

atummons:
> > @jabdoa2 The one CI error isn't related to what I touched:
> > build/lib/mpf/core/file_interface.py Line: 4 pylint: unused-import / Unused Union imported from typing
> > I can go remove Union from the import, but want to verify first since I didn't touch that file.
> 
> Feel free to fix it. Either I broke that by accident or maybe the linter got some update which resulted in new findings.

All good to go now!

PR #1620, opened 2021-11-28
Allow kwargs or player variable events, default "source" kwarg in variable_player

avanwinkle:
## Player Var-Event Kwargs
This PR proposes the addition of kwargs to player variable changes, and implements them by default on changes triggered by `variable_player:` events.

```
INFO : Event: ==='player_score'=== 
Args={'value': 1500, 'prev_value': 500, 'change': 1000, 'player_num': 1, 'source': 'skillshot'}
```

### Adding Kwargs and New Methods
This proposal adds two new methods to the Player class: 
* `add_with_kwargs(name, value, **kwargs)` will add the `value` to the player's `name` property and include kwargs
* `set_with_kwargs(name, value **kwargs)` will set the player's `name` property to the `value` and include kwargs

The Player class `__setattr__()` method modified to include kwargs, and those are passed to `_send_variable_event()` and included on the posted variable change event.

### Variable Player Includes Context
This proposal extends the Variable Player to automatically include the `context` value of its triggering events as the `"source"` kwarg in posted events.

### Use Cases
This proposal is targeted towards multipliers and jackpots that accrue value based on other modes' scoring. 

For example, a playfield multiplier may want to increase all scoring by a given multiple, but the multiplier adding to the score would itself increase the score, triggering the multiplier again, for infinite scores. By checking the `"source"`, the multiplier mode could listen to _player_score_ events and ignore those whose `source` is itself.

For another example, a jackpot may grow proportional to all points scored in a given mode. By checking the `"source"` kwarg, the jackpot could listen to _player_score_ events and catch those triggered by the mode it uses to grow—while ignoring scoring from all other modes.

Each of these use cases is possible already, but requires additional player variables and workarounds. This proposal just makes it easy to do score (and other player_var) filtering with conditional events and no custom code.

### DIFF
--- a/mpf/config_players/variable_player.py
+++ b/mpf/config_players/variable_player.py
@@ -109,29 +109,29 @@ class VariablePlayer(ConfigPlayer):
         if entry['action'] == "add":
             assert self.machine.game is not None
             assert self.machine.game.player is not None
+            # default to current player
+            player = self.machine.game.player
             if entry['player']:
                 # specific player
                 try:
-                    self.machine.game.player_list[entry['player'] - 1][var] += value
+                    player = self.machine.game.player_list[entry['player'] - 1]
                 except IndexError:
                     self.warning_log("Failed to set player var %s for player %s. There are only %s players.",
                                      var, entry['player'] - 1, self.machine.game.num_players)
-            else:
-                # default to current player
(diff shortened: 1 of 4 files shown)


jabdoa2:
A test would be nice. Good idea. Thanks for addressing this!

PR #1621, opened 2021-12-11
Cumulative drop_target reset delays for multi-coil banks

avanwinkle:
This PR continues to refine the DropTarget behavior for target banks with individual reset coils.

**The Setup**
The existing MPF behavior is smart enough to wait after resetting a drop target before checking the switch value and updating the target state, using the `ignore_switch_ms` property. MPF is also smart enough to stagger a series of drop target resets in order to smooth out PSU usage, using the `reset_coil_max_wait_ms` property. 

**The Problem**
When using staggered drop target resets for power optimization, MPF does not factor into account the additional time it takes to reset all of the drop targets. As a result, the `ignore_window_ms` may expire while the coils are still resetting, leading to unexpected events and behavior. Additionally, the firing of a reset coil may momentarily disturb the opto sensor on a neighboring target, causing false switch detections.

**The Solution**
This PR tracks how much additional time is being requested by the PSU to sequentially pulse all of the drop target coils, and _adds_ that time to the `ignore_switch_ms` property. This ensures that all of the coils have fired and switches have settled before MPF reads the state for updating the bank (or attempting a retry).

_Why *add* and not *max*?_
If the total delayed time is still less than the `ignore_switch_ms`, or vice versa, why not just use the larger value? My assumption is that the `ignore_switch_ms` time is provided by the designer based on the settling time of the switches _after_ all coils are done firing. Adding the values ensures that the coils are done *and* the settling is complete.

_This PR also adds `reconcile=True` to the `init_phase_4` reset callback to prevent the playfield from being marked active during startup._

### DIFF
--- a/mpf/devices/drop_target.py
+++ b/mpf/devices/drop_target.py
@@ -52,7 +52,7 @@ class DropTarget(SystemWideDevice):
 
         # can't read the switch until the switch controller is set up
         self.machine.events.add_handler('init_phase_4',
-                                        self._update_state_from_switch, priority=2)
+                                        self._update_state_from_switch, priority=2, reconcile=True)
         self.machine.events.add_handler('init_phase_4',
                                         self._register_switch_handlers, priority=1)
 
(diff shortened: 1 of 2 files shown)


jabdoa2:
Awesome. Thanks for digging into this!

PR #1622, opened 2021-12-29
Bugfix orphaned handlers on shot groups

avanwinkle:
This PR fixes a bug created in https://github.com/missionpinball/mpf/commit/58d3df636f6b1027d68024ba196e3cb852ca59a9 that causes shot groups to continually add event handlers.

The change in https://github.com/missionpinball/mpf/commit/58d3df636f6b1027d68024ba196e3cb852ca59a9 introduces new handlers for the ShotGroup's `self._check_for_complete` when the device is added to a mode, but these handlers are never removed. The next time the mode starts, they are all added again.

This PR removes these handlers in the ShotGroup's `self.device_removed_from_mode()` method, and adds a test to prevent regression.

_Also, this PR adds a catch during shutdown. During a catastrophic failure, e.g. a syntax error in an MPF file that prevents startup, `self.events` is undefined and that subsequent error blocks the stacktrace of the underlying syntax error. This catch allows the machine to gracefully shutdown and display a useful stacktrace._

### DIFF
--- a/mpf/core/machine.py
+++ b/mpf/core/machine.py
@@ -767,7 +767,8 @@ class MachineController(LogMixin):
         except RuntimeError:
             print("Failed to stop all tasks")
 
-        self.events.stop()
+        if hasattr(self, "events"):
+            self.events.stop()
         self.clock.loop.stop()
         self.clock.loop.run_forever()
         self.clock.loop.close()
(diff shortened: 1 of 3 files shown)


jabdoa2:
Thanks! Good one.

PR #1623, opened 2022-01-01
add OPP light subtype "incand" to VirtualHardware platform

tp-ops:
add OPP light subtype "incand" to VirtualHardware platform

Expand if statement with "incand" for using with OPP lights with subtype "incand"


jabdoa2:
Good one. Thanks!

PR #1625, opened 2022-01-03
Fix IMC Load Error

atummons:
IMC would not load because machine did not have the attribute
is_shutting_down.  This checks for the attirbute and its value of true
in one if statement now.

### DIFF
--- a/mpf/config_players/plugin_player.py
+++ b/mpf/config_players/plugin_player.py
@@ -3,6 +3,7 @@ from mpf.config_players.device_config_player import DeviceConfigPlayer
 
 
 class PluginPlayer(DeviceConfigPlayer):
+
     """Base class for a remote ConfigPlayer that is registered as a plug-in to MPF.
 
     This class is created on the MPF side of things.
--- a/mpf/config_players/plugin_player.py
+++ b/mpf/config_players/plugin_player.py
@@ -3,7 +3,6 @@ from mpf.config_players.device_config_player import DeviceConfigPlayer
 
 
 class PluginPlayer(DeviceConfigPlayer):
-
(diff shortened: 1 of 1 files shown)


jabdoa2:
Thanks! We really need a unit test for this. This broke quite a few times over the last years.

atummons:
That was my next question.  I will write one up.  How do I do one for IMC?  Is that in the MC code?  Any specific things I need to plug into to get IMC to run, versus just the full MC in tests?

jabdoa2:
Yeah that would be in mc. Mocking all the qt parts might be a bit tricky.

mpf/config_players/plugin_player.py
```diff
@@ -52,7 +52,8 @@ def register_player_events(self, config, mode=None, priority=0):
         events = super().register_player_events(config, mode, priority)
         # when bcp is disabled do not register plugin_player
         if self.machine.options['bcp']:
-            if self.machine.is_shutting_down:
+            # hasattr check is for IMC to work
+            if hasattr(self.machine, 'is_shutting_down') and self.machine.is_shutting_down:
```
avanwinkle, line 56: I know this has already merged, but you can combine the error-safe `hasattr` the truthiness of the value by passing a default value to `getattr`.

```
if getattr(self.machine, 'is_shutting_down', False):
  return events
```

Saves you one boolean evaluation. Hurray!

atummons: Anthony, thanks!  I am learning Python here to help out, so I just learned about hasattr, but getattr certainly seems like a better fit.  Didn't even know that was a think before now!  I have a new PR I will push to get this adjusted while it is top of mind.

PR #1627, opened 2022-01-07
Simplifies attr check related to IMC

atummons:
Users getattr instead of hasattr to remove one boolean.  Sets a default value to false if the attribute doesn't exist.


avanwinkle (approved):
Cool!

PR #1632, opened 2022-02-16
Text_UI: Updates player variables instantly

atummons:
Change adds handlers for all of the player_vars to ensure that they are updated in the UI at the same time as the value changes.  This handles player, ball and score in addition to all other player_vars.


jabdoa2:
Awesome. Thanks!

PR #1633, opened 2022-03-14
Adds a feature to jump to a state of shot profile

atummons:
This feature lets a user have a list of events that will jump a shot to a
specific state of its shot profile when fired.  This can be done even
when the shot is disabled based on the arguments provided.

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -1445,7 +1445,13 @@ shots:
     restart_events: event_handler|event_handler:ms|None
     advance_events: event_handler|event_handler:ms|None
     hit_events: event_handler|event_handler:ms|None
+    jump_states: list|subconfig(shot_jump_states)|None
     show_tokens: dict|str:template_str|None
+shot_jump_states:
+    events: list|event_handler|
+    state: single|int|0
+    force: single|bool|true
+    force_show: single|bool|false
 shot_groups:
     __valid_in__: mode
     __type__: device
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -1449,7 +1449,7 @@ shots:
     show_tokens: dict|str:template_str|None
 shot_control_events:
     events: list|event_handler|
-    state: single|int|0
+    state: single|int|
     force: single|bool|true
     force_show: single|bool|false
 shot_groups:
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -1445,9 +1445,9 @@ shots:
     restart_events: event_handler|event_handler:ms|None
     advance_events: event_handler|event_handler:ms|None
     hit_events: event_handler|event_handler:ms|None
-    jump_states: list|subconfig(shot_jump_states)|None
+    control_events: list|subconfig(shot_control_events)|None
     show_tokens: dict|str:template_str|None
-shot_jump_states:
+shot_control_events:
     events: list|event_handler|
     state: single|int|0
     force: single|bool|true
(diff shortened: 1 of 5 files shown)


mpf/devices/shot.py
```diff
@@ -412,6 +414,23 @@ def _delay_switch_hit(self, switch_name, **kwargs):
...
         self.active_delays.remove(switch)
 
+    def _register_jump_state_handlers(self):
+        for jump_state in self.config['jump_states']:
+            for event in jump_state['events']:
+                self._handlers.append(self.machine.events.add_handler(event, self._jump_states,
+                                                                      jump_state_config=jump_state))
+
+    def _remove_jump_state_handlers(self):
+        self.machine.events.remove_handlers_by_keys(self._handlers)
```
avanwinkle: The `_remove_switch_handlers()` method performs this same function and references the same `self._handlers` value, so this cleanup method will have no effect.

mpf/devices/shot.py
```diff
@@ -412,6 +414,23 @@ def _delay_switch_hit(self, switch_name, **kwargs):
...
+
+    @event_handler(7)
+    def _jump_states(self, jump_state_config, **kwargs):
+        """Takes in a jump_state event to move the state of the given shot to a specific state. (Default = State 0)"""
```
avanwinkle: You note the default state is 0, yet there are no defaults in this handler. I see that the config_spec sets 0 as the default value but I feel like the method documentation shouldn't make assumptions about other areas of the code. Would you update the handler, or update the comment?

mpf/devices/shot.py
```diff
@@ -266,6 +267,7 @@ def device_removed_from_mode(self, mode):
         """
         super().device_removed_from_mode(mode)
         self._remove_switch_handlers()
+        self._remove_jump_state_handlers()
```
avanwinkle: See note below on this method

mpf/config_spec.yaml
```diff
@@ -1445,7 +1445,13 @@ shots:
     restart_events: event_handler|event_handler:ms|None
     advance_events: event_handler|event_handler:ms|None
     hit_events: event_handler|event_handler:ms|None
+    jump_states: list|subconfig(shot_jump_states)|None
```
avanwinkle: This structure seems very similar to `control_events` in other device types. What are your thoughts on calling it the same, for consistency? Even though it's strictly jump-related?

avanwinkle (commented):
Neat stuff!

mpf/devices/shot.py
atummons: Removed in latest commit.  Thanks for the review!

mpf/devices/shot.py
atummons: Agreed and good catch.  I cleaned up to not state a default, since there truly isn't.  I guess there should be error handling on the processing of it?  If the state doesn't have a value, then throw an error that its missing a required field.  If this seems like a good idea, can you point me somewhere to use as a reference in how to message that, or require during config validation.

mpf/config_spec.yaml
atummons: Renamed to control_events.  Not a perfect match to the pattern elsewhere, but makes a lot more sense.

mpf/devices/shot.py
atummons: That was horribly worded.  To make sense of it.  It gives the general config error if I remove 0 as a default value.  Do we think it would be better to give you an error to make it clear that you should be telling it what state?  Or do we think it would be better to plug a value, even if its wrong?

My vote would be to remove the config_spec of 0 so that a user gets an error that they are missing a required argument per the standard cmd warning.  Obviously once this is approved, I'll update the docs to include that events and state are the two required fields.

atummons: I have updated to remove the default value.  This will be flagged as a required field.  Once approved, I will write the Help Documents that list event and state as required fields with no defaults.

avanwinkle (approved):
Very cool, thanks for the new functionality!

jabdoa2:
👍

PR #1634, opened 2022-04-12
Syncs up chimes with reel coil pulse

atummons:
Adds conssitency for the will_xx _xxing and _xxed format to the
advancement of the score reel.  Changed the event handler for chime
pulses to be the _advancing event, which triggers at the same time as
the coil advance pulse.

The only potential break is if someone is queueing custom events off the
reel_(name)_advance event as this is now _advanceD (past tense).

### DIFF
--- a/mpf/devices/score_reel_group.py
+++ b/mpf/devices/score_reel_group.py
@@ -67,7 +67,7 @@ class ScoreReelGroup(SystemWideDevice):
             if self.config['chimes'][i]:
                 if not self.reels[i]:
                     self.raise_config_error("Invalid reel for chime {}".format(self.config['chimes'][i]), 1)
-                self.machine.events.add_handler(event='reel_' + self.reels[i].name + '_advance',
+                self.machine.events.add_handler(event='reel_' + self.reels[i].name + '_advancing',
                                                 handler=self.chime,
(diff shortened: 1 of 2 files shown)


mpf/devices/score_reel_group.py
```diff
@@ -67,7 +67,7 @@ async def _initialize(self):
             if self.config['chimes'][i]:
                 if not self.reels[i]:
                     self.raise_config_error("Invalid reel for chime {}".format(self.config['chimes'][i]), 1)
-                self.machine.events.add_handler(event='reel_' + self.reels[i].name + '_advance',
+                self.machine.events.add_handler(event='reel_' + self.reels[i].name + '_advancing',
```
jabdoa2, line 70: We could just keep the old event here as well. That way we keep backward compatibility. Guess we should mark it as deprecated and remove it in some future major version.

atummons: In the other file like 168, I changed it from advance to advanced.  And if I put both of them there, it will double fire, once for event handler of advancing and again for advance/d.  But you might have a magic way to tell it to only do one or the other.

Maybe I am missing what you are saying, but the event is code generated, we are just moving it to a different code posted event.

PR #1635, opened 2022-04-12
Fixes issue with no player_vars in config

atummons:
This fixes an error when there are no player_vars defined in the main config file.  It causes an error on startup.  This goes through the advanced routine  if player_vars exist, but reverts to the simplified routine of just adding the three main handlers if none are identified.

### DIFF
--- a/mpf/core/text_ui.py
+++ b/mpf/core/text_ui.py
@@ -144,8 +144,13 @@ class TextUi(MpfController):
             self.machine.events.add_handler("mode_{}_started".format(mode.name), self._mode_change)
             self.machine.events.add_handler("mode_{}_stopped".format(mode.name), self._mode_change)
 
-        for player_var in self.machine.config['player_vars']:
-            self.machine.events.add_handler("player_{}".format(player_var), self._update_player)
+        if 'player_vars' in self.machine.config:
+            for player_var in self.machine.config['player_vars']:
+                self.machine.events.add_handler("player_{}".format(player_var), self._update_player)
+        else:
+            self.machine.events.add_handler('player_number', self._update_player)
+            self.machine.events.add_handler('player_ball', self._update_player)
+            self.machine.events.add_handler('player_score', self._update_player)
 
         self.machine.switch_controller.add_monitor(self._update_switches)
         self.machine.register_monitor("machine_vars", self._update_machine_vars)
--- a/mpf/core/text_ui.py
+++ b/mpf/core/text_ui.py
@@ -117,6 +117,9 @@ class TextUi(MpfController):
                                         self._bcp_connected)
         self.machine.events.add_handler('shutdown', self.stop)
         self.machine.add_crash_handler(self.stop)
+        self.machine.events.add_handler('player_number', self._update_player)
+        self.machine.events.add_handler('player_ball', self._update_player)
+        self.machine.events.add_handler('player_score', self._update_player)
         self.machine.events.add_handler('ball_ended',
                                         self._update_player)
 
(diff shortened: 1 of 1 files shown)


jabdoa2:
Could we also add this to mpfconfig.yaml to achieve the same?

atummons:
I looked at that, but it required a value or it yelled. Couldn't just be empty player_vars:. And if I did score or something already used that broke other things. And I didn't want to generically create a test var for everyone.

But you might know a setting I'm missing to make that work.

mpf/core/text_ui.py
```diff
@@ -144,8 +144,13 @@ def _init(self, **kwargs):
             self.machine.events.add_handler("mode_{}_started".format(mode.name), self._mode_change)
             self.machine.events.add_handler("mode_{}_stopped".format(mode.name), self._mode_change)
 
-        for player_var in self.machine.config['player_vars']:
-            self.machine.events.add_handler("player_{}".format(player_var), self._update_player)
+        if 'player_vars' in self.machine.config:
+            for player_var in self.machine.config['player_vars']:
+                self.machine.events.add_handler("player_{}".format(player_var), self._update_player)
+        else:
```
jabdoa2: Why is this else? Those three will always be displayed. I am not sure if the config reference above is correct at all. There is a separate setting to configure what text ui displays. I guess that should be used here.

atummons: Not sure exactly what you are getting at with the setting to configure the text UI.  If you look at the history of this, there was an issue where player_vars didn't update until a score update occurred.  So the text_ui would have a delay on updating player_vars.  I added the for loop here, which then immediately updated it in PR #1632 

What I then noticed is that it required player_vars to be part of it.  So if the player_vars are present in config then it will register all of them and you are off to the races.  If you don't have them, the else will at least set the main three so that text_ui can update.  Then when one of those hits, it will update all player_vars, whether it has a handler or not.

So if you have a way to trigger _update_player whenever a player_var updates without this loop, let me know.

jabdoa2: I do not fully understand the rational here. The old logic looked fine to me for score, ball and player number.

Player vars is a config setting to configure initial values for player vars. How does that relate to text ui? I think we got a setting somewhere to change which player vars are displayed. If that is configured I would expect it to be used here. Otherwise we should subscribe all player vars instead (as they could all be shown). Theoretically, this could also happen when we render text ui as we know which ones are displayed (obviously we would only do this once they change which should be super rare). I guess this would even speed up text ui as we would update less often.

atummons: The link is that the only subscriptions to call the update method for the text_ui for players was the four explicitly stated: ball_ended, player_number, player_ball, and player_score.  If a random player_var changes value, e.g., aliens, it will not update on the text_ui until one of those four listed above occurs.  So if it updated aliens, but not score, it won't updated on the text_ui.

The previous PR that I put through meant that if you describe it in config as a player_var, then it will have a handler to automatically update.  Somehow, and I honestly don't know, through that magic, it makes score, player, ball all update as well.  I had them and realized I could remove them from the for loop, and they were still registered and updated correctly.  If you don't describe it in the player_vars section, then the text_ui will not update it unless one of the four mentioned above are updated, and then all update.

The best way would be to add a handler to determine did a player_XX get posted, and then that calls the _player_update method.  If there is a way to do that, then we should rip out this and my last PR on the topic and replace it.  Then we don't need to explicitly register score, ball, etc. but can just say if player_{anything} is posted in the events run _player_update.  Or as you describe at the end, create a list of what is actively being displayed and register the handlers for only what is shown.

atummons: @jabdoa2 saw a help come through related to this. Do you have thoughts on this? If not I'll change the commit back to previous so it doesn't break. 

Doing so will mean that player_vars in text_ui won't update until score change or ball end. But text_ui being slow is better than it being broken.

jabdoa2: I kind of disagree. Especially on Windows or via SSH this makes MPF almost unusable. Your change will break number, score and ball (unless you add them to `player_vars`). At least the last thing needs fixing. Also `player_vars` != what is displayed. I am not sure if this really fixes the issue.

atummons: I will just update this pr to revert to what it was prior to the last PR on this tomorrow. That means score, player, and ball have explicit handlers and will be the only events that redraw player_vars in text UI (plus ball_end).

This code, and as I said, I don't know how updates score, player, and ball without an explicit handler, but it does as that is what's in dev branch right now. But it requires player_vars section in config to make that magic happen.

Ultimately this needs a way to add a wildcard handler for player_* to register all to call the update. But I have tried and can't get a dynamic runtime replacement. Or some alternative way to register. Otherwise player_vars won't update unless those three are called.

PR #1636, opened 2022-04-23
Correct delayed kick timing for OPP

cobra18t:
The commands for coil_pulse_delay are off by a factor of 2 in the OPP implementation. Also, OPP can only do a coil_pulse_delay up to 30ms but the code currently only checks to make sure the setting is <=255. 

This is a work in progress. Stand by.


cobra18t:
Code updated, unit test updated and checked, and verified on hardware.

I think this is good to go unless you or the automatic checks say otherwise.

jabdoa2:
Thanks!

PR #1638, opened 2022-05-23
Fast segment display update hz

toomanybrians:
Updates the FAST Platform interface so that segment display updates are sent out at a configurable hz rate. Also updates some deprecated modules.

### DIFF
--- a/mpf/platforms/fast/fast.py
+++ b/mpf/platforms/fast/fast.py
@@ -7,7 +7,7 @@ as FAST I/O boards.
 import asyncio
 import os
 from copy import deepcopy
-from distutils.version import StrictVersion
+from packaging import version
 from typing import Dict, Set, Optional
 from serial import SerialException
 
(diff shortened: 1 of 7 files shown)


mpf/platforms/fast/fast.py
```diff
@@ -244,8 +250,10 @@ def register_io_board(self, board):
 
     def _update_watchdog(self):
         """Send Watchdog command."""
-        if self.net_connection:
+        try:
             self.net_connection.send('WD:' + str(hex(self.config['watchdog']))[2:])
+        except:
```
jabdoa2, line 255: Why do we ignore all errors here? I am certain that this breaks the linter (which is not currently running in CI).

mpf/platforms/fast/fast.py
```diff
@@ -268,7 +276,7 @@ def process_received_message(self, msg: str, remote_processor: str):
             self.log.warning("Received malformed message: %s from %s", msg, remote_processor)
             return
 
-        # Can't use try since it swallows too many errors for now
+        # Can't use try since it swallows too many errors for now #TODO
```
jabdoa2, line 279: Double comment

mpf/platforms/fast/fast.py
```diff
@@ -335,22 +343,48 @@ def register_processor_connection(self, name: str, communicator):
...
+    def _start_seg_updates(self, **kwargs):        
+
+        for s in self.machine.device_manager.collections["segment_displays"]:
+            self.fast_segs.append(s.hw_display)
+        
+        self.fast_segs.sort(key=lambda x: x.number)
+
+        if self.fast_segs:
+            self._seg_task = self.machine.clock.schedule_interval(self._update_segs,
+                                                1 / self.machine.config['fast'][
```
jabdoa2, line 367: Invalid indent. Please run a linter.

PR #1640, opened 2022-06-12
Timer bugfix: set current player before updating tick values

avanwinkle:
### SUMMARY

This PR fixes a bug in the _Timer_ class related to multiplayer games and persisted tick values. This bug would trigger variable updates and events using the wrong player number, which could create unexpected downstream effects.

### DETAILS
When added to a mode, a _Timer_ instance sets `self.player = player` to track the current player—which it uses for saving the tick values as player variables, to be consumed by many others.

There is a bug where the `self.player` value is set at the *end* of the mode setup logic, *after* the various timer values and ticks have been set. In a multiplayer game, this means that the values are being written to the _previous_ player, instead of the current one.

This PR fixes the bug by setting `self.player = player` at the top of the mode setup logic, so that all subsequent calls will use the correct player.

tick tock

### DIFF
--- a/mpf/devices/timer.py
+++ b/mpf/devices/timer.py
@@ -94,6 +94,7 @@ class Timer(ModeDevice):
     def device_loaded_in_mode(self, mode: Mode, player: Player):
         """Set up control events when mode is loaded."""
         del mode
+        self.player = player
         self.tick_secs = self.config['tick_interval'].evaluate([])
 
         try:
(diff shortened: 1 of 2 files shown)


jabdoa2:
Thanks for debugging and fixing this! Do you got a testcase or example which we can use to regression test this?

avanwinkle:
Now with more test!

jabdoa2:
Awesome. Thanks!

PR #1643, opened 2022-06-30
Adds speed option to achievement shows

atummons:
Allows the user to set a speed parameter for shows that are played for
any of the achievement stages.


jabdoa2:
Simple but effective. Thanks!

jabdoa2:
Simple but effective. Thanks!

PR #1644, opened 2022-07-10
OPP Incand fix

cobra18t:
Add the command to stop OPP incandescent lights from blinking when MPF starts.


jabdoa2:
No longer needed?

cobra18t:
This is still needed. I should not have closed it and just left it as a draft. It will be ready soon.

PR #1647, opened 2022-08-11
Bugfix: FAST timed_enable hold ms/pwm inverted

avanwinkle:
This PR fixes a bug for `timed_enable` pulses on the FAST platform.

The original implementation incorrectly posted the hold PWM argument before the hold duration, but according to the FAST Protocol documentation, the duration should come _before_ the PWM.

```
[DL/DN]:<DRIVER_ID>,<TRIGGER>,<SWITCH_ID>,<10>,<PWM1_ONTIME>,<PWM1>,<PWM2_ONTIME>,<PWM2>,<REST_TIME><CR>
```

I don't know why I inverted the values initially, but now they'll be correct.

I was inverted

### DIFF
--- a/mpf/platforms/fast/fast_driver.py
+++ b/mpf/platforms/fast/fast_driver.py
@@ -206,8 +206,8 @@ class FASTDriver(DriverPlatformInterface):
                 self.number,
                 hex_ms_string,
                 self.get_pwm_for_cmd(pulse_settings.power),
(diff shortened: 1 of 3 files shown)


toomanybrians (approved):
looks good, thanks! I'll do another dev version bump in a few days when I merge in the new installer stuff and then we'll get that on on pypi.

PR #1648, opened 2022-08-12
Add "use_dots_for_commas" setting to segment displays

cobra18t:
The segment mappings assume you have a dot (dp). The dot on my displays makes a comma by default since commas are more relevant to pinball for things like thousands separators. I needed a way to convert commas into dots so they would display correctly. Some people may find this useful who have dots only on there segment displays, but want to take advantage of having thousands separators...even if they are dots

IMG_20220811_225600618
.


jabdoa2:
Awesome. Thanks for also adding a test!

PR #1649, opened 2022-08-19
Add support for NeoSeg display configuration

cobra18t:
NeoSeg displays are light_segment_displays but they were a pain to configure manually because so many lights needed to be generated and linked to the display. This PR creates a neoseg_display light_group and uses that to generate a correctly ordered list of lights. Also light_groups were added as a config option to light_segment_displays, so you could list the neoseg_displays to be used as one display. MPF Docs updates must surely follow.


jabdoa2:
Sweet

PR #1673, opened 2022-12-09
Update text_ui.py

atummons:
Reverts back to where it used to be.  This change throws an error (maybe an issue upstream) where various events players will error out.  This makes Text_UI have the issue from before where it won't update player variables unless score, ball, or player changes.  But better than breaking.  At some point I will dig into a real fix to make Text_UI match what is expected for all scenarios.

### DIFF
--- a/mpf/core/text_ui.py
+++ b/mpf/core/text_ui.py
@@ -327,7 +327,6 @@ class TextUi(MpfController):
 
         names = self.config.get('player_vars', player_vars.keys())
         for name in names:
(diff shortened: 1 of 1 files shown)


jabdoa2:
I guess a try catch inside the handler would work around all those issues with non trivial types.

atummons:
> I guess a try catch inside the handler would work around all those issues with non trivial types.

Good idea.  Put it in, and is working for my local tests.

jabdoa2:
Awesome. Thanks :-)

Issue #693, opened 2016-11-22
Service Mode Menu


jabdoa2:
Extend menu structure (extendable by overloading mode):
* ~~Settings (extendable by adding items to settings section in config)~~
  * ~~Lights~~
    * ~~Brightness~~
  * ~~Game~~
    * ~~Flipper strength~~
* Tests (registered in code; extendable in scriptlets)
  * ~~Coils~~
  * ~~Switches~~
  * ~~Lights (if there are matrix_lights)~~
  * ~~LEDs (if there are LEDs)~~
  * ~~GI (if there is GI)~~
  * Motors (if there are any; add test points to motor?)
  * Servos (if there are any; add test points to servos?)
  * Ball Device Test
  * Diverter (if there are any)
  * Magnets (if there are any)
  * Autofire Test (if there are any)
  * Kickback Test (if there are any)
  * Coil Cycle
  * Lights Cycle
  * LED Cycle
  * Trough Cycle (if there is a trough)
  * Active switches list/matrix
  * Sound/Musik/SFX
* ~~Adjustments~~
  * ~~Standard~~
  * ~~Game~~
  * ~~Coins~~
  * ~~Feature~~
* ~~Audits~~
  * ~~Earnings~~
  * ~~Switch~~
  * ~~Shot~~
  * ~~Event~~
  * ~~Player~~
* Pricing
  * Credits per game
  * Match Percentage
* Volume
  * Master Volume
  * Track X Volume (on per track)

jabdoa2:
Any of those points would be a good first issue if somebody is interested. You can have a look at existing devices (such as lights) in the service mode. Code for other devices (such as servos) should be very similar. Help would be very welcome!

PR #1694, opened 2023-04-24
Extend Service Menu

GraceHenke:
Extends the abilities of the service menu. Abilites implemented: change credits per game, match percentage, toggle between free play and credit mode, and lastly change the number of max credits. Also implemented some early testing of these abilities. Works on Issue #693

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -31,6 +31,7 @@ their spare time, unpaid, for the love of pinball!
  * James Cardona
  * Dave Ensminger <dave@ensadi.com>
  * Charles Duncan (nullbuilds)
+ * Grace Henke
 
 MPF is inspired by pyprocgame and skeletongame which were written by:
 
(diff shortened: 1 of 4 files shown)


cobra18t:
I thought the Free Play toggle and match percentage selection already existed in the service menu. I have used the Free Play toggle in the past and I have at least seen the match percentage as one of the options.

avanwinkle:
Yes, `match_percentage` exists but only appears in the Settings menu if match mode is enabled in the game (i.e. `match` is listed in the `modes:` section of the main config). Similarly, `free_play` will be added to the Settings menu if the `credits` mode is enabled.

Editing the `max_credits` is not currently available (hard-coded in the configs), which could be a nice addition if that's something people use (I'm not an operator, I don't know much about these things). Pricing tiers are also hard-coded to a preset list of options, which could be made dynamic as well if there's a use case.

My recommendation would be to isolate these changes to either their own mode or the credits mode (perhaps `max_credits` goes in credits mode and pricing tiers goes in its own mode) so that they only take effect for users who specifically want them, rather than being present in the global service menu. See _credits.py_ or _match.py_ for examples, and thanks for contributing!

PR #1695, opened 2023-05-17
feat: Floor division support in dynamic values

avanwinkle:
This PR adds floor division functionality to dynamic values, useful for rounding arbitrary numbers to clean multiples. 

For example, calculating a point score based on a timer count (hurryup) or a multiplier can yield tens and ones of digits, or even floats. Many pinball games like to have a minimum score multiple of 10 (e.g. no single-digit points) or even 100, and this PR enables a dynamic value to crop to the nearest iteration using the standard syntax of floor division * divisor.
```
variable_player:
  jackpot_shot_hit:
    score: current_player.jackpot_value * current_player.float_multiplier // 100 * 100
```

The placeholder templating engine uses native python operations so it inherently understands that `//` means floor division, but the code module wasn't mapping the `op.floordiv` method for executing it. This PR simply maps the method, and it works. Tests added to verify and prevent regressions.

division is great

### DIFF
--- a/mpf/core/placeholder_manager.py
+++ b/mpf/core/placeholder_manager.py
@@ -20,7 +20,7 @@ if MYPY:   # pragma: no cover
 
 
 # supported operators
-OPERATORS = {ast.Add: op.add, ast.Sub: op.sub, ast.Mult: op.mul,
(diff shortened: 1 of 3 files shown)


toomanybrians:
Awesome feature, and such a simple and powerful change to make! Thanks for the tests, and I'll add this to the docs.

PR #1696, opened 2023-05-19
Add Player Vars to High Score Mode

atummons:
This updates the high_score.py file to pass through both the player_num and the category of the award.  This is pushed through a new unique event (to prevent breaking existing games). The high_score.yaml file is updated as well to show how to use these new features to create a new specific slide for the awards where player and machine variables should be shown.


atummons:
I don't know if we care about those two fails.  I didn't touch that file, and it's odd it failed both Windows and Mac for 3.8 but passed the rest.  I will leave that up to someone wiser than myself if it needs addressed, and how to address it.

toomanybrians:
Nice!

PR #1698, opened 2023-05-29
added documentation to have a fully working (simple) code example to …

worldpeace-germany:
…use events in code

I have to admit I have no idea how to generate the documentation locally to see if everything is well formatted. But since the change should be rather simple I believe that everything should be okay.


toomanybrians (approved):
Thanks! I'm going to pull these docstrings into the new docs website as well.

PR #1699, opened 2023-06-14
Save High Score Variables in high_scores.yaml

atummons:
This extends the high score mode to allow the user to define vars that they want to be saved along with the initials and title of the high score that is achieved.  The user can define any number of player or machine vars that should be saved along with the files.  These are written into the high_scores.yaml file as a third tuple.  The code parses the list and for those with two elements (no vars) or three elements (with vars) it will create the machine variables so that it can be used on attract mode, etc.  Then it will save to the file and update the machine vars after the game ends.

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -725,6 +725,7 @@ high_score:
     enter_initials_timeout: single|secs|20s
     reverse_sort: list|str|None
     reset_high_scores_events: list|event_handler|high_scores_reset,factory_reset
+    vars: dict|str:list|None
 info_lights:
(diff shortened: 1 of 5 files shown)


toomanybrians:
This is great, thanks!

I want to get this updated in the docs as well. I created a tracking issue for the docs so I won't forget about it. It would be cool if you wanted to take a stab at the docs too? I can do it this week otherwise.

https://github.com/missionpinball/mpf-docs/issues/463

atummons:
@toomanybrians I'll get the docs updated this week. Just wanted to make sure it was all good first.

PR #1700, opened 2023-06-22
add support for using * tag to return all devices

ericselk2018:
My use case is to turn off all lights when a skill shot is hit (like Deadpool).  I didn't want to tag every light with a tag like "all" or "*" to accomplish this, but if there is a better way or this feature seems problematic, no worries if this PR is declined.

I didn't see any existing tests on this area -- if there are any, I'm happy to add a new one for this change.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -31,6 +31,7 @@ their spare time, unpaid, for the love of pinball!
  * James Cardona
  * Dave Ensminger <dave@ensadi.com>
  * Charles Duncan (nullbuilds)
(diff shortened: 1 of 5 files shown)


toomanybrians:
This looks great, thanks!

I just checked the tests, looks like there’s a test for tags you can jump onto here:
https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_DeviceCollection.py#L29

ericselk2018:
Test updated.

PR #1708, opened 2023-08-29
Bugfix settings machine vars not persisting on subsequent data writes

avanwinkle:
This PR fixes a bug that would occasionally (okay, _often_) forget saved machine vars and cause operator-configured settings to reset to default. After much digging and consternation, I believe I've figured out the cause and it's a simple fix.

### The Background
When MPF initializes, it first reads all machine variables from the _machine_vars.yaml_ file and then appends those with the (default) variables from the `machine_variables:` config section. All settings values are backed under-the-hood by machine variables, so this logic covers operator-adjustable settings as well.

### The Bug
While the code flow for loading machine vars from config section includes the configuration of a `persist` value, loading machine vars from the data file does _not_, which means all saved settings are defaulted to `persist: False`. Which is okay for a short while, but eventually something else is going to write an update to the persisted data (a new setting adjustment, or a player score, or something).

### The Impact
The way the logic of writing the data works is that the code looks through all the machine vars, finds all of the ones that are `persist: True`, and writes a complete new data file with those values. Remember from before that values previously loaded from the data file are `persist: False`, so all of those values get wiped.

### The Fix
This PR tweaks the flow for loading persisted values from a data file, explicitly flagging them each as `persist: True` in the machine variables controller. After all, if the value was persisted before, it should be persisted still, right?

The updated persistence logic takes place _after_ data validation and expiration checks, so it shouldn't impact data retention policies, and it fixes the issue of saved settings being forgotten on subsequent data writes.

machine vars be forgettin

### DIFF
--- a/mpf/core/machine_vars.py
+++ b/mpf/core/machine_vars.py
@@ -45,7 +45,8 @@ class MachineVariables(LogMixin):
 
                 continue
 
-            self.set_machine_var(name=name, value=settings['value'])
+            # Any value that was persisted before should be persisted again
+            self.set_machine_var(name=name, value=settings['value'], persist=True)
 
         self._load_initial_machine_vars()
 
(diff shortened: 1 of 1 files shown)


atummons:
@avanwinkle Seeing that tests fail now.  I looked at it a little bit, and it is probably just mock data doesn't matter in real life.  But values that are expected to not persist are being set to true on the load with this change.  I didn't want to just go fudge the data to pass the tests in case there is an edge case that is not being caught.

PR #1709, opened 2023-08-30
Fix tests from #1708 machine variable persistence

avanwinkle:
Thank you @atummons for catching the test breakage in #1708 . When I saw the failures I was distracted by the sea of `'Event loop is closed.'` failures and didn't catch that I had actually introduced new ones. This PR won't fix all the tests to pass green, but it _does_ fix the tests in _test_MachineVariables.py_ so we're at least not worse off than we were.


atummons:
Perfect! Same thing I tweaked on local, so looks good. The other two were mine that I caught but hadn't been pushed before your PR. All looks good now! I have another PR, but didn't want to push until we get back to green. Thanks!

PR #1723, opened 2023-11-17
-V option was not honored.

mrechte:
The console QueueListener handler was not honouring the loglevel of the console_log handler.

Using -v (without -V) was producing DEBUG messages on the console.

Also change the usage note on -V option.


avanwinkle (approved):
Great fix, thanks!

PR #1770, opened 2024-03-04
Service Menu: Light Chains

avanwinkle:
This PR adds a new service menu option for Light Chains, to turn on all lights in a single chain for verifying chain power and connection fidelity.

It is tested on FadeCandy (each output pin is a chain), modern FAST (each number modulo 64 is a chain), and matrix lights (each row/column is a chain). Matrix lights will offer both rows and columns as chains for cross-matrix debugging, but note that it is currently hard-coded to an 8x8 matrix. Multiple platforms are also supported, e.g. a machine using both FAST and FadeCandy lights will render chains for each of them.

Other lighting platforms may require additional work, please reach out if you are using one.

The light chain menu will turn on one chain at a time, with the same index-cycle and color-cycle behavior as the single light test.

untangle these lights

### DIFF
--- a/mpf/modes/service/code/service.py
+++ b/mpf/modes/service/code/service.py
@@ -11,7 +11,7 @@ from mpf.core.switch_controller import MonitoredSwitchChange
 from mpf.core.utility_functions import Util
 
 ServiceMenuEntry = namedtuple("ServiceMenuEntry", ["label", "callback"])
-
+LightChainMap = namedtuple("LightMap", ["board", "chain", "light"])
 
 class Service(AsyncMode):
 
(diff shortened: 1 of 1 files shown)


cobra18t:
Is there an ability to adjust the brightness of the "on?" For OPP, a chain is up to 256 LEDs and that may be a fair bit of power draw that is otherwise unnecessary or unplanned for in a game.

Alternatively, is there a path to genericize the modulo across platforms?

avanwinkle:
@cobra18t Thanks for the feedback! There is not currently an option to adjust the brightness, though it does include color selection so you could run the chain at red, green, or blue for 33% of the draw (or yellow for 66%). While it may be un_planned_ to run 256 chained lights at once, it's still _possible_, which would indeed draw power concerns and most would advise some power boosts to ensure nothing gets funky. That said, adding some "color" options for reduced brightness would be simple enough to do.

As far as cross-platforms go, the feature tracks itself based on the light number so unless a platform uses an unconventional structure for identifying lights it should work (currently supported are single numbers, double numbers (with a dash), and triple numbers (also dashed)).

That said, it would be possible to make the setup more configurable, e.g. defining the maximum number of lights in a "chain" so that longer chains could be broken up, as well as the shape of a matrix and customizable names for the chains.

PR #1784, opened 2024-03-22
score reels no longer activate chime on score reset.

mrechte:
This is to avoid noise during score reset.

### DIFF
--- a/mpf/devices/score_reel.py
+++ b/mpf/devices/score_reel.py
@@ -41,6 +41,9 @@ class ScoreReel(SystemWideDevice):
 
         self._destination_value = 0
         # Holds the index of the destination the reel is trying to advance to.
+        
+        self._quiet = True
+        # Whether to use chime when advancing
 
         self._runner = None
         # asyncio task which advances the reel
@@ -154,7 +157,8 @@ class ScoreReel(SystemWideDevice):
         while self._destination_value != self.assumed_value:
             self.machine.events.post('reel_{}_will_advance'.format(self.name))
             wait_ms = self.config['coil_inc'].pulse(max_wait_ms=500)
-            self.machine.events.post('reel_{}_advancing'.format(self.name))
+            if not self._quiet:
+                self.machine.events.post('reel_{}_advancing'.format(self.name))
             previous_value = self.assumed_value
 
             await asyncio.sleep((wait_ms + self.config['repeat_pulse_time']) / 1000)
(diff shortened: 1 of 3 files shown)


mpf/devices/score_reel.py
```diff
@@ -154,7 +157,8 @@ async def _advance_reel_if_position_does_not_match(self):
         while self._destination_value != self.assumed_value:
             self.machine.events.post('reel_{}_will_advance'.format(self.name))
             wait_ms = self.config['coil_inc'].pulse(max_wait_ms=500)
-            self.machine.events.post('reel_{}_advancing'.format(self.name))
+            if not self._quiet:
+                self.machine.events.post('reel_{}_advancing'.format(self.name))
```
avanwinkle, line 161: This is a nice fix, but limited to your specific application. An alternative approach would be to post the event with an argument like `reset: true` to indicate whether the reel is resetting or not. This way there's still an event for people who want/need to know the advancement of the reel, and chimes can be made conditional on that argument to determine whether to chime or not.

mpf/devices/score_reel_group.py
```diff
@@ -75,7 +75,7 @@ def chime(cls, chime, **kwargs):
         del kwargs
         chime.pulse()
 
-    def set_value(self, value):
+    def set_value(self, value, quiet=False):
```
avanwinkle, line 78: Aligned with my previous comment, changing the argument from `quiet` to `is_reset` will provide usecase-independent context and help future developers understand what's happening.

avanwinkle (changes requested):
Thank you for the improvement! I think it can be abstracted a little more to be useful in a wider range of situations, but it's a great fix!

mpf/devices/score_reel_group.py
mrechte: Hello. Please could you help me ?
Everything starts from the score_reel_controller, that will call set_value() of ScoreReelGroup class in 3 situations:
1. _game_starting
2. _rotate_player
3. _score_change
If I understand correctly, ScoreReel would always 'post reel_{}_advancing', with an extra boolean argument 'reset'.
Then we would modify the chime handler of ScoreReelGroup, that would not pulse if reset is True.
How do I retrieve the reset argument in chime method ? Is it in kwargs, or do I just add a reset arg after chime ?

avanwinkle: When you post an event, you can include kwargs like so:
```
self.machine.events.post(f'reel_{self.name}_advancing', is_reset=False)
```

And then in an event listener, you can make it conditional:
```
event_player:
  reel_myreel_advancing{not is_reset}: play_chime
```

bosh:
We just added event management of chimes being enabled, via https://github.com/missionpinball/mpf/pull/1972/changes . I think it can cover the needs presented in this PR, so I'm proposing closing this one.

PR #1793, opened 2024-06-04
Fast channel offset fix

bosh:
In FAST.py:679 there is an issue if you use channel offset declaration on your light number property, such as exp_playfield-2-6-2.

Normally you'd just declare 2-6 which is implicitly 2-6-0, and consumes 2-6-1 and 2-6-2 as well. When you declare 2-6-1, you would get 261, 262, and 270 -- only one of the three channels rolled over to the next full light number.

But if you use 2-6-2 you currently get 262, 270, and 280. This is because the loop is reusing the same parts array on every iteration instead of a copy, and the rollover code modifies that initial array. It also explicitly codes 0 as the final part instead of channel+i-3.

This should generate the correct sequence, but probably still needs touching up and could use an automated test as well.

### DIFF
--- a/mpf/_version.py
+++ b/mpf/_version.py
@@ -10,7 +10,7 @@ PyPI.
 
 """
 
-__version__ = '0.57.2.dev1'  # Also consider whether MPF-MC pyproject.toml should be updated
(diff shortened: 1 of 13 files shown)


bosh:
Hm I did something wrong here. Closing.

PR #1794, opened 2024-06-04
FAST light config has a bug when using channels

bosh:
In FAST.py:679 there is an issue if you use channel offset declaration on your light number property, such as exp_playfield-2-6-2.

Normally you'd just declare 2-6 which is implicitly 2-6-0, and consumes 2-6-1 and 2-6-2 as well. When you declare 2-6-1, you would get 261, 262, and 270 -- only one of the three channels rolled over to the next full light number.

But if you use 2-6-2 you currently get 262, 270, and 280. This is because the loop is reusing the same parts array on every iteration instead of a copy, and the rollover code modifies that initial array. It also explicitly codes 0 as the final part instead of channel+i-3.

This should generate the correct sequence, but probably still needs touching up and could use an automated test as well.

### DIFF
--- a/mpf/platforms/fast/fast.py
+++ b/mpf/platforms/fast/fast.py
@@ -676,13 +676,14 @@ class FastHardwarePlatform(ServoPlatform, LightsPlatform, RgbDmdPlatform,
                 if 0 <= channel <= 2:
                     result = []
                     for i in range(3):
+                        working_parts = parts.copy()
                         if i + channel > 2:
                             # Channel rolls over, increment the LED number
-                            parts[3] = str(int(parts[3]) + 1)
-                            parts[4] = '0'
+                            working_parts[3] = str(int(working_parts[3]) + 1)
+                            working_parts[4] = str((channel + i) % 3)
                         else:
-                            parts[4] = str(channel + i)
-                        result.append({'number': '-'.join(parts)})
+                            working_parts[4] = str(channel + i)
+                        result.append({'number': '-'.join(working_parts)})
                     return result
                 raise AssertionError(f"Invalid LED channel: {channel}")
             raise AssertionError(f"Invalid LED number: {number}")
--- a/mpf/tests/machine_files/fast/config/exp.yaml
+++ b/mpf/tests/machine_files/fast/config/exp.yaml
@@ -63,11 +63,11 @@ lights:
   led3:
     number: brian-1-3
   led4:
-    number: brian-b1-2-1
+    number: brian-b1-2-1-0
   led5:
-    number: brian-b1-2-2
+    number: brian-b1-2-2-1
   led6:
-    number: aaron-b2-1-1
+    number: aaron-b2-1-1-2
   led7:
     number: aaron-1-1
   led8:
--- a/mpf/tests/test_Fast_Exp.py
+++ b/mpf/tests/test_Fast_Exp.py
@@ -89,6 +89,12 @@ class TestFastExp(TestFastBase):
         self.assertEqual(self.led5.hw_drivers['green'][0].number, '88121-2')
         self.assertEqual(self.led5.hw_drivers['blue'][0].number, '88122-0')
 
+        # Intentionally-failing test to prove that third channel rollover has an issue
+        self.assertEqual(self.led6.hw_drivers['red'][0].number, '89200-2')
+        self.assertEqual(self.led6.hw_drivers['green'][0].number, '89201-0')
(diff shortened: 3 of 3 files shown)


avanwinkle:
Thanks for the fix! Do you have an example from your config that causes this issue? If you're not comfortable with the automated tests yet, I could put one together. Having a known-triggering config would speed that up a bunch :)

bosh:
I'm down to try the tests myself, but I can't get single file tests running.

`python3 -m unittest discover mpf/tests` worked to get me 859 passing (2 skips)
but trying the single file test with `mpf test mpf/tests/platforms/fast.py` first gave me a kivy missing error (havent needed it yet since not running mc) and after installing that, now gives an issue importing: `mpf.file_interfaces.yaml_roundtrip`

```
   File "/home/knight/mpf80env/bin/mpf", line 8, in <module>
     sys.exit(run_from_command_line())
   File "/home/knight/workspace/mpf/mpf/commands/__init__.py", line 194, in run_from_command_line
     CommandLineUtility(path).execute()
   File "/home/knight/workspace/mpf/mpf/commands/__init__.py", line 174, in execute
     _module = import_module('mpf.commands.%s' % command)
   File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
     return _bootstrap._gcd_import(name[level:], package, level)
   File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
   File "<frozen importlib._bootstrap>", line 991, in _find_and_load
   File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
   File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
   File "<frozen importlib._bootstrap_external>", line 848, in exec_module
   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
   File "/home/knight/workspace/mpf/mpf/commands/test.py", line 12, in <module>
     from mpf.tests.MpfDocTestCase import MpfDocTestCase, MpfDocTestCaseNoFakeGame
   File "/home/knight/workspace/mpf/mpf/tests/MpfDocTestCase.py", line 11, in <module>
     from mpf.file_interfaces.yaml_roundtrip import YamlRoundtrip
 ModuleNotFoundError: No module named 'mpf.file_interfaces.yaml_roundtrip'

```
I searched briefly around the repo and missionpinball org on github, but couldnt find this defined anywhere. Any tips?

bosh:
I have this single-file config crashing `mpf -bt` due to the expected (bugged) duplicate light issue against mpf `dev` branch but starting into a game and sitting happily when running against the pr branch.

I'm still down to try and get tests done myself, but I made this anyway so here it is :)

```
#config_version=6

hardware:
  platform: fast

fast:
  exp:
    port: /dev/ttyACM1
    boards:
      exp_playfield:
        model: FP-EXP-0071

lights:
  #we assume channels 2-1-0 and 2-1-1 are already in use, causing the offset, but we dont actually need a real light declaration to just use offsets for fun
  light_two:
    number: exp_playfield-2-1-2 #this should use 2-1-2, 2-2-0, and 2-2-1, but instead uses 2-1-2, 2-2-0, and 2-3-0 
  light_three_should_not_collide:
    number: exp_playfield-2-3-0 #this channel is not likely to be a LED start channel in a real world situation, but it should not collide and crash either
```

bosh:
Pushed an update with a failing unit test for the base case of this function (not the bug case I have).

`/tests/test_Platform_Fast.py`
test_Fast.py seemed like the more correct name, but the file exists and seems to be a test against actual FAST hardware. I steered clear.

I'm kind of stuck now - I have no idea how to instantiate a good-enough instance of the test subject. I think this is probably going too far as a newbie to this codebase, but the object design of the FastPlatformHardware class feels like it owns far too much, and this utility calculator method (and its three privates that it calls into) should move into a use case or stateless utility function -- that would be super easy to test, and could help shrink that class down. I'm willing to give it a shot if it's not a step in the wrong direction!

Or maybe I'm trying to write the wrong level of test for this? Grepping the whole codebase, lots of platforms have "parse_light_number_to_channels" but none of them are directly unit tested.

bosh:
Alright I found the machine configs, and then the relevant test. I restarted the branch and added three commits--
The first adds more testing to the file that should all be green (was locally :) )
The second adds a failing test without changing any code, to prove the issue
The third adds the fix and should be all green!

Thanks @avanwinkle for the help!

avanwinkle (approved):
Thanks for the fix and the tests!!

PR #1795, opened 2024-06-23
Add logging to shows.

mrechte:
Although show controller features logging, it seems it was not yet implemented in shows.


avanwinkle (changes requested):
Excellent improvement! 

My only note is that `_ctrl` is an unestablished pattern that's not particularly clear about what it is or why it's there. Since there are only four references to it, I think it'd be worth the extra few characters to explicitly call `self.machine.show_controller.info_log()` rather than `self._ctrl.info_log()`. 

Other than that, fantastic work!

avanwinkle (approved):
Thanks!

PR #1796, opened 2024-06-24
Fix fast exp hex decode

bosh:
I branched this off the last green 0.80.x, so it should be all green still

-first commit adds a failing test (not even any assertions -- just defining a FAST EXP LED using "previous" is enough to trigger lower-case-hex behavior if it's on a board with id such as "B4".
-second commit adds hex normalization and passes tests again. Not sure if I picked the best hex massaging function from Util, but it was better than just .upcase() in the callsite :)

### DIFF
--- a/mpf/platforms/fast/fast.py
+++ b/mpf/platforms/fast/fast.py
@@ -512,6 +512,7 @@ class FastHardwarePlatform(ServoPlatform, LightsPlatform, RgbDmdPlatform,
 
         return switch
 
+    # pylint: disable-msg=too-many-locals
     def configure_light(self, number, subtype, config, platform_settings) -> LightPlatformInterface:
         """Configure light in platform."""
         del platform_settings
@@ -567,7 +568,10 @@ class FastHardwarePlatform(ServoPlatform, LightsPlatform, RgbDmdPlatform,
 
             elif int(parts[0]) > 255:
                 # EXP LED in int form, which is how "previous:" values are calculated
-                this_led_number = hex(int(parts[0]))[2:]
+
+                raw_hex_string = hex(int(parts[0]))[2:]  # lowercase with 0x prefix stripped"
+                this_led_number = Util.normalize_hex_string(raw_hex_string, len(raw_hex_string))
+
                 exp_board = self.exp_boards_by_address[this_led_number[:2]]
 
                 if this_led_number not in self.fast_exp_leds:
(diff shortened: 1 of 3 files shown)


mpf/platforms/fast/fast.py
```diff
@@ -567,7 +568,10 @@ def configure_light(self, number, subtype, config, platform_settings) -> LightPl
 
             elif int(parts[0]) > 255:
                 # EXP LED in int form, which is how "previous:" values are calculated
-                this_led_number = hex(int(parts[0]))[2:]
+
+                raw_hex_string = hex(int(parts[0]))  # lowercase and prefixed with "0x"
```
avanwinkle: Minor nit, if you were to keep the `[2:]` here then you wouldn't need to have both it and the ` - 2` in the next line, but it's not a huge deal either way.

avanwinkle (commented):
This looks great! Thanks for investigating and making a fix.

I see you added some new lights to the test file, but I don't see any new assertions to verify that the new lights are being correctly parsed and ordered. Is there an assertion or two you could add to the test to make sure this doesn't get broken again?

Speaking of broken, I broke the tests on 0.80 branch, sorry. I'm working on that now. But this fix would be helpful for all users, so rebasing the changes against the `dev` branch would benefit 0.57 games and then get pulled into 0.80 as well!

mpf/platforms/fast/fast.py
bosh: Oh yeah I had a separate variable at first for hex(int(.)) but got a linter error on too many local vars. I think I added a linter exception above, so I could take this back out again

bosh: No wait I confused myself - too many variations of this same line. I think my thought for this version was that something like "0x9000beef" made more sense as a value I would pass to a cleanup util function than, but then none quite matched what I wanted and I had to strip the 0x here anyway.

bosh:
Sure I can add assertions - as far as I could tell a few of the lights in the file just existed to test for crashes rather than getting asserted on directly, but I'll think of something for mine.

Re 0.57 backporting - .... I wrote something here but I was wrong. For some reason I thought the "previous" handling was new to 0.80.x but I checked again and it's in 57- so yeah lemme rebase this for the older versions

bosh:
Alright I did the things :)

avanwinkle (approved):
Excellent, thanks and nice work!

PR #1820, opened 2024-08-16
Initial Implementation for display_segment update_method replace

worldpeace-germany:
config_spec.yaml:
changed default from `not_set` to `off`, left `not_set` in enum though I believe it can be removed. I couldn't see why not_set is a good default. If really needed for the changes in update_method `replace` I can add code to exchange `not_set` for `off`.

segment_display:
Only changed code for update_method `replace`, `stack` should behave like before. I tested with various segment displays shows and it seems to behave well. Note:
- for `replace` the `priority` parameter is not used (and doesn't make sense if you don't have a stack)
- my shows work with `replace`, but completely misbehave if using `stack`
- I would propose (but have not done it since unsure about consequences for others) to change the default to `replace` in here https://github.com/missionpinball/mpf/blob/35a84ba29c54acc6ce1afab28bb0e9edca07e301/mpf/devices/segment_display/segment_display.py#L139
- I am not able to test `color` changes since my display is single color (though would believe it should work if it works for `stack`)
- `flashing`, `transition`, `transition_out`, `expire` parameter tested (and working)

If the PR gets accepted I would do some additons to mpf-docs 0.80. Just would need to know how to contribute version specific changes.


worldpeace-germany:
In regards to the last commit [fixed and clean up of color handling in segment displays](https://github.com/missionpinball/mpf/pull/1820/commits/ae93e49173dcee4bab6d4b570d950b983ba4925c)

In general there are some issues when defining segment display shows due to the fact how colors are being handled and how that is being known to user and if that is perceived as a logical definition flow.

In segment display the default color (white) is used

https://github.com/missionpinball/mpf/blob/35a84ba29c54acc6ce1afab28bb0e9edca07e301/mpf/devices/segment_display/segment_display.py#L90

to expand the given color array to the right until the length of the segment display is being filled up. So for a 4 digit display, specified as default color "blue, yellow" this would end up as "blue, yellow, white, white".

During initlize an empty string is sent to the segment display

https://github.com/missionpinball/mpf/blob/35a84ba29c54acc6ce1afab28bb0e9edca07e301/mpf/devices/segment_display/segment_display.py#L101

, each string is colored

https://github.com/missionpinball/mpf/blob/35a84ba29c54acc6ce1afab28bb0e9edca07e301/mpf/devices/segment_display/segment_display_text.py#L94

Since the length is 0, the left_pad_color is used to set the color. So it is not the default color being used, the color set now for this example is "blue, blue, blue, blue". Which on the first view doesn't matter since the display is blank, but at later points in time the code prefers to use the current_state color and not the default_color if no color is being sent, for example

https://github.com/missionpinball/mpf/blob/35a84ba29c54acc6ce1afab28bb0e9edca07e301/mpf/devices/segment_display/segment_display.py#L303

if the other digits cannot display blue they will simply stay blank.

In other parts of the code, if you start a transition the given colors are being expanded

https://github.com/missionpinball/mpf/blob/35a84ba29c54acc6ce1afab28bb0e9edca07e301/mpf/devices/segment_display/segment_display.py#L186, where expanded could as well mean shortend to the length of the text and not the length of the display, so later the empty places are being padded with a
color again (see above). If the colors need to be expanded then this

https://github.com/missionpinball/mpf/blob/35a84ba29c54acc6ce1afab28bb0e9edca07e301/mpf/devices/segment_display/segment_display.py#L230

is done by taking the most right color and reuse it for all the other text to the right of this position (so no default color being used).

If a transition text is being used and no color is specified, then the color of the first character of the new text is being used (again not the default color).

https://github.com/missionpinball/mpf/blob/35a84ba29c54acc6ce1afab28bb0e9edca07e301/mpf/devices/segment_display/transitions.py#L139

In general we have default color, sometimes padded to the left and sometimes padded to the right. Question is if the user can follow that behavior or if it is over complicated and thus introduces user mistakes.

The PR will change some of the behavior to make it more consistent

- unchanged is that rather the current_state is used and not the default_color
- changed is that if color information is missing that the as the default color white will be used, padded to the left, that ensure that at least something is

visible and not that because of a wrong color invisible text is shown and confuses the user. But the PR not only makes the handling more consistent but I would consider it a bug if a color is used for padding which might be able to be used for the digits of that segment display.

Once the PR is accepted I will update the documentation to explain the color handling in detail for the users.

mpf/devices/segment_display/segment_display.py
```diff
@@ -155,11 +165,56 @@ def add_text_entry(self, text, color, flashing, flash_mask, transition, transiti
...
-                                           color)
-        self._update_display(SegmentDisplayState(text, flashing, flash_mask))
+
+        ###############################
+        #Store current color and text as previous text/color of next run even if no transition in this step,
+        #the next step might have a transition, that the old text/color needs to be included into that transition
+        ###############################
+
+        # Handle new and previous text
+        if self._previous_text:
```
avanwinkle: For pythonic convenience, this could be rewritten as:
```
previous_text = self._previous_text or ""
```

Same below with `previous_color`

mpf/devices/segment_display/segment_display.py
```diff
@@ -155,11 +165,56 @@ def add_text_entry(self, text, color, flashing, flash_mask, transition, transiti
...
+
+        # Handle new and previous color
+        if self._previous_color:
+            previous_color = self._previous_color
+        else:
+            previous_color = self._default_color
+        self._previous_color = color  # Save the new color as the next previous color
+
+        if transition or self._previous_transition_out:
+            if transition: #if transition exists, then ignore transition_out of previous text/color
```
avanwinkle: This second nested block could be removed by passing both possible values as arguments.
```
if transition or self._previous_transition_out:
    transition_conf = TransitionManager.get_transition(self.size,
                                                       self.config['integrated_dots'],
                                                       self.config['integrated_commas'],
                                                       self.config['use_dots_for_commas'],
                                                       transition or self._previous_transition_out)
    self._previous_transition_out = transition_out or None
```

mpf/devices/segment_display/segment_display.py
```diff
@@ -155,11 +165,56 @@ def add_text_entry(self, text, color, flashing, flash_mask, transition, transiti
...
+            if transition_out:  #in case transition_out is set we need to preserve it for the next step but only after the previous transition_out is in this step's config
+                self._previous_transition_out = transition_out
+
+            #start transition
+            self._start_transition(transition_conf, previous_text, text,
+                                   previous_color, color,
+                                   self.config['default_transition_update_hz'], flashing, flash_mask)
+
+        else: #No transition configured
+            if transition_out:  #in case transition_out is set we need to preserve it for the next step
```
avanwinkle: The gotcha here is that if there is no `transition_out` on _this_ transition but the previous transition _did_ have an out, then `self._previous_transition_out` will still have the value of the last transition out. Best to be explicit and not use an if statement.

```
self._previous_transition_out = transition_out or None
```

Which is also in the above if block, so it could be called just once at the end of this method outside of the if/else block.

mpf/devices/segment_display/segment_display.py
```diff
@@ -171,7 +226,7 @@ def add_text(self, text: str, priority: int = 0, key: str = None) -> None:
     def remove_text_by_key(self, key: Optional[str]):
         """Remove entry from text stack."""
         if self.config['update_method'] != "stack":
-            self.info_log("Segment display 'remove' action is TBD.")
+            self.add_text_entry("", self._previous_color, FlashingType.NO_FLASH, "", None, None, 100, key)
```
avanwinkle, line 227: This is a good use of an existing method to clear out a display, but the `key` part is misleading because the argument is passed to the method to define which key to _remove_, but it's actually _adding_ empty text with that key. 

I would suggest instead that this behavior check the current text to see if it has the requested key, and if so remove the text and pass an empty key, and if not do nothing.

mpf/devices/segment_display/segment_display_text.py
```diff
@@ -76,28 +76,36 @@ def _create_characters(cls, text: str, display_size: int, collapse_dots: bool, c
...
-        colors = colors[-len(uncolored_chars):]
-        for char_code, char_has_dot, char_has_comma in uncolored_chars:
-            color = colors.pop(0) if colors else default_right_color
-            char_list.append(DisplayCharacter(char_code, char_has_dot, char_has_comma, color))
 
-        # ensure list is the same size as the segment display (cut off on left or right justify characters)
-        current_length = len(char_list)
+        # Adujust the color array if needed
+        color_length = len(colors)
+        if color_length > display_size:
```
avanwinkle: How much do we really want to handle here? I think it is reasonable for a designer to send one color (for all segments) or a list of colors (one for each segment), and those are the two options. 

It doesn't seem reasonable to expect MPF to "correctly" handle coloring a seven segment display when you send it four colors or something 🤷. I think we should strip down all this color-filling logic and throw an exception if the list of colors is neither 1 nor the display size. Unless I'm missing a use case?

mpf/devices/segment_display/segment_display_text.py
```diff
@@ -76,28 +76,36 @@ def _create_characters(cls, text: str, display_size: int, collapse_dots: bool, c
...
         if current_length > display_size:
             for _ in range(current_length - display_size):
-                char_list.pop(0)
+                uncolored_chars.pop(0) # remove very left char of array if too long
         elif current_length < display_size:
             for _ in range(display_size - current_length):
-                char_list.insert(0, DisplayCharacter(SPACE_CODE, False, False, left_pad_color))
+                uncolored_chars.insert(0, (SPACE_CODE, False, False))
+
+        for _ in range(len(uncolored_chars)):
```
avanwinkle: By convention, using the underscore as a variable in a loop is meant to indicate that the variable is never used, i.e. the loop code doesn't care which iteration it's on. If you are going to use that variable, a letter is preferred (either the generic `i` or the first letter of what's being iterated, in this case `c`).

mpf/devices/segment_display/segment_display.py
```diff
@@ -145,7 +148,14 @@ def add_text_entry(self, text, color, flashing, flash_mask, transition, transiti
 
         This will replace texts with the same key.
         """
+
+        if len(color) == 0:
```
avanwinkle: This could be more readable (and do the same thing) with a generic truthiness check:
```
if not color:
  color = self._current_state.text.get_colors()
```

avanwinkle (commented):
This is a lot of great work, thanks for tackling this part of the code that most people don't ever look at.

I've noted in my comments that I think we could strip down a lot of the color-filling logic because I don't think it should be MPF's job to try and figure out what the designer means. The designer should be clear and provide the correct number of colors.

As far as `priority`, I think it's worth keeping for `replace` mode because there may be situations where multiple modes have segment_display_players and the highest priority should have precedence. Otherwise designers will have a lot of manual work to do if they want to suppress lower-priority mode text from showing up during a higher-priority mode (e.g. multi ball)

avanwinkle (changes requested):
Oops, I meant to post that review as Request Changes, not Comment. See above for requested changes :)

mpf/devices/segment_display/segment_display_text.py
avanwinkle: On second thought, that's not necessary because `uncolored_chars` is iterable, so you can use `enumerate()` to get both the index and the value (which saves time on all the lookups in DisplayCharacter).

```
for i, char in enumerate(uncolored_chars):
  color = colors[i]
  char_list.append(DisplayCharacter(char[0], char[1], char[2], color))
```

mpf/devices/segment_display/segment_display.py
worldpeace-germany: I would tend to disagree on this point. But maybe I misunderstood something
here. I tried to change the code as little as possible that is why we have
this early return (which I would usually try to avoid), maybe that is
misleading. Initially that function was only implemented for the `stack` case
(and that implementation has **not** changed).

For the `replace` case I would claim neither `key` nor `priority` is relevant since
we don't have a stack where we can remove something based on a key (or sort
based on a priority). Of course a `key` is passed as well for the `replace` case,
but there is nothing you can really do with it (in the `replace` case). So
removal in that case means rather emptying the display. Especially since that
remove method is as well called when the context is being removed (when a show
gets the `stop` command).

On a side note: I am working on one fix for the show `stop` since if you stop a
show then the current transition is still being played (in the `replace` case)
and a `transition_out` would be as well played. But it still empties the display.

If you really believe that the `key` is relevant for the `replace` case then please rephrase your feedback. Happy to understand it better and to improve.

mpf/devices/segment_display/segment_display_text.py
worldpeace-germany: Very true, that is what I asked myself as well. I tried to be minimal invasive with my code changes, so that is what I came up with. But it seems we have the same understanding here and thus I am happy to change the color handling. Will provide a new PR in the next days including all the other feedback you gave.

mpf/devices/segment_display/segment_display.py
avanwinkle: Here's a use case I'm picturing where key and priority would be helpful, let me know what you think.

A player starts a non-exclusive hurry-up mode (priority 1000) with a jackpot on the left loop, and that hurry-up mode sends the text "HIT LEFT LOOP" to a display. Instead of going for the hurry-up, the player instead starts a multi ball (priority 5000) and the multi ball mode sends the text "JACKPOT: 5M" to the display. During the multi ball, the hurry-up mode times out and ends.

By default, the hurry-up mode is set to remove the "HIT LEFT LOOP" text from the display, so it will send a command to the displays to replace the text with empty string. Without a key, the display doesn't know that the text being shown is _not_ the text that's expected to be removed, and will instead remove the "JACKPOT: 5M" text from the display even though the multi ball is still running. As a result, the multi ball will have a blank display.

Reverse the situation and start the multi ball before the hurry-up. The multi ball mode is higher priority so its display text should have precedence, and the hurry-up display text should be ignored. Otherwise the multi ball "JACKPOT: 5M" will turn into "HIT LEFT LOOP" and then disappear after the hurry-up, again leaving the multi ball with a blank display.

I realize that this type of use case is exactly what the `stack` mode is meant to do, but I believe there is value in `replace` also respecting keys and priorities because that creates more *consistent* behavior in the game. Designers will have the option to override the behavior by specifying keys and priorities if they want to, but regardless the behavior will be the same every time. Ignoring keys and priorities will yield different displays based on timing, which is not ideal.

Thoughts?

worldpeace-germany: The described use case is very valid and is very much a use case for the `stack` case. The initial implementation of `replace`

https://github.com/missionpinball/mpf/blob/7a9665173916bb6621366893c5d053d1cc329657/mpf/devices/segment_display/segment_display.py#L157

indicates that the text is written straight to the display.  Can you describe in words what would be the logic of `replace` if it includes `key` and `priority` and where this behavior differs from the `stack` case? I simply don't see the logic of such an implementation.

avanwinkle: You're right, I'm overthinking it and trying to keep stack-like functionality where it doesn't belong. The concept of replace doesn't need to have any considerations for priority or key or what's currently on the display, and should always overwrite when called.

worldpeace-germany:
@avanwinkle May I check if you believe that something is missing to accept that PR? Feedback to improve is of course more than welcome.

mpf/config_players/segment_display_player.py
```diff
@@ -106,12 +106,23 @@ def _remove(self, instance_dict, key, display):
...
 
     def clear_context(self, context):
-        """Remove all texts."""
-        instance_dict = self._get_instance_dict(context)
-        for display, keys in instance_dict.items():
-            for key in dict(keys).keys():
-                self._remove(instance_dict=instance_dict,
-                             key=key, display=display)
+
+        ##############
```
avanwinkle: No need for the larger comment block lines here, just the comments are good

mpf/config_players/segment_display_player.py
```diff
@@ -106,12 +106,23 @@ def _remove(self, instance_dict, key, display):
...
+        # Remove all texts. Ignore what keys are available, that will be checked later in the segment display code.
+        # Especially important for update_method replace since there are no keys.
+        ##############
+
+        instance_dict = self._get_instance_dict(context) # key of the dict is the display, the value is another dict
+
+        for display, keys_dict in instance_dict.items(): # keys_dict key is the show key, the value is a boolean (with yet unknown usage)
+            if(keys_dict): #depending on the situation the keys_dict might be empty, still need to clear the display
```
avanwinkle: This can be `if keys_dict:` directly, no need for parenthesis

mpf/config_players/segment_display_player.py
```diff
@@ -106,12 +106,23 @@ def _remove(self, instance_dict, key, display):
...
+                for key in dict(keys_dict).keys():
```
avanwinkle, line 118: In what situation would keys_dict *not* be a dict? It feels weird to convert it to dict here instead of `for key in keys_dict.keys()`

mpf/config_players/segment_display_player.py
```diff
@@ -106,12 +106,23 @@ def _remove(self, instance_dict, key, display):
...
+                    display.clear_segment_display(key)
+                    if instance_dict[display][key] is not True:
```
avanwinkle, line 120: I'm not really sure what this means, in what scenario would the value be `True` and what would it be otherwise? Could you add a comment?

mpf/devices/segment_display/segment_display.py
```diff
@@ -155,11 +169,46 @@ def add_text_entry(self, text, color, flashing, flash_mask, transition, transiti
...
 
         # For the replace-text update method, skip the stack and write straight to the display
-        new_text = TextTemplate(self.machine, text).evaluate({})
-        text = SegmentDisplayText.from_str(new_text, self.size, self.config['integrated_dots'],
-                                           self.config['integrated_commas'], self.config['use_dots_for_commas'],
-                                           color)
-        self._update_display(SegmentDisplayState(text, flashing, flash_mask))
+
+        ###############################
+        #Store current color and text as previous text/color of next run even if no transition in this step,
```
avanwinkle: The linter will complain that you don't have a space between the # and the comment text, please add one and remove the top/bottom block comment lines

mpf/devices/segment_display/segment_display_text.py
```diff
@@ -74,30 +74,32 @@ def _embed_dots_and_commas(cls, text: str, collapse_dots: bool, collapse_commas:
...
-        default_right_color = colors[len(colors) - 1] if colors else None
         uncolored_chars = cls._embed_dots_and_commas(text, collapse_dots, collapse_commas, use_dots_for_commas)
-        colors = colors[-len(uncolored_chars):]
-        for char_code, char_has_dot, char_has_comma in uncolored_chars:
-            color = colors.pop(0) if colors else default_right_color
-            char_list.append(DisplayCharacter(char_code, char_has_dot, char_has_comma, color))
 
-        # ensure list is the same size as the segment display (cut off on left or right justify characters)
-        current_length = len(char_list)
+        # Adujust the color array if needed
```
avanwinkle: Typo :)

avanwinkle (commented):
My apologies for letting this languish for so long, thanks for your patience. A few small notes and one previous unaddressed comment, please take a look. Thanks!

avanwinkle:
Looks good! Some linter/formatting fixes and some tests needs updated, but otherwise a strong improvement!

worldpeace-germany:
@avanwinkle I hope I have implemented all feedback. Can you take a look to check if I have missed something? I am aware that, test cases are failing, that is mainly due to the fact that we changed the behavior on what colors the system tried to "guess" in the past. We have agreed to simplify it (I have explained that in my PR for the documentation). I have to admit I have no idea how to fix the testing code. If someone gives me a hand I am willing to learn and fix it.
I am as well aware that my explanations on the dictionaries being used are somewhat weak. The reasons are a) debugging was a long time ago... :-) and b) I have not fully understood in mpf what the boolean values in the second dict are good for. Again, with some help and input I am willing to spend time on it if necessary.

worldpeace-germany:
@avanwinkle Finally found the time to get into Python testing and fixed all the failing tests. Sorry that it took a while, but time is limited and there was a learning curve... If you have time please take a look and let me know if you spot something before the PR can be accepted.

worldpeace-germany:
@bosh may I check with you since you "touched" my PR what I can do to get it accepted? From my pov it is all complete. Even if there is a reason why it will never be accepted it would be nice to get some feedback. Thanks!

bosh:
No worries! I've recently gained commit abilities, so I've been trying to clear out old things by merging or killing :)
What I did was a "rebase", so that it would be a PR based on the most recent dev code (good for reducing merge conflicts, removing merge commits, and cleaning up history) -- however this does somewhat weird things with the branches on Github, and it may cause issues with your own pushes to the branch. Your best bet for any further changes is to rebase on latest `dev` yourself and force push the branch again.

It looks like it's all green, but I haven't really dug in on the content here yet. I see in the first reviewed file, `mpf/config_players/segment_display_player.py`, Anthony has a couple review comments that do seem reasonable to me to still be cleaned up. I haven't investigated the longer comments, maybe I'll have time in the next couple days to learn this corner of the codebase.

worldpeace-germany:
@bosh  Thanks for looking into it. I believe I have fixed all comments from Anthony. If I have missed something then of course I will look into it at the moment I simply don't see what I might have missed. Anyways, if you take a look and find something then please let me know and I will update the code where needed.

bosh:
Yeah it looks like just two minor nits in the file I mentioned. It looks like the very long comment chain is actually resolved by doing nothing. I'll give it a once over this weekend, since I don't want to merge anything I don't understand.

mpf/config_players/segment_display_player.py
bosh: I think the weirdness of this here comes from copying `_remove`. I think the resolution here would be to make a second _remove function (with an additional word or two of context in the name) that looks like the existing _remove function, so the quirks of the dict management and delay removal will be more apparently parallel (and possibly refactorable)

bosh:
I finally have a working segment display unit, so I can play with this for real. I think probably we can get this in after I cut the next dev version this weekend (that's `0.(81|58).0.dev3`, this would then be part of pending dev4).

I'll test on a rebased version and can push that up if it works, unless you'd prefer to rebase to dev before then. This has been pending for far too long, so I'm happy to manage the rebase+test and get it in, no pressure :)

worldpeace-germany:
Sounds good to me, please go ahead. Happy to hear your test results, I hope I have not missed anything :-) If I have let me know and I will help to finish it. It probably doesn't matter, just in case I tested on a Cobra board and the segment displays are from Cobra as well.

mpf/devices/segment_display/segment_display.py
```diff
@@ -145,6 +157,11 @@ def add_text_entry(self, text, color, flashing, flash_mask, transition, transiti
 
         This will replace texts with the same key.
         """
+        if not color:
+            color = self._current_state.text.get_colors()
+
+        self._expand_colors(color, len(text))
```
bosh, line 163: I believe this should be assigned back into color=, because otherwise either we're intending for the inbound array to be mutated, or we're not using the return.

But if you add this setting, then a bunch of tests fail. (contd in comments on tests)

mpf/devices/segment_display/segment_display.py
```diff
@@ -86,8 +90,16 @@ async def _initialize(self):
 
         self.size = self.config['size']
         self._default_color = [RGBColor(color) for color in self.config["default_color"][0:self.size]]
-        if len(self._default_color) < self.size:
-            self._default_color += [RGBColor("white")] * (self.size - len(self._default_color))
+
+        if (len(self._default_color)) == 1:
```
bosh, line 94: I think we need to follow up with naming convention consistency fixes, because this intends to be a list length check but it reads like a string length check, but it actually _is_ a list.

mpf/devices/segment_display/segment_display_text.py
```diff
@@ -75,29 +75,35 @@ def _create_characters(cls, text: str, display_size: int, collapse_dots: bool, c
...
-            char_list.append(DisplayCharacter(char_code, char_has_dot, char_has_comma, color))
 
-        # ensure list is the same size as the segment display (cut off on left or right justify characters)
-        current_length = len(char_list)
+        # Adjust the color array if needed
+        if (len(colors)) == 1:
+            colors = colors * display_size
+        elif len(colors) != display_size:
+            #TODO: Log that colors were adjusted to white as default
+            colors = [RGBColor("white")] * display_size
```
bosh, line 91: I think we can use defaults or padding rules here (e.g. last known color expands to fit the rest of the space) rather than having to bail to full white. Though I guess maybe going to white at least guarantees on and could be a clear indicator of problem (if we expect length mismatches should never actually happen).

mpf/tests/test_SegmentDisplay.py
```diff
@@ -476,8 +476,8 @@ def test_segment_display_text(self):
         self.assertEqual(5, len(test_text))
         colors = test_text.get_colors()
         self.assertEqual(5, len(colors))
-        self.assertEqual([RGBColor("white"), RGBColor("red"), RGBColor("red"),
-                          RGBColor("red"), RGBColor("red")], colors)
+        self.assertEqual([RGBColor("white"), RGBColor("white"), RGBColor("white"),
```
bosh, line 479: I'm unsure about these test changes. Given no test setup differences, it seems that we're seeing formerly-working reds be replaced with fallback/default white. Seeing five tests all lose color information and go to default white is concerning, is this correct?

bosh (changes requested):
I pulled down this branch and did some cleanup (squashing + removing the default set and removal commits that netted no diff) and I think it's really close, but there are several issues that make me think there's more to do still.

I also changed the segment display `update_method` option to an enum to better document and enforce the choices. I'm willing to continue doing a bit more from my branch as a starting point, but I don't think I can really prove anything until I have a real rgb segment display to test on (and I might soon!) (or at least a virtual config that replicates a rgb segment unit)

worldpeace-germany:
> I pulled down this branch and did some cleanup (squashing + removing the default set and removal commits that netted no diff) and I think it's really close, but there are several issues that make me think there's more to do still.
> 
> I also changed the segment display `update_method` option to an enum to better document and enforce the choices. I'm willing to continue doing a bit more from my branch as a starting point, but I don't think I can really prove anything until I have a real rgb segment display to test on (and I might soon!) (or at least a virtual config that replicates a rgb segment unit)

I don't have a RGB display either, at what model for an RGB-7-segment display are you looking at? I tried to find some using Neopixel that I can add it to my Cobra board. I didn't find much, for example https://www.crowdsupply.com/maksmakes/neosegment but that would have full color for each segment, I believe in the mpf code the RGB refers only that each digit can have any color, but not each segment different colors. So it can get more complicated...

Of course I cannot prove working RGB segments since I don't have any. If you get one, that would be good. Now comes the but, I only started the whole implementation because I was not able to get it to work at all with the original code (using stack, since replace wasn't implemented at all). Are you getting your segment display to work with the original code? If yes, then I have missed something back then. If not, the question is if this PR should be accepted to have a start that at least non-RGB displays are working, knowing that we can only hope that RGB works as well. A step into the right direction, but not yet crossed the finish line. If you know someone who has an RGB display then I am happy to work with that person to do some test and fix potential RGB issues. If you know an RGB 7 segment display I can use with Cobra board I would even consider to buy one (depending on price obviously) to test it myself.

PR #1828, opened 2024-08-25
Add shot block test cases

bosh:
I've been learning about shot blocking and I realized through testing that the MPF docs are incorrect (on https://github.com/missionpinball/mpf-docs/blob/main/docs/config/shot_profiles.md ) about the default behavior of Block.

In order to prove that the default behavior is as block: False instead of True, I've added test cases for the False and True options. (config_spec:1617 for shot_profiles:block shows the default to be false as well)


bosh:
It looks like the three failures in linting are preexisting on dev

avanwinkle (approved):
Thanks for the tests!

PR #1837, opened 2024-09-20
FAST Neuron should ignore switches on other platforms

bosh:
If using multiple platforms for switches, the Neuron communicator will try to look up every switch as though they are connected to FAST hardware, causing a crash. This is a bad assumption and it keeps a machine creator from declaring virtual switches when using a Neuron.

Virtual switches can be useful in letting a machine developer add switches to configs before they're wired into the machine, or even by creating virtual-only switches that will _never_ be physically implemented, but can be used in mpf-monitor for debugging and arbitrary event creation.

First commit changes Neuron platform tests to be explicit about which platform they're using (instead of defaulting) but should be green. (Action run: https://github.com/bosh/mpf/actions/runs/10966582957 )
Second commit changes the test to include a cross-platform switch that causes crashes (9x crash in that test file due to Neuron not filtering switch platform) (Action run: https://github.com/bosh/mpf/actions/runs/10966584585 )
Third commit fixes the Neuron communicator code. I'm not sure of the best syntax for this in Python (could we filter the .values() call?) but it works :)

### DIFF
--- a/mpf/platforms/fast/communicators/net_neuron.py
+++ b/mpf/platforms/fast/communicators/net_neuron.py
@@ -293,7 +293,7 @@ class FastNetNeuronCommunicator(FastSerialCommunicator):
         This will silently sync the switch.hw_state. If the logical state changes,
         it will process it like any switch change.
         """
-        for switch in self.machine.switches.values():
+        for switch in [sw for sw in self.machine.switches.values() if sw.platform == self.platform]:
             hw_state = self.platform.hw_switch_data[switch.hw_switch.number]
 
             if hw_state != switch.hw_state:
--- a/mpf/tests/machine_files/fast/config/neuron.yaml
+++ b/mpf/tests/machine_files/fast/config/neuron.yaml
@@ -34,6 +34,10 @@ fast:
                 order: 4
 
 switches:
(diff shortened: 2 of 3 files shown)


mpf/platforms/fast/communicators/net_neuron.py
```diff
@@ -294,6 +294,9 @@ def update_switches_from_hw_data(self):
         it will process it like any switch change.
         """
         for switch in self.machine.switches.values():
+            if switch.platform != self.platform:
```
bosh: I expect there's a better way in python to filter, and then not even need the continue

avanwinkle: You should be able to do it like this:
```
for switch in self.machine.switches.values() if switch.platform == self.platform:
```

bosh: Took a few whacks at it and found the final syntax. It's a little awkward imo, but it's in :)

bosh:
PR is all ready for final review! I've also tested the current version (with the list comprehension) on my neuron, no problems.

avanwinkle (approved):
Great fix, thanks!

Issue #1839, opened 2024-09-22
Issue with segment display update method: replace


mpcarr:
When using the update method: replace, if you do not specfic a color or a flash properties in segment_display_player then the text is not shown on the display.

PR #1840, opened 2024-09-22
Segment display replace

mpcarr:
This PR addresses issue #1839. The newly added replace feature ([https://github.com/missionpinball/mpf/pull/1779](Segment Displays: Stack vs Replace Option).

This update using the current states color, flashing and flash_mask settings for the incoming text if non are specified on the segment_display_player. Without this, the text isn't displayed due the the color being missing and the text is set to flash.

Additionally this update removes the text from the display when the update method is not stack.

### DIFF
--- a/mpf/devices/segment_display/segment_display.py
+++ b/mpf/devices/segment_display/segment_display.py
@@ -161,7 +161,7 @@ class SegmentDisplay(SystemWideDevice):
             flash_mask = self._current_state.flash_mask
         if not color:
             color = self._current_state.text.get_colors()
-        
+
         new_text = TextTemplate(self.machine, text).evaluate({})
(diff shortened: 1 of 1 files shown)


avanwinkle:
Thanks for the input, have you looked at https://github.com/missionpinball/mpf/pull/1820 to see where your respective changes align/diverge?

worldpeace-germany:
@mpcarr the PR @avanwinkle has referenced is from me. I believe my fix should cover your case and even some more. I have the impression that segment displays are not that widely used and it is hard to get test cases validated. Would you mind to test my code changes in your project and report back? If you find issues I am willing (happily with you together) to fix these to get a better segment display support into mpf.

PR #1862, opened 2025-01-09
Add restart grace period to multiball

bernarma:
Currently Stern pinball machines allow a grace period to restart a multiball for a mode that is in-progress. There are work-arounds to get this to function within the configuration but is not elegant nor very reusable. I propose this change to allow multiballs to be restarted during a grace period via add-a-ball. This grace period by default is set to 0 therefore existing configurations will continue to function as normal behavior.

### DIFF
--- a/mpf/devices/multiball.py
+++ b/mpf/devices/multiball.py
@@ -239,20 +239,20 @@ class Multiball(EnableDisableMixin, SystemWideDevice, ModeDevice):
 
         restart_grace_period_ms = self.config['restart_grace_period'].evaluate([])
 
-        if not self.machine.game or (self.machine.game.balls_in_play - balls < 1 and restart_grace_period_ms==0):
+        if not self.machine.game or (self.machine.game.balls_in_play - balls < 1 and restart_grace_period_ms == 0):
             self._multiball_end()
-        elif restart_grace_period_ms>0:
+        elif restart_grace_period_ms > 0:
             if self.machine.game.balls_in_play == 0:
-                self.delay.run_now(f"{self.name}_restart_grace_period")
+                self.delay.run_now("restart_grace_period")
             elif self.machine.game.balls_in_play - balls < 1:
-                self.machine.events.post(f"multiball_{self.name}_will_end", duration=restart_grace_period_ms)
+                self.machine.events.post(f"multiball_{self.name}_will_end", grace_period=restart_grace_period_ms)
                 '''event: multiball_(name)_will_end
                 desc: The multiball called (name) does not have multiple balls active, enabling grace period.
                 args:
-                    duration: duration of the restart grace period
+                    grace_period: restart duration grace period in ms
                 '''
 
-                self.delay.add(name=f"{self.name}_restart_grace_period",
+                self.delay.add(name="restart_grace_period",
                                ms=restart_grace_period_ms,
                                callback=self._multiball_end)
 
@@ -265,7 +265,6 @@ class Multiball(EnableDisableMixin, SystemWideDevice, ModeDevice):
         desc: The multiball called (name) has just ended.
         '''
         self.debug_log("Ball drained. MB ended.")
-
 
     @event_handler(5)
     def event_stop(self, **kwargs):
@@ -304,10 +303,9 @@ class Multiball(EnableDisableMixin, SystemWideDevice, ModeDevice):
 
     def add_a_ball(self):
         """Add a ball if multiball has started."""
-
         # remove if it exists as multiball now has multiple balls
-        if self.delay.check(f"{self.name}_restart_grace_period"):
-            self.delay.remove(f"{self.name}_restart_grace_period")
+        if self.delay.check("restart_grace_period"):
+            self.delay.remove("restart_grace_period")
             self.machine.events.post(f"multiball_{self.name}_will_resume")
             '''event: multiball_(name)_will_resume
             desc: Multiball (name) has been resumed during end of multiball grace period.
--- a/mpf/devices/multiball.py
+++ b/mpf/devices/multiball.py
@@ -237,15 +237,33 @@ class Multiball(EnableDisableMixin, SystemWideDevice, ModeDevice):
         desc: The multiball called (name) has lost a ball after ball save expired.
         '''
 
-        if not self.machine.game or self.machine.game.balls_in_play - balls < 1:
-            self.balls_added_live = 0
-            self.balls_live_target = 0
-            self.machine.events.remove_handler(self._ball_drain_count_balls)
-            self.machine.events.post("multiball_{}_ended".format(self.name))
-            '''event: multiball_(name)_ended
-            desc: The multiball called (name) has just ended.
-            '''
-            self.debug_log("Ball drained. MB ended.")
+        restart_grace_period_ms = self.config['restart_grace_period'].evaluate([])
+
+        if not self.machine.game or (self.machine.game.balls_in_play - balls < 1 and restart_grace_period_ms==0):
+            self._multiball_end()
+        elif restart_grace_period_ms>0:
+            if self.machine.game.balls_in_play == 0:
+                self.delay.run_now(f"multiball_{self.name}_grace_period_restart")
+            elif self.machine.game.balls_in_play - balls < 1:
+                self.machine.events.post("multiball_{}_will_end".format(self.name))
+                '''event: multiball_(name)_will_end
+                desc: The multiball called (name) does not have multiple balls active, enabling grace period.
+                '''
+
+                self.delay.add(name=f"multiball_{self.name}_grace_period_restart",
+                               ms=restart_grace_period_ms,
+                               callback=self._multiball_end)
+
+    def _multiball_end(self):
+        self.balls_added_live = 0
+        self.balls_live_target = 0
+        self.machine.events.remove_handler(self._ball_drain_count_balls)
+        self.machine.events.post("multiball_{}_ended".format(self.name))
+        '''event: multiball_(name)_ended
+        desc: The multiball called (name) has just ended.
+        '''
+        self.debug_log("Ball drained. MB ended.")
+
 
     @event_handler(5)
     def event_stop(self, **kwargs):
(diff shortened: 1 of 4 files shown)


mpf/devices/multiball.py
```diff
@@ -237,15 +237,33 @@ def _ball_drain_count_balls(self, balls, **kwargs):
...
-            desc: The multiball called (name) has just ended.
-            '''
-            self.debug_log("Ball drained. MB ended.")
+        restart_grace_period_ms = self.config['restart_grace_period'].evaluate([])
+
+        if not self.machine.game or (self.machine.game.balls_in_play - balls < 1 and restart_grace_period_ms==0):
+            self._multiball_end()
+        elif restart_grace_period_ms>0:
+            if self.machine.game.balls_in_play == 0:
+                self.delay.run_now(f"multiball_{self.name}_grace_period_restart")
```
avanwinkle: Clever use of `run_now` to cancel the delay and execute the ending!

mpf/devices/multiball.py
```diff
@@ -284,6 +302,15 @@ def event_add_a_ball(self, **kwargs):
 
     def add_a_ball(self):
         """Add a ball if multiball has started."""
+
+        # remove if it exists as multiball now has multiple balls
+        if self.delay.check(f"multiball_{self.name}_grace_period_restart"):
```
avanwinkle: I feel like this name implies that it's _restarting_ the grace period, rather than _being_ a grace period _for_ a restart. For consistency, this name should be `restart_grace_period`—that's still a bit ambiguous, but I can't think of anything better :) 

Because the delay's scope is encapsulated to `self.delay`, there's no need for `multiball_<name>_` in the delay name.

mpf/devices/multiball.py
```diff
@@ -284,6 +302,15 @@ def event_add_a_ball(self, **kwargs):
...
+            self.delay.remove(f"multiball_{self.name}_grace_period_restart")
+            self.machine.events.post("multiball_{}_will_resume".format(self.name))
```
avanwinkle: Might as well use f-string formatting here

mpf/devices/multiball.py
```diff
@@ -237,15 +237,33 @@ def _ball_drain_count_balls(self, balls, **kwargs):
...
-            self.debug_log("Ball drained. MB ended.")
+        restart_grace_period_ms = self.config['restart_grace_period'].evaluate([])
+
+        if not self.machine.game or (self.machine.game.balls_in_play - balls < 1 and restart_grace_period_ms==0):
+            self._multiball_end()
+        elif restart_grace_period_ms>0:
+            if self.machine.game.balls_in_play == 0:
+                self.delay.run_now(f"multiball_{self.name}_grace_period_restart")
+            elif self.machine.game.balls_in_play - balls < 1:
+                self.machine.events.post("multiball_{}_will_end".format(self.name))
```
avanwinkle: I've never encountered/noticed this feature before, so I'm not sure how its used or communicated (if at all). Would there be value in including the grace period duration as an event kwarg? Does it matter to anyone?

avanwinkle (commented):
Looks good! I tried my best to find gaps or gotchas that might lead to breaking situations or confusion, but I couldn't find any!

Could you please add some tests to validate the behavior of successful restarting and grace timeout?

mpf/devices/multiball.py
bernarma: I "borrowed" it from bonus :-)

mpf/devices/multiball.py
bernarma: I'll add it in. It could be useful if the grace period is dynamically set.

mpf/devices/multiball.py
bernarma: See latest PR

mpf/devices/multiball.py
bernarma: See latest PR

mpf/devices/multiball.py
bernarma: See latest PR

mpf/devices/multiball.py
```diff
@@ -237,15 +237,35 @@ def _ball_drain_count_balls(self, balls, **kwargs):
...
-            self.balls_live_target = 0
-            self.machine.events.remove_handler(self._ball_drain_count_balls)
-            self.machine.events.post("multiball_{}_ended".format(self.name))
-            '''event: multiball_(name)_ended
-            desc: The multiball called (name) has just ended.
-            '''
-            self.debug_log("Ball drained. MB ended.")
+        restart_grace_period_ms = self.config['restart_grace_period'].evaluate([])
+
+        if not self.machine.game or (self.machine.game.balls_in_play - balls < 1 and restart_grace_period_ms==0):
```
avanwinkle: The linter is going to complain here that you need spaces between values and comparators, i.e. `restart_grace_period_ms == 0` and two lines down, `restart_grace_period_ms > 0`

mpf/devices/multiball.py
```diff
@@ -237,15 +237,35 @@ def _ball_drain_count_balls(self, balls, **kwargs):
...
+            self._multiball_end()
+        elif restart_grace_period_ms>0:
+            if self.machine.game.balls_in_play == 0:
+                self.delay.run_now(f"{self.name}_restart_grace_period")
```
avanwinkle: Please remove `{self.name}` from the delay name, it's a computational cost to generate the formatted string and unnecessary because `self.delay` is isolated in scope to just this particular instance of multiball.

mpf/tests/test_MultiBall.py
```diff
@@ -1228,3 +1228,52 @@ def testAddABallSaverDuringShootAgain(self):
...
+        self.assertAvailableBallsOnPlayfield(2)
+
+        self.drain_one_ball()
+        self.advance_time_and_run(5)
+
+        self.assertEventCalled("multiball_mb_grace_period_restart_will_end", 1)
+        self.assertEventCalled("multiball_mb_grace_period_restart_ended", 0)
+
+        self.advance_time_and_run(5)
+        self.assertEventCalled("multiball_mb_grace_period_restart_ended", 1)
```
avanwinkle, line 1250: For thoroughness, this test should validate the number of balls on the playfield (`assertAvailableBallsOnPlayfield(1)`) and that the _multiball_mb_grace_period_restart_will_resume_ event did *not* get called.

mpf/tests/test_MultiBall.py
```diff
@@ -1228,3 +1228,52 @@ def testAddABallSaverDuringShootAgain(self):
...
+        # start mb, default ball save, 5s restart grace period
+        self.post_event("mb_grace_period_restart_start")
+        self.advance_time_and_run(10)
+        self.assertAvailableBallsOnPlayfield(2)
+
+        self.drain_one_ball()
+        self.advance_time_and_run(5)
+
+        self.assertEventCalled("multiball_mb_grace_period_restart_will_end", 1)
+        self.assertEventCalled("multiball_mb_grace_period_restart_ended", 0)
```
avanwinkle: For readability, prefer `assertEventNotCalled()` instead of `assertEventCalled` zero times. Just easier to scan the code and understand what's happening :)

mpf/tests/test_MultiBall.py
```diff
@@ -1228,3 +1228,52 @@ def testAddABallSaverDuringShootAgain(self):
...
+        self.assertAvailableBallsOnPlayfield(2)
+
+        self.drain_one_ball()
+        self.advance_time_and_run(3)
+
+        self.assertEventCalled("multiball_mb_grace_period_restart_will_end", 1)
+        self.assertEventCalled("multiball_mb_grace_period_restart_ended", 0)
+
+        self.post_event("add_ball")
+        self.assertEventCalled("multiball_mb_grace_period_restart_ended", 0)
```
avanwinkle: Best to add a single-cycle clock loop after `post_event` just to make sure every handler has processed. That can be run by passing no time argument: `self.advance_time_and_run()`

mpf/tests/test_MultiBall.py
```diff
@@ -1228,3 +1228,52 @@ def testAddABallSaverDuringShootAgain(self):
...
+        self.assertEventCalled("multiball_mb_grace_period_restart_will_resume", 1)
+        self.advance_time_and_run(5)
+        self.assertAvailableBallsOnPlayfield(2)
+        self.assertEventCalled("ball_save_mb_grace_period_restart_timer_start")
```
avanwinkle, line 1284: And for thoroughness, best to test that the multiball ending events did _not_ post here

mpf/devices/multiball.py
```diff
@@ -237,15 +237,35 @@ def _ball_drain_count_balls(self, balls, **kwargs):
...
+            elif self.machine.game.balls_in_play - balls < 1:
+                self.machine.events.post(f"multiball_{self.name}_will_end", duration=restart_grace_period_ms)
```
avanwinkle: For clarity, I think it'd be best to name the kwarg "grace_period" instead of "duration", so there's no confusion about what that value applies to.

In this setup, the _multiball_<name>_will_end_ event is only posted if there is a grace period, otherwise it gets skipped and just the _multiball_<name>_ended_ event is sent. That seems fine to me, a designer will know whether or not a grace period exists and can manage event handlers accordingly, but I just wanted to check with you that that's your intention.

avanwinkle (commented):
This is looking great! A few tweaks and if you are setup to run testing and prospector locally, that'll be a good safety check.

`python -m unittest discover mpf/tests` (from root)
`prospector` (also from root, after `pip install prospector`)

If not that's okay, we can trigger the CI to run the tests after addressing my few little comments. Thanks for this nice improvement!

bernarma:
Happy to do it. I don't mind making the changes. It's going into a code base used by many. I haven't used the tools but will run it and get the changes made as required.

mpf/devices/multiball.py
bernarma: I ran the linter... Never used so added to the toolbelt for the future.

bernarma:
PR updated.

avanwinkle (dismissed):
Fantastic work, thanks!

avanwinkle:
Oh, I just saw that this is targeting the 0.80 branch. I tried targeting at `dev` so the changes will propagate to 0.57 as well as 0.80, for feature parity between the two, but that pulled in all the deltas between 0.80 and 0.57. Could you rebase this PR against `dev`?

bosh:
Replacing with dev-based PR, all four commits cherry-picked without conflicts https://github.com/missionpinball/mpf/pull/1969

PR #1865, opened 2025-01-20
Update VPE dependencies and rebuild interface

### LINKED ISSUE #1864: Pip installation of VPE dependencies fails on Python >3.9
Running `python3 -m pip install mpf[vpe] --pre` with any Python version newer than 3.9 (I tried 3.11 and 3.12) fails with a *very* long error log that I cannot upload here (probably some recursive madness). This is the end: 

```
Failed to build grpcio grpcio_tools
ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (grpcio, grpcio_tools)`
```

Reproduced in Ubuntu 2024 and Windows 11.

arthurkehrwald:
I installed the latest versions of grpcio, grpciotools and protobuf, regenerated the platform interface scripts using `mpf/platforms/visual_pinball_engine/generate.py` and tested it with VPE.

Fixes #1864


avanwinkle (approved):
We're generally reticent to roll package upgrades because they can cause downstream issues, but all the tests passed and we gotta do it sometime, right?

Thanks for taking care of this!

arthurkehrwald:
I've added a new RPC to the service definition. It is a simple ping function that I will use in the VPE integration to check if MPF is ready. Previously, VPE simply waited for 1,5 seconds and assumed it was ready, but this can fail on slow computers and wastes time on fast ones. GRPC provides ways to wait for the connection to be ready that would be preferrable, but they are not supported on grpc-dotnet with .netstandard2.1.

arthurkehrwald:
I've added the MPF version to the ping response so VPE can accommodate different versions of MPF in the future.

PR #1870, opened 2025-02-04
chore: address prospector for segment_mappings

unRARed:
Seems like builds are failing due to this file having style issues.


avanwinkle (approved):
Thanks so much!!

PR #1871, opened 2025-02-04
add pin2dmd HD support

4r3:
Hello

here are some changes to add Pin2DMD HD ( 256x64) support


avanwinkle (approved):
I have no way to test this, but since you made it I assume you're happy with how it works. Thanks for the update!

PR #1872, opened 2025-02-08
shh pylint, we need those locals

bosh:
Didn't seem like there's a good seam or reason to refactor to reduce locals in this instance, so I added a disable


avanwinkle (approved):
Great, thanks!

Issue #1874, opened 2025-02-09
Coil Test crash on .80 with dual_wound_coils


bosh:
I'll try my hand at a fix, just putting the issue out for organization first. Maybe dual wound can be skipped during coil test because the component coils will each be tested instead? Or do whatever dual wound flippers do during the coil test, if that has special handling.

I get the following crash when running the coil test in the 0.80.x service menu in the Godot player window.

```
01:20:26.276 : INFO [EventManager] Event: ======'service_button'====== Args={'button': 'ENTER'}
01:20:26.302 : INFO [EventManager] Event: ======'service_trigger'====== Args={'action': 'coil_test', 'sort': 'false', '_from_bcp': True}
01:20:26.303 : ERROR [asyncio] Exception in callback AsyncMode._mode_ended(<Task finishe...'hw_driver'")>)
handle: <Handle AsyncMode._mode_ended(<Task finishe...'hw_driver'")>)>
Traceback (most recent call last):
  File "C:\Users\Bosh\AppData\Local\Programs\Python\Python312\Lib\asyncio\events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "C:\Users\Bosh\workspace\mpf\mpf\core\async_mode.py", line 48, in _mode_ended
    future.result()
  File "C:\Users\Bosh\workspace\mpf\mpf\modes\service\code\service.py", line 105, in _run
    await self._start_main_menu()
  File "C:\Users\Bosh\workspace\mpf\mpf\modes\service\code\service.py", line 150, in _start_main_menu
    await self._service_mode_main_menu()
  File "C:\Users\Bosh\workspace\mpf\mpf\modes\service\code\service.py", line 157, in _service_mode_main_menu
    key = await self._get_key()
          ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Bosh\workspace\mpf\mpf\modes\service\code\service.py", line 94, in _get_key
    await self._service_trigger()
  File "C:\Users\Bosh\workspace\mpf\mpf\modes\service\code\service.py", line 762, in _service_trigger
    await self._coil_test_menu()
  File "C:\Users\Bosh\workspace\mpf\mpf\modes\service\code\service.py", line 285, in _coil_test_menu
    items = self.machine.service.get_coil_map(do_sort=self._do_sort)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1876, opened 2025-02-09
skip coils without hardware drivers when generating coil info for service menu

bosh:
https://github.com/missionpinball/mpf/issues/1874 resolution - skip coils without hw_drivers (such as dual wound coils) for service controller use cases

### DIFF
--- a/mpf/core/service_controller.py
+++ b/mpf/core/service_controller.py
@@ -9,6 +9,7 @@ from collections import namedtuple
 from typing import List
 
 from mpf.core.mpf_controller import MpfController
+from mpf.devices.dual_wound_coil import DualWoundCoil
 
 SwitchMap = namedtuple("SwitchMap", ["board", "switch"])
 CoilMap = namedtuple("CoilMap", ["board", "coil"])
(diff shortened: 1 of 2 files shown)


avanwinkle:
Interesting! I honestly had no idea that dual wound coils existed in MPF, but glad to know!

Service mode definitely shouldn't crash on them, and should know what they are. Something like
```
for coil in self.machine.coils.values():
  if coil is DualWoundCoil:
    continue
  ...
```

So that takes care of the dual wound coils. As for other coils, that assert statement is pretty old and I'm not wild about crashing the game unnecessarily, so maybe replacing the assert with an `if` statement and logging a warning that the coil has no hardware driver and won't be included. That's a nicer approach that won't break the game but still informs the user of the issue. Putting that line after the `DualWoundCoil` check will acknowledge that dual wound coils are expected and don't need a warning.

Thanks for working on this!

bosh:
This is ready to go, if tests are green. I added to the service test config file -- the new device declaration would have caused testServiceCli and testServiceMode to crash without the fix, but with the fix there's nothing new to see or assert on.

I went with the class check, but TBH I slightly prefer the has_attr("hw_driver") variant I had written previously. Upside of not using the class check is one less import, and the place I cribbed from suggested there might be other objects in `machine.coils` that also don't have a hw_driver. ([Driver.py:53](https://github.com/missionpinball/mpf/blob/f9c7cab113e9e368d16a92990508acdba99d778a/mpf/devices/driver.py#L53))

avanwinkle (approved):
Fantastic!

bosh:
Lint is fixed by https://github.com/missionpinball/mpf/pull/1881

PR #1877, opened 2025-02-11
simplify placeholder manager ast integration

bosh:
Used to be conditionally-included, but is now always included just a few lines up.


avanwinkle (approved):
Nice!

PR #1879, opened 2025-02-18
Boolean setting support

bosh:
Add support to for boolean settings. There's no real reason why you couldnt use a more complex type for a boolean setting, and you would be more prepared if you have to add a third (or more) option in the future -- HOWEVER! I think bool settings are still useful because you should be able to embed them in other templates and logic checks more simply than the other settings types.

Tests include a backfill for float support, and both the cases of true and false defaults (just in case the yaml-to-python conversion could have an issue with e.g. any non-empty-string converting as truthy.

I still need to test the boolean setting in the settings management interface, and also embedded in other templates, but I'm posting now because it's close enough to be worth comment already.

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -1414,7 +1414,7 @@ settings:
     label: single|str|
     sort: single|int|
     values: dict|str:str|
-    key_type: single|enum(str,float,int)|str
+    key_type: single|enum(str,float,int,bool)|str
     default: single|str|
     machine_var: single|str|None
     settingType: single|str|standard
--- a/mpf/core/utility_functions.py
+++ b/mpf/core/utility_functions.py
@@ -57,6 +57,8 @@ class Util:
             return float(value)
         if type_name == "str":
             return str(value)
+        if type_name == "bool":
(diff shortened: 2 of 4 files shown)


avanwinkle:
This seems like a nice improvement. The only reason I know of why booleans weren't previously supported is that they are literally indifferent from 0 and 1, so there's nothing a boolean could do that couldn't be done with an int. But also, no reason not to have booleans.

I'm a little concerned that the other values will create some "magic" behavior. I realize that this is part of MPF canon, allowing "yes" and "t" to stand for true, et cetera. Something about it just rubs me wrong, but I'm not strongly opinionated enough to push for removing it.

bosh:
Yeah I agree that the extra "true" parsing values are awkward. It's through this that I became aware of my expectations of YAML parsing rules are from Ruby in particular, which is subtly different than Python's standard. I do think it's weird that True vs true is okay in many places, but config fields that expect template_booleans specifically require True (true has an explicit check and raise in placeholder_manager.py#_eval_name (line 756).

I still haven't gotten around to testing the boolean setting by hand to make sure the service menu etc work fine with it ... maybe this weekend.

bosh:
I finally got around to testing it, was fine all this time!

PR #1880, opened 2025-02-19
fix: remove line breaking mypinballs segment displays

unRARed:
This might be another naive PR from our side, but we have been trying to get our "mypinballs" segment displays going and have been running into a strange issue. The displays were initializing as expected, but any value we entered only ever displayed `"      0"` and was the case for all 4 displays we're running. I reached out to the guy who wrote the MyPinballs firmware and he provided instructions for the formats the displays expect which is as such:

`Control Code`  :  `Display ID`  :  `Integer Value to Display`

So to display a score of 500 on the 2nd display, the value is `1:2:500`. We were able to confirm our displays are working by simply connecting from the terminal as such: `echo "1:1:1234567\n" > /dev/ttyACM1`.

With this confirmed, we looked at the MyPinballs mpf source and noticed this curious piece:

```python
# special char for spaces
text = text.replace(" ", "?")
```

As a hunch, I removed the line... and instantly our attract values displayed. Further, starting a game reinitialized the displays and switch hits were scoring. I'm wondering if there are perhaps different versions of the MyPinballs segment display controller or maybe other updates downstream broke the existing implementation? Anyone else running these displays that could test against the `0.57.x` branch and report back?

### DIFF
--- a/mpf/platforms/mypinballs/mypinballs.py
+++ b/mpf/platforms/mypinballs/mypinballs.py
@@ -27,8 +27,6 @@ class MyPinballsSegmentDisplay(SegmentDisplayPlatformInterface):
             # remove any non-numbers and spaces
             text = re.sub(r'[^0-9 ]', "", text)
 
-            # special char for spaces
-            text = text.replace(" ", "?")
             # set text
             if flashing == FlashingType.FLASH_ALL:
                 cmd = b'2:'


avanwinkle (approved):
Thanks for figuring this out!

avanwinkle (changes requested):
Change seems simple enough but some tests have broken as a result, hopefully it's straightforward for you to resolve!

unRARed:
Sorry, this wasn't ready to be merged. In the PR description, I mentioned:

> I'm wondering if there are perhaps different versions of the MyPinballs segment display controller or maybe other updates downstream broke the existing implementation? Anyone else running these displays that could test against the 0.57.x branch and report back?

So I suspect the sketch that's on the Arduino that MyPinballs ships changed at some point. I think @jabdoa2 wrote that code originally. Is he still contributing? Does anyone else have this hardware? I could delete those tests for the line we removed, but that might break the integration with the older versions (again, assuming it changed at some point).

jabdoa2:
I think I wrote this based on some spec. Not sure if I actually have the hardware. If this does not work on real hardware with recent firmware just change it.

bosh:
I've fixed the tests and rebased here: https://github.com/missionpinball/mpf/pull/1973

mypinballs:
I found this from a mpf thread about using alphanumerics and this driver randomly, i'm not sure why i wasn't asked for more detail or maybe i missed a message somewhere. This stuff was written 10 years ago nearly..  Anyway, The ? character for blanks was introduced for compatibility with updates on the hardware firmware. Using spaces was not a good idea and caused some bugs iirc. Because the Bally segment display control ics only handle 0-9. a blank is anything after 10 so i chose 15 as the blank which when you subtract 48 of the asciii codes (for numerical use) you get a '? ' char 

Character 0 is ascii 48 , therefore 48-48 =0 for hardware use
Character ? is ascii 63, therefore 63-48 = 15

The above code should be re-instated and you should use the latest firmware from me :) Or an option for use of both depending on what version of hardware you have. As said though earlier versions will do weird stuff with space use sometimes.

bosh:
Gotcha, so MPF's best choice is to leave the ' ' -> '?' replacement in place, and recommend a firmware upgrade for users having issues? Thanks!

mypinballs:
> Gotcha, so MPF's best choice is to leave the ' ' -> '?' replacement in place, and recommend a firmware upgrade for users having issues? Thanks!

Yep :)

PR #1882, opened 2025-02-27
Dedup 0.80.x-only config error codes

bosh:
I'm doing a sweep of error code fixes and copy tweaks for dev, and I noticed this code being used in 0.80.x with a duplicate number with another error in the file. The rest of the fixes are on another branch to merge separately.

I went up to 20 for this code because there are about a dozen codes already in use in the file, with many different cases actually grouping under case No. 5. I might add individual number codes for each of those grouped errors in the future.


avanwinkle (approved):

bosh:
Rebased to show green lint

avanwinkle (approved):

PR #1883, opened 2025-03-02
Adding support to PPUC ZeDMD device

JerOrc:
From the C++ library libzedmd of PPUC  https://github.com/PPUC/libzedmd , 
I write a pybind11 extension to link C++ with Python here : https://github.com/PPUC/libzedmd-python-pybind11-extension 
and integrate this work into MPF

Numpy is needed
Tested on Linux with this MPF-Godot project : https://github.com/PPUC/MPF-pinball-ZeDMD-demo


mpf/config_spec.yaml
```diff
@@ -2243,3 +2243,15 @@ window:
     top: single|int|None
     no_window: single|bool|false
     effects: ignore
+    
+zedmd:
+    __valid_in__: machine
+    __type__: config_dict
+    old_cookie: single|bool|false
```
bosh: I don't see this used anywhere, is `old_cookie` necessary? It looks like in MPF code it's something used only in SmartMatrix config and behavior.

JerOrc: You're right, I don't use it, how can I change this ? I have to make a commit and ask another pull request ?

JerOrc: I'm a newbie on Github sorry :-)

bosh: You can either amend the original commit on your machine and force push the branch on your fork- this PR will track the head of the branch and update fine -- or you can add another commit at the end of the branch that makes this change and any other tweaks Anthony might suggest.

bosh:
You should consider editing the PR target branch to be `missionpinball:dev` instead of 0.80.x, as it doesn't seem like you need anything from the GMC/Godot upgrades and could make this available for 0.57 as well. I think you have to put the PR back into draft mode to change the merge target branch, and then you can leave draft mode again after the change.

mpf/config_spec.yaml
JerOrc: old_cookie removed

JerOrc:
As asked by bosh I want to change the source branch of this PR but with Github it seems not to be possible so I close this PR and open another

PR #1885, opened 2025-03-20
Update Tilt slide behaviors for GMC

bosh:
GMC moved slide definition from yaml to Godot tscns in large part, removing the slides: config value from MPF. Slide expiration handling thus had to move into the slide_player configurations.

This PR is accompanied by https://github.com/missionpinball/mpf-gmc/pull/28 and documented by https://github.com/missionpinball/mpf-docs/pull/585

### DIFF
--- a/mpf/modes/tilt/config/tilt.yaml
+++ b/mpf/modes/tilt/config/tilt.yaml
@@ -19,9 +19,22 @@ tilt:
   tilt_warnings_player_var: tilt_warnings
 
 slide_player:
-  tilt_warning_1: tilt_warning_1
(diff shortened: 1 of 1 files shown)


avanwinkle (commented):
As noted on the GMC PR, consolidating into one slide and using event args would reduce your just one slide play call and one remove call.

bosh:
The removes are mostly unnecessary, I put them in as a safety -- the warnings auto expire in 1s so should be gone before a clear would be expected anyway.

avanwinkle (approved):

PR #1888, opened 2025-04-13
stop "shots" control handlers removal when the shot is destroyed

gbouthenot:
Currently, in the shots config, "control handlers" are mixed with "switch handlers". When the mode ends, all handlers are unregistered (both switch and control handlers).
But only the switch handlers should, because when the mode is activated again, only the switch handlers are registered back. The Control handlers are not.

This PR does register the control handlers without adding them to the switch handlers.

### DIFF
--- a/mpf/devices/shot.py
+++ b/mpf/devices/shot.py
@@ -425,8 +425,8 @@ class Shot(EnableDisableMixin, ModeDevice):
     def _register_control_event_handlers(self):
         for control_event in self.config['control_events']:
             for event in control_event['events']:
-                self._handlers.append(self.machine.events.add_handler(event, self._control_events,
-                                                                      control_event_config=control_event))
+                self.machine.events.add_handler(event, self._control_events,
+                                                control_event_config=control_event)
 
     @event_handler(7)
     def _control_events(self, control_event_config, **kwargs):


avanwinkle:
This is interesting! Following the code path, I see what you're saying. What baffles me is that this is how the code has been for a long time, but nobody has encountered this issue.

Can you describe your setup and the steps to reproduce the error? A test case to validate the fix would also be great.

I'm curious about solving the issue by making the shot control handlers global rather than re-creating the handlers on mode start (when the switch handlers are re-created). What's your thought there?

gbouthenot:
Thank you Anthony. I will build a minimal test case to explore this
behaviour.

gbouthenot:
Hello Again.

I built a minimum mpf game to explore the shot problem I have. There are instructions in the readme.

It is here: https://github.com/gbouthenot/mpf-shotbehaviour

mpf 0.57 + mpf-mc

I don't know (yet) how to build / run test cases, but it would be great if I learn how to do. 

You can see our pinball here : https://155601.wixsite.com/rebellionpinball there is a carousel with some photos. I am the one playing next to Gary.

gbouthenot:
About your question:
>I'm curious about solving the issue by making the shot control handlers global rather than re-creating the handlers on mode start (when the switch handlers are re-created). What's your thought there?

It took me some times, but I think it would be better to remove the control handlers when the shot is destroyed and recreate them when the shot is recreated. You will probably agree.

I will submit another patch shortly (or modify this one).

EDIT: i was beginning to modify the patch, but I saw that this can lead to some bad consequences.

Currently `self._register_control_event_handlers()` is called in `initialize`, and so the control event handlers are global, and created at boot time.
They are deleted when the shot is created and deleted (and never recreated). I think the deletion of the control handlers is a bug.

This Pull Request respects this behavior and fixes the control events from being removed. I think that's the way MPF handle control events. They are always here from boot.

Changing this behavior (having the control events only added at create time), even if is seems more correct may break some games.

What do you think ? Should I submit another Pull Request that will modify control events behavior (only create them when the shot is created) ?

avanwinkle:
Fantastic documentation, thanks for the sample project.

**Regarding YOUR situation**
I feel like your shot setup is more complex than it needs to be, and using control events to change states while simultaneously managing enabled/disabled is why you're experiencing a bug that nobody has discovered before.

Your first profile state "dis" is only used when the shot is disabled, which means the shot will never register a hit and therefore this profile state will never be triggered. The preferred approach would be to remove this state and have "lit" be the initial state, and instead of `enable_events` on the shots use `restart_events` which will both enable the shot _and_ set it to the initial profile state. 

When set like this, you won't need any `control_events` on your shots and you won't experience this issue.

**Regarding the actual Bug**
Nevertheless, this is indeed a real bug and should be fixed! 

I think it makes sense to always register the control handlers when the shot is enabled, and remove the control handlers when disabled. But when the feature for control events was first written, it was specifically called out that it would work even if a shot was disabled, so probably best not to break that behavior.

I suppose... we could just never delete the control handlers? Most shots don't use them, so it's not like the machine would be overburdened by extra handlers I guess. The other option would be to determine if the shot is a mode device or a machine device, and create/remove the handlers on mode start/end for mode devices and do it on initialization for machine devices.

avanwinkle (approved):

PR #1890, opened 2025-04-22
add event_player hooks to high_score built-in mode

bosh:
Uses `left_flipper`, `right_flipper`, and `start` that many players will already have incorporated in their switch configurations. I'll make a docs PR as well to go along with this (the switch tags list(s), the MPFTextInput reference, and eventually a custom high_score slide override guide).

I looked at the high score tests and wasn't sure what would be right to add for this addition. I could add a test that says "when I press a switch with a relevant tag, the text_input event with the correct action is fired" but that feels pretty mock-y vs. the usual integration style I would describe as the MPF standard. I think it's because the "text_input" event behaviors are only handled on the GMC side, which is out of scope for this test suite.

### DIFF
--- a/mpf/modes/high_score/config/high_score.yaml
+++ b/mpf/modes/high_score/config/high_score.yaml
@@ -7,7 +7,7 @@ mode:
   use_wait_queue: true
 
 # Instructions on how to use this mode:
-# https://missionpinball.org/game_logic/high_scores
+# https://missionpinball.org/latest/game_logic/high_scores
 
 high_score:
   enter_initials_timeout: 60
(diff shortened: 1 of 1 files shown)


bosh:
Also for a guide, we should revisit the attract carousel to incorporate the various high score display slides

Much later edit: https://github.com/bosh/trogdor-pinball/blob/78d7253cf58f22fefdbc143dbf8de34090b46eea/modes/attract/config/attract.yaml#L16 has an example of the `selectable_item` for the yaml side, and https://github.com/bosh/trogdor-pinball/blob/78d7253cf58f22fefdbc143dbf8de34090b46eea/gmc/slides/attract.tscn#L387 has the godot node to match

bosh:
Documentation update in https://github.com/missionpinball/mpf-docs/pull/596

avanwinkle (approved):

PR #1894, opened 2025-05-27
Fast exp light redistribution

bosh:
Work in progress - MPF now can accept a custom count per header and send the ER commands necessary to make it happen

Notes: BR will reset the board state, so ER must be done after it
Tests should generally work, but no test exists to check that >128 errors

Addressing of the LEDs uses the old numbering, which is a little confusing. EG if you configure port 1 to have 1 light, and port 2 to take the other 31 (so it has 63), then exp_board-1-1 is the first light on the first string, then exp_board-1-2 is the first light on the second string. exp_board-2-1 would actually be the 32nd light on the second string, because all of the port 1 lights would have been pushed in front of it. This can be fixed in fast.py configure_light/port_idx_to_hex/parse_light_number_to_channels but it may be a surprisingly wide area we need to fix up to support this. It also somewhat requires the lights-per-header details to be resolved before any lights can be registered. Maybe that setup order dependency is already satisfied though

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -670,7 +670,6 @@ fast_breakout:
     led_ports: list|subconfig(fast_led_port)|None
 fast_led_port:
     port: single|str|
-    type: single|enum(ws2812,apa-102)|ws2812
     leds: single|int|32
 fast_aud:
     port: list|str|auto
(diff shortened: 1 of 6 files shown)


bosh:
FAST has published .48 firmware for EXP boards, which supports 128 lights on a single header (thus using all four headers' addresses in one). I think my plan is to change the limit provided in this PR to 128 and use count + offset from 0 to define what you want the chain to show. This allows for funny things like the other headers having mirrored subchains, which may be desirable (and is not possible with the syntax provided in this PR.) 

I'm going to close this one and resubmit later with a revamp.

Note that anyone using 128 LEDs in a chain will find that voltage drop will keep the lights from all lighting, and thus will need to inject power and ground at secondary points in the chain.

PR #1895, opened 2025-05-28
Adding definition for FAST internal EXP bus. Needed when using Raspbe…

albano353:
…rry Pi directly connected to Neuron board, when Neuron LED expansion is needed in addition to primary EXP bus. Allows MPF config to explicitly set serial port for internal and external EXP bus. These interfaces are merged together when connecting via USB, but are broken out to separate pins on the Neuron Raspberry Pi header.

Here is a working example MPF config using exp_int:
```
fast:
  net:
    debug: false      # change to true to see full NET serial communication in the log. Search for `[NET]`
    port: /dev/ttyAMA2
    controller: neuron
    io_loop:
      upper_playfield_io: 
        model: FP-I/O-1616
        order: 1
      lower_playfield_io: 
        model: FP-I/O-3208
        order: 2
      cabinet_io:
        model: FP-I/O-0024
        order: 3
        
  exp_int:
    debug: false      # change to true to see full EXP serial communication in the log. Search for `[EXP]`
    port: /dev/ttyAMA3 # ttyAMA3 for Neuron "internal" EXP bus
    boards:
      neuron_exp:
        model: FP-EXP-2000

  exp:
    debug: false      # change to true to see full EXP serial communication in the log. Search for `[EXP]`
    port: /dev/ttyAMA4  # ttyAMA4 for "external" EXP bus
    boards:
      playfield_exp:
        model: FP-EXP-0071
      stepper_exp:
        model: FP-EXP-0061
```

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -662,13 +662,6 @@ fast_exp:
     console_log: single|enum(none,basic,full)|none
     file_log: single|enum(none,basic,full)|basic
     boards: dict|str:subconfig(fast_exp_board)|
-fast_exp_int:
-    port: list|str|auto
-    baud: single|int|921600
-    debug: single|bool|false
-    console_log: single|enum(none,basic,full)|none
-    file_log: single|enum(none,basic,full)|basic
-    boards: dict|str:subconfig(fast_exp_board)|
 fast_exp_board:
     model: single|str|  # FP-EXP-0071, etc.
     address: single|str|None  # overrides the lookup address, useful for partner-specific firmware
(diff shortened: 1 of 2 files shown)


mpf/config_spec.yaml
```diff
@@ -661,6 +662,13 @@ fast_exp:
     console_log: single|enum(none,basic,full)|none
     file_log: single|enum(none,basic,full)|basic
     boards: dict|str:subconfig(fast_exp_board)|
+fast_exp_int:
```
bosh: I'm not sure if this section is used, because you used "fast_exp" as the subconfig type on 631, which should reuse the definition at 658 instead. But I'm also pretty new at this config_spec structure so I could totally be wrong

albano353: Good catch! I missed that when cleaning up from my initial testing. Thanks, I'll remove it.

avanwinkle (approved):

PR #1900, opened 2025-07-01
allow FAST port detector to actually retry on timeouts

bosh:
Using pass silences follow-on errors, and I think we're getting one (maybe data is populated with something truthy but crashy?) and never actually retrying. I'm also not certain that log lines from the async tasks are actually getting printed (even with -vV etc), so I had to resort to debugging with just `print` heh.


bosh:
I've tested this live now on a FAST board setup that _always_ fails on first EXP connection attempt after powering up the FAST system (and sometimes fails on NET after first power, though this is more like 10% of the time instead of 100%) and found that MPF is now able to establish the connection successfully (after a try or two) on many repeated attempts.

avanwinkle (approved):
So simple!

PR #1902, opened 2025-07-26
Update item highlighted event comment

bosh:
Note: Comment-only change.

Noticed this while helping Ernie debug his new game.

0.57 called this (carousel)_(item)_highlighted but .80 changed to the static name "item_highlighted" so that GMC could interpret it cleanly (see: https://github.com/missionpinball/mpf-gmc/blob/43170784272f4e89c754288532b874a5f11efb99/addons/mpf-gmc/scripts/bcp_server.gd#L349 )

I'm pretty radical, but I think that no device names nor custom field values should be in event names, preferring everything in event params instead, partly because it would make mapping other events into the GMC handler trivial. OTOH I haven't had any need to map any more events either. Could be fun for a 1.0 overhaul something though when we all win the lottery :P

### DIFF
--- a/mpf/modes/carousel/code/carousel.py
+++ b/mpf/modes/carousel/code/carousel.py
@@ -101,10 +101,12 @@ class Carousel(Mode):
                                  carousel=self.name,
                                  direction=direction,
                                  item=self._get_highlighted_item())
-        '''event (carousel_name)_(item)_highlighted
+        '''event item_highlighted
             desc: Player highlighted an item in a carousel. Mostly used to play shows or trigger slides.
             args:
+               carousel: The name of the carousel that posted this event.
                direction: The direction the carousel is moving. Either forwards or backwards. None on mode start.
+               item: The name of the item that is now highlighted.
             '''
 
     def _get_available_items(self):


bosh:
Hm, this might still be a good opportunity to make the carousel events all have static naming (though it's a last minute change for our early .80 adopters) -- I've been writing up a PR adding all of the events to the event reference pages in the docs repo, and the events I have are:

* item_highlighted
* (carousel_name)\_items_empty
* (carousel_name)\_(item_name)\_selected
* (carousel_name)\_item_selected

where the last two are posted at the same time, and only differ in that one is intended to be used for hooking onto (for whichever mode or followup behavior the player has selected), and the other is more generic and (per the comment in MPF next to it) is used to dismiss the carousel mode itself.

I propose we change to: (params in parens)
`carousel_item_highlighted(carousel name, item, direction)`
`carousel_items_empty(carousel name)`
`carousel_item_selected(carousel name, item)`

I think this is our best chance to do this without causing even more disruption for the users.

avanwinkle (approved):
Thanks for this! Interesting thoughts on revamping the names all up, GMC is definitely moving towards arguments instead of one-off event names. Let's explore it!

bosh:
I'm down. I've made a [PR to touch up the file contents for .57](https://github.com/missionpinball/mpf/pull/1906) so that any changes (like the event names) made for .80 will have a smaller surface area, and the two different versions will be easier to compare in the future. 

EDIT: Wait, this PR was about carousel events! I've been in high score land for a little while and was thinking high scores could do it too!

PR #1903, opened 2025-08-18
increase default number of top player scores kept in audits

bosh:
We've been keeping only 10 scores in the audit log for over a decade. I propose the default be changed to 1000 or more. I think the default audits are good enough that most developers will not dig into customizing them for a long time, and before they ever do it would be nice to store more than just 10 scores.

I guess what I'd really rather have is a time+score audit trail rather than top score, since top will resort over time. But rather than build out yet another thing, let's start with making the default more useful.

### DIFF
--- a/mpf/mpfconfig.yaml
+++ b/mpf/mpfconfig.yaml
@@ -249,7 +249,7 @@ auditor:
       -  game_started
       -  ball_ended
       -  game_ended
-    num_player_top_records: 10
+    num_player_top_records: 50
     audit:
       -  shots
       -  switches


bosh:
Failure on one test build is just a http 500 flake

avanwinkle:
All default config values are baseline and superseded by any project config files. In your _config.yaml_ you could put the following:
```
auditor:
  num_player_top_records: 1000
```
and your project will store 1000 :)

bosh:
Was trying to press the "add a comment" button on my phone but it seems to not work, so I hit the close/open by accident lol.

Re overriding: yeah I did that for myself! My point is that most devs aren't even going to notice this setting to change it for a long time, and there is likely no space concern (1000 scores would result in a few kbs of stored data), so why not bump it up from 10 to something higher as a default. 10 scores is 2.5 four-player games. Why not up it to at least 100 trailing scores? I picked 1000 because why stop at 100?

I could buy either of these following counterpoints:
- this trailing score history isn't intended to be your long-term score repository, so we shouldn't encourage it by flooding it with data
- keeping the number low will encourage devs who find it to explore and learn more about the audit system

But IMO I like convention over configuration in this case and would double down on saying "why not 100?"

PR #1907, opened 2025-08-20
Breaking change -- rename carousel events to have static names

bosh:
Name and item have been moved to event params where previously part of the name.

Note that this is a divergence from .57 and before, and will require upgrading users to revisit their carousel hooks.

GMC also will need a change (PR link in a moment) to expect the new carousel_item_highlighted event name, which also means that the GMC minimum MPF version would be the next 0.80.0.dev release, dev10 (assuming this gets in there)


avanwinkle (approved):
This makes sense and is a good direction. Looks like the GMC/MPF cross version-checking came just in time!

bosh:
I've prepared the corrected documentation for the new event names and fixed up a minor typo in this PR's comments

PR #1906, opened 2025-08-20
Refactor high score py

bosh:
This is a preparation PR targeting .57 to get the file closer to what the work proposed in https://github.com/missionpinball/mpf/pull/1902 (renaming event names) would look like. This PR should have 0 behavioral changes, and be purely format (and some commonly-used variable extractions)

PR #1908, opened 2025-08-20
High score event refactor

bosh:
This changes from sending three events on high score submission to just one, `high_score_award_display`, which has all the params of the previous two included, so a dev can filter and hook onto whatever they need.

This branch includes two cherry-picks from https://github.com/missionpinball/mpf/pull/1906 and for a cleaner history will be rebased once dev and 0.80 have that branch merged in.


bosh:
Replaced by https://github.com/missionpinball/mpf/pull/1909

PR #1909, opened 2025-08-22
GMC - Add high score award slide hooks + change to static event names only for high score

bosh:
Replacing https://github.com/missionpinball/mpf/pull/1908 because the changes need to go together

From previous:
This changes from sending three events on high score submission to just one, high_score_award_display, which has all the params of the previous two included, so a dev can filter and hook onto whatever they need.

This branch merges dev into 0.80.x to get the refactor commits I made so that the two versions look somewhat similar after all the other changes made in .80

Additional:
This also re-enables the default hooks for showing an award slide between each prompt for initials.
Because the slides are removed instead of stacked, the changes suggested in GMC https://github.com/missionpinball/mpf-gmc/pull/33 are not necessary. However, I think that because the class is generically Text Input, not specifically a high score slide, it's reasonable to clear and reset the input anyway. But I'm happy either way on that PR.

### DIFF
--- a/mpf/modes/high_score/code/high_score.py
+++ b/mpf/modes/high_score/code/high_score.py
@@ -102,12 +102,10 @@ class HighScore(AsyncMode):
         for category, entries in self.high_score_config['categories'].items():
             try:
                 for position, (label, (name, value, *hs_vars)) in (
-                        enumerate(zip(entries,
-                                      self.high_scores[category]))):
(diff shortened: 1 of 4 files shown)


bosh:
Converted to draft while prerequisite commits in https://github.com/missionpinball/mpf/pull/1906 await merge to dev and 0.80.x

avanwinkle (approved):

PR #1910, opened 2025-08-31
add default queue_event_player events_when_finished: "queue_event_complete"

bosh:
If the config provides events_when_complete, those are used instead. The queue_event is included in either case, along with the optional args provided in the config.

Test updated (saw em fail without my code change!) and doc comment added :)

### DIFF
--- a/mpf/config_players/queue_event_player.py
+++ b/mpf/config_players/queue_event_player.py
@@ -18,23 +18,28 @@ class QueueEventPlayer(ConfigPlayer):
         """Post queue events."""
         del kwargs
         del calling_context
-        if settings['events_when_finished']:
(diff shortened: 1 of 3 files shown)


bosh:
Alternatively we could provide a default via config_spec.yaml, but since we have the doc strings next to the event posting in code, I like having the name there instead of spread across files. (And if someone were to specify their config with a blank value overriding the default, they might get a crash anyway, vs. the device code forcing a safe default)

avanwinkle:
As discussed on the google group, adding a default value to queue events finished will cause undesirable behavior and should be avoided.

If passing the queue event or other params in the callback has a use case, that could be useful thougH!

PR #1913, opened 2025-09-06
try enabled python 3.13 builds

bosh:
let's see what fails!


bosh:
I think I worked through it all, opening a PR on the better branch

PR #1914, opened 2025-09-06
enable python 3.13 builds and upgrade pillow to avoid setuptools incompatibility

bosh:
Looks like python 13 setuptools had an issue with pillow < 10.1 or so (nested execs when establishing the __version__ info? No matter!)

There was a note that 10.x broke Kivy somehow, but no tests seem to be failing, not sure if there's something else I need to check.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -11,10 +11,6 @@ jobs:
       matrix:
         include:
         - os: windows-latest
-          python-version: 3.8
-        - os: windows-latest
-          python-version: 3.9
-        - os: windows-latest
           python-version: "3.10"
         - os: windows-latest
           python-version: 3.11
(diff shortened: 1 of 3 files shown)


bosh:
See: https://github.com/missionpinball/mpf/pull/1915#issuecomment-3260389436

avanwinkle:
No tests will fail here because Kivy doesn't run with MPF, but MC depends on Kivy so upgrading that could be finnicky.

I would suggest we target these changes at 0.80 to avoid any Kivy-related shenanigans entirely, and leave 0.57 at 3.12 where MPF and Python can fade away into deprecation together.

Thanks for tackling this!

bosh:
Gotcha, will retarget on this branch. What would you think about dropping Python 3.8 and 3.9 with the .80 bump as well?

avanwinkle:
Sounds like a good idea! I'm all in favor of keeping current with versions. Thanks!

Issue #1912, opened 2025-09-06
User wants to use MPF with python 3.13


bosh:
Looks like 3.13 is about a year old, would be nice to add support

<img width="1114" height="630" alt="Image" src="https://github.com/user-attachments/assets/531e4ee5-85af-4f63-aa40-4eb6b009128b" />

I'm guessing the user is specifically running into this line https://github.com/missionpinball/mpf/blob/00460954febea09ae82d7b18412edd99fedf9260/pyproject.toml#L5

bosh:
Obviously that's just our guard, but after (more like before...) testing out 3.13 we run into a setuptools issue with Pillow and maybe nested execs :)
https://github.com/pypa/setuptools/issues/4415
https://github.com/python/cpython/issues/118888#issuecomment-2104944287

bosh:
Upgrading pillow fixed it -- looks like there were early 10.0 versions that were broken for various reasons
https://github.com/missionpinball/mpf/pull/1914

bosh:
See: https://github.com/missionpinball/mpf/pull/1915#issuecomment-3260389436

bosh:
Closing with resolution - MPF .80 is dropping 3.8 and 3.9 and adding 3.13 support, along with an upgrade of pillow necessary to fix an issue with nested execs or perhaps variable passing

Issue #1893, opened 2025-05-24
Python 3.8 is EOL and 3.9 soon


bosh:
I noticed last week when helping with a working .56 machine going to .57 that there are some differences between old python versions that MPF doesn't really support. I can't remember the specifics, but it was stuff like standard library imports changing namespaces. Today I noticed that latest pip has dropped support for Python 3.8 as well.

https://devguide.python.org/versions/ 
per this chart, Python 3.9 is close to EOL as well -- I propose for MPF .58 we drop support for 3.8 and 3.9 going forward. I don't think there's any rush to cut a new MPF version, but by the time we do, I think 3.9 will have hit EOL (< half a year away at this writing)

It would be nice to add official support to the new pythons at the same time as well, maybe I'll try to dig back into that soon

bosh:
See: https://github.com/missionpinball/mpf/pull/1915#issuecomment-3260389436

bosh:
Closing with prospective drop of 3.8 and 3.9 from mpf .80 release

PR #1915, opened 2025-09-06
Pi thon 3.14

bosh:
Updates python support to latest, 3.14

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -60,10 +60,17 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install Linux dependencies
-        if: matrix.os == 'ubuntu-latest'
-        run: sudo apt-get install -y libjpeg-dev
-
+      - name: Install jpeg support for pillow
+        shell: bash
+        run: |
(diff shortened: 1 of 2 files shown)


bosh:
Python 3.8 is EOL and 3.9 will be soon -- Python 3.14 beta is only supported by Pillow 11.3, and 11.3 drops support for Python 3.8.

IMO we should cut 57.4 as is with existing support back to 3.8, then drop 3.8 and 3.9 and add 3.13 and upgrade Pillow to 11.3, and then release that as .58.0.

Then we merge 58 into 80 and thus release 80 without 3.8 or 3.9 ever officially supported, rather than immediately moving from .80 to .81 with the deprecation of 3.8/9


Python 3.14 seems to work (having tested rc2), but as it is not fully released and we haven't tested anything live (and adding support wouldnt need to bump the MPF minor version), so we can hold off on that and add 3.14 support in a few months to .58 and .80

bosh:
Previous steps towards the end goal
https://github.com/missionpinball/mpf/pull/1914
https://github.com/missionpinball/mpf/issues/1912
https://github.com/missionpinball/mpf/issues/1893

bosh:
Holding off until official release of Python 3.14 in October

avanwinkle (approved):

PR #1917, opened 2025-10-02
Dev10 needs gmc 0.1.6 due to breaking event name changes

bosh:
We should rerelease dev10 with the gmc dependency upped from this change. I’ve helped one user who ran into this issue already, luckily it seems that most other gmc users who have upgraded in the last days just use dev branch instead of numbered gmc releases


avanwinkle:
Oops again!

Overwriting an existing release is messy with reinstalls and caches and stuff, let's just roll out dev11 and keep it clean.

Thanks for staying on top of this!

PR #1919, opened 2025-10-07
Fix and improve sequence_shot

gbouthenot:
Sequence_shot did not support duplicate events. A test exists, but it tests another thing.
This pull request:
- Add a proper sequence shot with duplicate test (it fails)
- Fix `sequence_shot.py`, so it passes.
- Allow cancel_switches to be part of a sequence (with a new test) -- see the commit messages for more information.


bosh:
Dang this is really cool! I didn't realize sequence shots automatically managed multiple parallel progressions in the first place. I think sequence-included cancel switches being like a filter (where progressing sequences survive and step forward, and rest are reset) makes sense as a rule set as you have coded it, but certainly will be a fun time to document clearly, so thank you for the examples in both event and switch style.

I do wonder whether the logic-block style "sequence" object is a better place for managing complex event or switch hit chains turning into success/reset decisions. Looking into that type, it seems significantly less powerful than `sequence_shots` (no automatic timeout, no multi-sequence management) so I think sequence_shots is a better fit this continued evolution.

My only concern is that sequence_shots may have been intended specifically for certain use cases within a machine, which may be why there is a distinction between sequence shots and sequences (and of course the general fallback of using state machines). None of those other types support multiple live evaluations of their progress through various states though, so sequence_shots seems to have a unique place among them.



I don't have any authority here/I'm not asking for any changes (looks good to me!) but I have a couple questions, since you have just been thinking about sequence shots for a bit, so if you do not mind exploring with me:

- Not thinking about cancel events at all now -- I noticed the existing comment:

```
# Get the seq_id of the first sequence this switch is next for.
# This is not a loop because we only want to advance 1 sequence
```

It seems like the decision to only advance one active sequence is arbitrary (why not advance all actives that are waiting on that step?). Does advancing multiple active copies each by one seem reasonable ever to you? Both options seem useful in different cases to me, and I could be interested in building that distinction. Adding a config boolean that lets the dev decide whether one or all should advance seems like an easy config upgrade and guard on just one place in the code.

- Similarly, it seems that the sequence_shot config doesn't have any option to _disallow_ multiple active sequence copies. I didn't know about this feature at all, so it's more surprising to me that multiple can run at different progression points in the first place. I would be interested in building another config boolean that would default allow multi-sequence tracking, but allow restricting it to just one active sequence at a time. Does that seem like a reasonable feature to you?

gbouthenot:
Hello Bosh

I am glad you asked. Why ? Because I though the same thing: "the advance sequences should all be advanced !". And that's what I did and ...

```
Test interleaved_sequences (Two balls pass through the sequence) FAILED !
```

Let's see:
The sequence is waiting for "event1", "event2", "event3"

The test pattern is "event1", "event1", "event2", "event2", "event3", "event3"

and is waiting for the sequence to be hit at both "event3" (1 time per "event3")

Let's see step by step:

CASE 1: WE ADVANCE ONLY ONE SEQUENCE
(1): 'event1':
sequence A: step 1, waiting "event2"

(2): 'event1':
sequence A: step 1, waiting "event2"
sequence B: step 1, waiting "event2"

(3): 'event2':
sequence A: step 2, waiting "event3"
sequence B: step 1, waiting "event2"

(4): 'event2':
sequence A: step 2, waiting "event3"
sequence B: step 2, waiting "event3"

(5): 'event3':
sequence A: SUCCEED
sequence B: step 2, waiting "event3"

(5): 'event3':
sequence B: SUCCEED

Ok perfect


CASE 2: WE ADVANCE ALL SEQUENCES

(1): 'event1':
sequence A: step 1, waiting "event2"

(2): 'event1':
sequence A: step 1, waiting "event2"
sequence B: step 1, waiting "event2"

(3): 'event2':
They both advance !
sequence A: step 2, waiting "event3"
sequence B: step 2, waiting "event3"

(4): 'event2':
no change (!)
sequence A: step 2, waiting "event3"
sequence B: step 2, waiting "event3"

(5): 'event3':
sequence A: SUCCEED
sequence B: SUCCEED
BOTH sequence A and sequence B will succeed at the same time !

(6): 'event3':
nothing

That's why Jan (Initial commit: 628a5ff3) made it work like this. So I kept this
idea of only advance one sequence (but I advance the more advanced one, instead of random).
I say random because in the code, in `event_cancel` and `_sequence_timeout` the list is re-created,
and there is no fixed logic in the previous code

But I wanted to be able to have a sequence "1,1,2,2" that is not triggered by "1,2,1,2", that's why I added the ability to have `cancel_switches` that are part of the sequence (but this only work with switches, because cancel_event does not provide the event name to `event_cancel`).


I agree sequence_shots is more powerful than sequence.

I Hope I answered your thoughts.

bosh:
I think I follow -- if you allow multiple sequences at once, and one iteration starts then progresses further in, then a second iteration starts and catches up to the first one, then allowing multiple to advance really just means they have converged and will now always move together. That does seem like undesirable behavior (since the goal of the sequence shot is to confirm a ball traveling a given path between switches). So a config option to `allow_multiple_advance` or whatever could cause unexpected behavior there. I think you're right that the most-deep-into-sequence instance is the one that should advance, not whichever happened to be first or last created.

It does seem like it might be still worthwhile to have a config option that disallows multiple sequence iterations at the same time. I guess the downside with single-iteration-only is that you then have to decide whether the right implementation is to ignore new sequences until the ongoing one is completed or times out, OR to always replace the in-process sequence when the first switch of the sequence is hit. I don't like the latter option, so maybe my suggestion is:

Property: `sequence_shots:allow_multiple_active` -- `boolean, default True`
This property is True by default to preserve existing behavior on upgrade. If set to false, then the sequence shot will not allow further iterations of the sequence to process when a new sequence is started and has not finished resolving. The active sequence cancelling, timing out, or completing will resolve it, and then a new switch hit can then start a new instance.

I do think the intended behavior of sequence_shots is to do things like track balls going over direction paths, such as an orbit that can be traveled from either end. So it makes sense to allow multiple instances of the same sequence when doing things like multiball, so that if the player shoots the same orbit with two balls at once, two pairs of hit first switch, hit second switch can resolve and report two successful orbits in the end.

gbouthenot:
I added "allow_multiple_active" property, and created a test case. Hope it is ok.

bosh:
Awesome, thank you! I envision that being useful for the case where the sequence is a ball bouncing in and out of something, for instance:
sequence is 1-2-1-3, where 1 is a rollover sensor, 2 is a target that the ball bounces off, and 3 is some confirmation switch for a well-bounced shot. I wouldn't want multi-active because the second 1 would trigger a second sequence while not representing  a second ball actually traveling the path.

bosh (approved):
Approved now that I can officially say so! :)

PR #1921, opened 2025-10-24
Disable crash reporter by default

bosh:
Quiet the crash reporter by default because the service does not exist.

Could go further and remove the feature entirely if not desired, but at least this stops asking users to do something that will just add more error lines to their log.


avanwinkle (approved):
Great!

PR #1922, opened 2025-11-04
Shaker virtual platform support

bosh:
Still looking into tests/etc

### DIFF
--- a/mpf/devices/shaker.py
+++ b/mpf/devices/shaker.py
@@ -15,7 +15,7 @@ if MYPY:    # pragma: no cover
 
 class Shaker(SystemWideDevice):
 
-    """A digital output on either a light or driver platform."""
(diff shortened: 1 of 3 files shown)


avanwinkle (commented):
Looks good! Might as well target 0.57 for folks :)

bosh:
@avanwinkle I'm down for 0.57 but you only merged the shaker support into 0.80.x. I'm not sure the best strategy to pull just that from 0.80.x back to dev -- is there a better choice than cherry-pick?

avanwinkle:
Right right, I forgot that I merged it first into 0.80 so people could test it. Haven't heard any complaints yet, so I've opened #1928 to get it into dev.

avanwinkle (approved):

PR #1925, opened 2025-11-08
Remove outdated configs

bosh:


avanwinkle (commented):
Great!

bosh:
@avanwinkle I'm happy to upgrade this to a real PR and get it in -- the lint errors are shaker-specific and are handled in my shaker-related branch, so this is in fact green once including that work.

avanwinkle (approved):

avanwinkle (approved):
Great! We still have issues with folks putting sounds and keyboard in their config and it not working, so hopefully this will get them sorted out faster. Thanks!

Issue #1920, opened 2025-10-08
dev (57 and 80 latests) handles multi-channel lights incorrectly when using "previous:" on the next light *maybe virtual only*


bosh:
Per the draft PR branch: https://github.com/missionpinball/mpf/pull/1918

If registering multi-channel lights in virtual, using latest dev version (confirmed also against latest .80 dev version), when using the "previous:" addressing scheme, channels ending in -9 are sorted later than channels using -10 or -11 (for example). This causes the successor number algorithm to pick +9 as the last seen address, and thus generate +10 as next. The dedup enforcer then sees +10 as already used (as it is!) and rejects the config.

I spent about 3 hours hacking different parts of the +1 system in different ways to address this, but was unsuccessful at getting the simple replications seen in the PR to pass. I'm thinking Anthony might have more context on the overall +1 addressing system due to work on https://github.com/missionpinball/mpf/pull/1889 , so I'm going to step back to see if any other ideas are out there.

One workaround for now is to change any lights using `previous:` that cause this crash to just `number:` definitions. The downside here is that the +N chain resets to +1, meaning the next crash will likely now occur 10 channels down the line (so, every 3-5 lights).

My thoughts on possible solutions are to make the `__lt__` function on the virtual light class do position-wise comparisons of integers, so that channel base+10 properly sorts after +9 instead of before, OR make the +values resolve immediately if the address of the "previous" light is definite and ends in an integer. Though maybe you'll still need to cast to int on the digit chunks from the address string, because it'll just move the rollover problem (for an address that might be called BASE-X+Y now) from the +Y to the -X value in the number.

An option I dont think works is to start zero-padding addresses to 2 digits -- you'll just hit the problem on the channel offset 99 to 100 rollover, which is ~25 lights into an RGBW string. So three digits? Maybe not

avanwinkle:
This is an interesting conundrum! I'm surprised there's any sorting but doing a custom less-than may be the way to go. It already converts values to int so there's gotta be a place to adjust where the comparison happens.

If you can make a simple sample project with a test then I can hammer away at the math of it.

Interested to know whether it's virtual-only, that could definitely reduce the scope of the change.

bosh:
@avanwinkle this single-file config is sufficient to trigger the crash when run with `mpf game -btx`, since the crash occurs in `init_phase_4`

```yaml
#config_version=6

mpf:
    report_crashes: never

lights:
    first:
        type: grbw
        start_channel: 0-0-0

    second:
[...]

PR #1934, opened 2025-12-12
Fixed auto light numbering for virtual lights.

jimhenshaw:


mpf/platforms/virtual.py
```diff
@@ -529,7 +529,15 @@ def split_light_address(address):
...
-        return self.number < other.number
+        self_address, self_offset = VirtualLight.split_light_address(self.number)
+        other_address, other_offset = VirtualLight.split_light_address(other.number)
+        #test address first
+        if self_address > other_address:
+            return False
+        elif self_address < other_address:
+            return True
+        else: #then test offset as int instead of string
+            return int(self_offset) < int(other_offset)
```
bosh: The whole conditional feels weird to me -- is there some subtlety to the comparison order executing as it is written here, or could this start with a == check?

```py
if self_a == other_a:
  int(self_o) == int(other_o)
else:
  self_address < other_address
```

But actually this kinda rings a bell -- won't this still have issues ordering lights with the 9 vs 10 issue, but one address unit up? split_light_address leaves the left part of the split as a string, so it should misorder 9 vs 10 as 9 vs 1 again.

I.E. instead of the issue being on `led-0-0-10` vs `led-0-0-9`, it'll be on `led-0-10-*` vs `led-0-9-*`? I think my conclusion in my research before was that you would have to check addresses segment by segment. I could be wrong, but a test case could help prove it.

It's also possible that we only ever compare light addresses of near-neighbors, so the higher-order comparisons being wrong might not matter. Or it could be that at a certain order, the lights aren't even on the same string, and order comparison is nonsensical (Is light 30 on string 2 truly "less than" light 1 on string 3?)

bosh:
I downloaded this change and it did avoid the crash I saw on dev with my sample config: https://github.com/missionpinball/mpf/issues/1920#issuecomment-3380307223

I think this could be helpful to unbreak current users who are blocked from upgrading to 80 or latest 57 (and dont want to renumber their uses of `previous`) but it might not yet be a perfect solution around light ordering.

mpf/platforms/virtual.py
jimhenshaw: This solution will still have the issue of ordering addresses you mentioned above since it is a string comparison. Address comparison will only be used when machine.lights is called and sorted in things like when light maps are created for the service controller and listing lights in the bcp_interface. The get successor function in virtual.py never augments the address though.

    def get_successor_number(self):
        """Return the number with the suffix '+1'.
       As there is not real number format for virtual is this is all we can do here.
        """
        addr, offset = VirtualLight.split_light_address(self.number)
        return f"{addr}+{int(offset) + 1}"
        
get_successor_number in light.py will add drivers for each color channel to an empty array and sort them. It will then call the above get_successor_number on the tail.

    def get_successor_number(self):
        """Get the number of the next light channel.

        We first have to find the last channel and then get the next number based on that.
        """
        all_drivers = []
        for drivers in self.hw_drivers.values():
            all_drivers.extend(drivers)
        sorted_channels = sorted(all_drivers)
        return sorted_channels[-1].get_successor_number()

Drivers are created in light.py def _load_hw_driver_sequentially(self, next_channel): on line 227
https://github.com/missionpinball/mpf/blob/a19c3a330034d426b56d7655804e4014e9cd4f20/mpf/devices/light.py#L227
 
They eventually call parse_light_number_to_channel
 https://github.com/missionpinball/mpf/blob/a19c3a330034d426b56d7655804e4014e9cd4f20/mpf/platforms/virtual.py#L232C9-L232C39
which just copy/pastes the number

            return [
                {
                    "number": str(number) + "-r",
                },
                {
                    "number": str(number) + "-g",
                },
                {
                    "number": str(number) + "-b",
                }
            ]

This means the address will always be the same for all drivers in a virtual lights. For uses cases outside of virtual light:

machine.lights is used in check_duplicate_light_numbers in light.py (not affected by bad ordering)
service_controller get_light_map  creates a light_map and calls sort (not negatively affected by bad ordering)
This is only called to list lights in bcp_interface.py and for the light test in service.py (not negatively affected by bad ordering)

In the end, virtual is a way to simulate every platform. Each platform can have it's own address schema which has no standard. String comparison on address is a quick/easy close approximation of the native address sorting without having to be coded to work for every platform in every case. In the cases where it doesn't sort properly, it shouldn't negatively impact the user. Some user may notice that their service mode light ordering is different in virtual vs non-virtual.

Reviewing platform implementations they either don't allow it:

    raise AssertionError("Not possible in FASTMatrix.")

or they did a tuple comparison

    return (self.number.board_address_id, self.number.led_number) < (
            other.number.board_address_id, other.number.led_number)

This is essentially the same as the proposed solution.

I did see that OPP uses the following:

    def __lt__(self, other):
        """Order lights by their position on the hardware."""
        return (self.chain_serial, self.addr, self.pixel_num) < (other.chain_serial, other.addr, other.pixel_num)

I'm unsure how the chain_serial will be affected. Guessing it will either be ignored or absorbed into the address.

jimhenshaw:
Added test case that fails/crashes on previous and passes on updated code. I'm assuming the test case will be auto discovered and doesn't need to be added to a list somewhere.

avanwinkle (approved):
This makes sense to me, and implements the same logic as we did for successor calculations so it's reasonable that it should also be used for comparisons.

If it works and folks like it, I'm good with it. Might cause some ordering changes in virtual but shouldn't impact any real games or really be noticable. 

Thanks for the fix and thorough research!

Issue #1933, opened 2025-12-10
counter control event activating and crashing game when mode not active


borgdog:
I found this by having a counter in a mode that contained:

`counters:
  briefing_counter:
    count_events: timer_briefing_carousel_timer_complete
    count_complete_value: 6
    control_events: 
      - event: flipper_cancel
        action: jump
        value: 6`

If I happened to hit both flippers at once (flipper_cancel) at anytime when that mode was not active, even during attract, the game would crash with an error in the log that ended with:

`Traceback (most recent call last):
  File "C:\Python\Python311\Lib\site-packages\mpf\core\machine.py", line 810, in _run_loop
    raise self._exception['exception']
  File "C:\Python\Python311\Lib\asyncio\events.py", line 84, in _run
    self._context.run(self._callback, *self._args)
  File "C:\Python\Python311\Lib\site-packages\mpf\core\events.py", line 846, in process_event_queue
    self._process_event(event=event[0],
  File "C:\Python\Python311\Lib\site-packages\mpf\core\events.py", line 808, in _process_event
    result = self._run_handlers(event, ev_type, kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python\Python311\Lib\site-packages\mpf\core\events.py", line 757, in _run_handlers
    raise EventHandlerException(
mpf.core.events.EventHandlerException: Exception while processing RegisteredHandler(callback=<bound method Counter.event_jump of <counter.briefing_counter>>, priority=1, kwargs={'value': <NativeTemplate 6>}, key=UUID('a046db0f-46a3-4368-b0e6-757246eb2131'), condition=None, blocking_facility=None) for event flipper_cancel. 'NoneType' object has no attribute 'value'
2025-12-10 09:51:58,633 : INFO : crash_reporter : Will not report crash as "report_crashes" mode is "never"
2025-12-10 09:51:58,633 : INFO : root : MPF run loop ended.`

in chatting bosh and others in the trident discord bosh stated "I looked into the logic block and counters code and indeed the control event handler bindings are never disconnected"

bosh:
It looks like state is nulled when it is [removed from mode](https://github.com/missionpinball/mpf/blob/a19c3a330034d426b56d7655804e4014e9cd4f20/mpf/devices/logic_blocks.py#L129):
```py
    def device_removed_from_mode(self, mode: Mode):
        """Unset internal state to prevent leakage."""
[...]

PR #1935, opened 2025-12-12
Counter control event split

bosh:
https://github.com/missionpinball/mpf/issues/1933

I think the code change itself is kinda gross, I'm sure there's a better way (and the commit that moves my_counter.mode=mode earlier seems like it could cause a meaningful difference elsewhere in the code (nothing fails, but it seems like there would have been a reason why mode was attached only at mode-start-time)).

Still, this solution is close enough to a real one that I figured it's worth putting up to get comments on before I go trying to figure out a good test to prove it.

### DIFF
--- a/mpf/devices/logic_blocks.py
+++ b/mpf/devices/logic_blocks.py
@@ -394,7 +394,7 @@ class Counter(LogicBlock):
         for entry in self.config['control_events']:
             if entry['action'] in ('add', 'subtract', 'jump'):
                 handler = getattr(self, "event_{}".format(entry['action']))
-                kwargs = {'value': entry['value']}
+                kwargs = {'value': entry['value'], 'priority': entry['priority']}
                 results.append((entry['event'], handler, kwargs))
             else:
                 raise AssertionError("Invalid control_event action {} in counter".format(entry['action']), self.name)
--- a/mpf/devices/logic_blocks.py
+++ b/mpf/devices/logic_blocks.py
@@ -365,23 +365,40 @@ class Counter(LogicBlock):
 
         # Add control events if included in the config
         if self.config['control_events']:
-            self._setup_control_events(self.config['control_events'])
+            self._setup_machine_control_events()
 
     def add_control_events_in_mode(self, mode: Mode) -> None:
-        """Do not auto enable this device in modes."""
+        """Activate control events given a mode context"""
+        self._setup_mode_control_events(mode)
 
-    def _setup_control_events(self, event_list):
-        self.debug_log("Setting up control events")
+    def _setup_machine_control_events(self):
+        if self.mode:
+            self.debug_log("Not setting up machine control events for mode counter {}".format(self.name))
+            return
 
-        kwargs = {}
-        for entry in event_list:
+        self.debug_log("Setting up machine control events for counter {}".format(self.name))
+        for (event_name, handler, kwargs) in self._generate_control_event_settings():
+            self.machine.events.add_handler(event_name, handler, **kwargs)
+
+    def _setup_mode_control_events(self, mode: Mode):
+        if not self.mode == mode:
+            self.debug_log("Not setting up mode control events for machine counter {}".format(self.name))
+            return
+
+        self.debug_log("Setting up control events for counter {} in mode {}".format(self.name, mode.name))
+        for (event_name, handler, kwargs) in self._generate_control_event_settings():
+            mode.add_mode_event_handler(event_name, handler, **kwargs)
+
+    def _generate_control_event_settings(self):
+        results = []
+        for entry in self.config['control_events']:
             if entry['action'] in ('add', 'subtract', 'jump'):
                 handler = getattr(self, "event_{}".format(entry['action']))
                 kwargs = {'value': entry['value']}
+                results.append((entry['event'], handler, kwargs))
             else:
-                raise AssertionError("Invalid control_event action {} in mode".
-                                     format(entry['action']), self.name)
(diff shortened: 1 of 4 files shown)


avanwinkle (commented):
This all makes sense, though it's hard for me to understand what the purpose of the previous code was and why logic blocks were explicitly blocked from the normal control events flow.

mpf/devices/logic_blocks.py
```diff
@@ -365,23 +365,34 @@ async def _initialize(self):
...
-        kwargs = {}
-        for entry in event_list:
-            if entry['action'] in ('add', 'subtract', 'jump'):
-                handler = getattr(self, "event_{}".format(entry['action']))
-                kwargs = {'value': entry['value']}
-            else:
-                raise AssertionError("Invalid control_event action {} in mode".
-                                     format(entry['action']), self.name)
-            self.machine.events.add_handler(entry['event'], handler, **kwargs)
+            for entry in self.config['control_events']:
```
avanwinkle: This seems identical to the code block on 378, should we have a subroutine to setup the control events that could be called by both places?

bosh: Yeah for sure -- they differ only by the use of add_mode_event_handler vs events.add_handler, but the second takes a mode param, so probably it can even be unified

bosh:
The unknown for me here is that I'm moving logic_block.mode=mode much earlier, which might have unintended side effects if that being assigned has some effect. Also should check to see if it gets cleared ever (or otherwise gets put in an inconsistent state)

bosh:
@avanwinkle this latest push is ready for review and possible merging :)

(commit below adds a test that fails with two different crashes in dev, and shows noop on mode-specific counters with my fix!)

jimhenshaw:
Some questions I had on the commits:

How does priority affect control events? Why is it needed? Is priority a key word that affects firing order? Where was the offset code implemented?

Why delete the mode in mode_device.py device_loaded_in_mode? Is doesn't appear to get used directly after this function?

Why remove setting self.mode to none in a call to device_removed_from_mode? That seems like proper functionality based on method name. What if the device was attached to a different mode? Is a check needed for self.mode == mode?

bosh:
All good questions! It might be slightly easier to talk about things in-context though next time, so I definitely recommend using the [Files](https://github.com/missionpinball/mpf/pull/1935/files) view in the PR and then pressing the green button "Start Review" so that you can comment on several specific lines in the PR and have contextual discussions on each) (dont forget to hit "Submit Review" at the end!)


> How does priority affect control events? Why is it needed? Is priority a key word that affects firing order? Where was the offset code implemented?

Event bindings in general should allow priority management -- because control event handlers are bound slightly differently than your usual event binding (say, with a config_player, where you can attach conditional and priority modifiers inline, e.g.:
```yaml
event_player:
  my_event{my_condition==1}.1: my_resulting_events
```
)
The offset management occurs inside the implementation of `add_mode_event_handler` https://github.com/missionpinball/mpf/blob/461294ae248cc42032d9401ae919faab2e64f377/mpf/core/mode.py#L537
so all I had to do was provide the param, and the 0 default will be overridden.


> Why delete the mode in mode_device.py device_loaded_in_mode? Is doesn't appear to get used directly after this function?

You'll find a lot of `del paramname` lines in various event handler functions in this codebase. I believe it's an attempt at memory management, but I have a feeling that almost all of the times you see it are completely unnecessary and may not actually do anything at all. It's really just trying to tell the memory management of Python that the param will not be used any further in that function.


> Why remove setting self.mode to none in a call to device_removed_from_mode? That seems like proper functionality based on method name. What if the device was attached to a different mode? Is a check needed for self.mode == mode?

This is necessary because I moved the setting of self.mode earlier in the device lifecycle. It used to be set every time the mode started, and then unset when the mode stopped. My changes require that the counter device knows that it is a mode-specific instance much earlier than before, so I changed it to be set at machine load time, which means it is only set once, instead of every mode start. If I left the `self.mode = None`, we would be in a situation where the mode is set once on load, and then unset whenever the mode stops, and then _not_ set back in place when the mode starts again later, which will definitely lead to bugs.

It's possible that devices can exist in multiple modes, or that devices attached to modes actually do use whether their .mode property is None or is an actual mode when executing some behavior. I haven't ever tried to redefine a device in multiple modes (I'm not sure how I would do that even, I assume that if I define the same instance name of the same device type in two places, the config validator will tell me it's not allowed in all cases (I've definitely seen it do that before with say, two of the same named switch or light). But if it allows it, would only one instance need all the configuration? What if I configured the same device name with two different configurations?) but if you could define a device in multiple modes, wouldn't it need a .modes array, not a single .mode value? My guess is that this is a leftover from an older MPF version, or a feature that never got fully built out.

I think it's a pretty reasonable thing that devices defined in modes should always know the mode they are attached to, rather than only knowing it when the mode is running (thus, moving .mode=mode earlier and not clearing it out), but I do think this is the riskiest part of the change, though the tests dont seem to mind :)

Edit: but yeah, so if a device could attach to multiple modes, wouldnt the old implementation imply that the device only really knows about the most recently-attached one? I did actually include a self.mode==mode check in one place, though honestly I'm pretty sure it's overkill, as it will only ever be called with a matching mode.

bosh:
@jimhenshaw aw I was wrong, to start an inline review you go to the files tab, then you hover a changed line, click the blue + icon, type your comment, and then press Start Review there. In the end you submit your review comments all together as a batch with that upper right button.
image

avanwinkle (approved):

PR #1937, opened 2025-12-19
Counter action in non-running mode crash fix

jimhenshaw:
Fixed issue of counter actions crashing game when mode they are defined in is not running. [https://github.com/missionpinball/mpf/issues/1933]

Added tests to duplicate issue and verify actions still function when mode is running.

### DIFF
--- a/mpf/devices/logic_blocks.py
+++ b/mpf/devices/logic_blocks.py
@@ -427,7 +427,7 @@ class Counter(LogicBlock):
             value: Value to add to the counter.
             kwargs: Additional arguments.
         """
-        if not self.enabled:
+        if (not self.enabled):
             return
         evaluated_value = value.evaluate_or_none(kwargs)
         if evaluated_value is None:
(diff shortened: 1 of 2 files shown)


bosh:
I think the enabled checks are reasonable additions on top of my changes in #1935 . I think we want both changes, as mine ensures that the control events are only bound when a counter defined in a mode actually has the mode running, but it's possible that a counter could be disabled/enabled manually during the run of its mode.

Seems like maybe MPF has too many ways of doing the same thing here, but at least we're guarding them now!

For the tests, I would suggest swapping to a more direct observation of changes in value (see my tests when I compare the counter value to None) rather than mocking the secondary event that is posted.
Also I'm not sure if there's a way that you can get your Github Actions workflow approved on your end (probably a button somewhere on your fork's page, maybe in the actions tab?) or if Anthony has to press something to approve you (since the runners here may be in the missionpinball fork's context)

avanwinkle:
Thanks for the great fixes! Just a few small tweaks for code compliance. If you'd like you can use pip to install prospector and run it locally to check for linter errors before pushing, which saves the wait for the CI job.

avanwinkle (approved):

Issue #1939, opened 2025-12-31
Issues with RGBW LEDs (SK6812) on FAST EXP RGB headers (WS2812 native output)


bosh:
There's nothing we can do about this except implement the [FAST ER command](https://fastpinball.com/fast-serial-protocol/exp/er/) for native RGBW support. That will then require native RGBW users on FAST to [upgrade](https://fastpinball.com/firmware/exp/) their [EXP firmware](https://github.com/fastpinball/fast-firmware) to 0.48.

I've spent far, far too long on this problem and had such little success finding relevant information online that I think it's worth spending a little more time documenting for posterity, if someone ever is foolish enough to try what happened here and wants to understand what's happening. Good luck to you! I can describe the problem, the hack solution that we use today (and hopefully not for long into 2026), the underlying reason this all happens, and even the pattern and how to derive it. But it still hurts my head to think about, and the numbers get kind of swimmy if you look too hard at them.

Let's begin.

So you have a FAST EXP board and some RGB LEDs, maybe standalone, maybe GI inserts, maybe some strings of several lights. Awesome! You wire them up in series, configure them in MPF using `number` or `previous`, write some shows, and it just works. Great!

Then you pick up some sweet RGBW lights, maybe some Cobrapin rings, or maybe a full-on string. You wire them up to an open EXP header, or maybe at the end of an existing RGB chain, add them to the config (using the RGB type!?)

(Note, light show not shown, but it's a simple `R, G, B, W, off` test loop for [lights by catchall "*"](https://missionpinball.org/latest/config/light_player#wildcard-light-tag).

Common setup for all examples:
```yaml
hardware:
  platform: fast

fast:
  exp:
    boards:
      exp:
        model: FP-EXP-0081
```

Your first RGBW config!

```yaml
lights:
  one: {number: exp-1-1}
  two: {number: exp-1-2}
  three: {number: exp-1-3}
```

(vid 1) 

Well that was weird - the first light was mostly correct, but the second and third lights were completely wrong, and sometimes off when they should have been lit!

So you switch to rgbw and you get a crash:
```yaml
lights:
  one: {type: rgbw, number: exp-1-1}
```

```sh
[...]

PR #1941, opened 2026-01-02
Fast er lights

bosh:
WIP -- IMO this is a breaking change and requires a minor version bump to .58/.81, or we even just target .81 and leave .57 series behind. This PR includes the version bump commit, which we can rebase out when the true merge point is determined.

I'm kind of in favor of this -- thinking about the upgrade steps required to move from .57 to .80, if someone really wants the ER-based features in legacy, they will move to .58 to get them. But then they no longer can upgrade directly to .80, since the ER support will disappear (leaving them with either misbehaving lights, or possibly incompatible config specification).
I suppose the steps to go between .58 and .81 would be identical to the 57->80 upgrade path, but that also then requires that .81 not have any additional breaking changes between .80 and .81 (or else the developer must solve everything at once!)

Since this is the lights system, there are existing hacks (https://github.com/missionpinball/mpf/issues/1939, where the hand-channeled example is given) to make RGBWs work on FAST without ER, and RGB and RGBW lights are among the cheapest and most substitutable electronic components in a build, I vote make this .81-only.

Most of the early commits are simple refactors, so it's possible we could backport much of this into .57.4 if desired, but
I'd rather deliver the whole package at once than risk getting some stuff merged in and other stuff languish for a long time.

### DIFF
--- a/mpf/_version.py
+++ b/mpf/_version.py
@@ -10,10 +10,10 @@ PyPI.
 
 """
 
-__version__ = '0.57.4.dev3'  # Also consider whether MPF-MC pyproject.toml should be updated
+__version__ = '0.58.0.dev1'  # Also consider whether MPF-MC pyproject.toml should be updated
 '''The full version of MPF.'''
 
-__short_version__ = '0.57'
+__short_version__ = '0.58'
 '''The major.minor version of MPF.'''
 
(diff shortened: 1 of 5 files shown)


bosh:
I've split this into several stages

PR #1943, opened 2026-01-04
Lights and FAST EXP code cleanup

bosh:
Nonbreaking refactors and cleanup in the lighting code in prep for FAST EXP ER support

### DIFF
--- a/mpf/devices/light.py
+++ b/mpf/devices/light.py
@@ -275,10 +275,7 @@ class Light(SystemWideDevice, DevicePositionMixin):
         if not platform.features['allow_empty_numbers'] and channel['number'] is None:
             self.raise_config_error("Light must have a number.", 1)
 
-        config = LightConfig(
-            name=self.name,
-            color=LightConfigColors[color.upper()]
-        )
+        config = LightConfig(name=self.name, color=LightConfigColors[color.upper()])
 
         try:
             return platform.configure_light(channel['number'], channel['subtype'], config, channel['platform_settings'])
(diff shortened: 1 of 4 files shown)


mpf/devices/light.py
```diff
@@ -275,85 +261,101 @@ def _load_hw_driver(self, channel, color):
...
-            self.config['default_on_color'] = RGBColor(self.config['default_on_color'])
+        # map channels to colors
+        if self.config['type']:
+            color_channels = self.config['type']
+        else:
+            if len(channel_list) == 1:
+                # for one channel default to a white channel
+                color_channels = "w"
+            elif len(channel_list) == 3:
+                # TODO this seems like a bug waiting to happen -- what if the channel list ISNT R G B ordered?
```
avanwinkle, line 310: I think if the channel list isn't RGB ordered then the user should specify a `type` to clarify what the channels are. It's a fair assumption that if the user doesn't specify, we have _some_ type of fallback.

Alternatively, we could make `type` required with a default value of "rgb"

bosh: This case in particular only happens if you dont have a `config[channels]` or a `config[type]`, so I guess really the user is letting the platform defaults do their thing. Since `channel_list` is always from a simple number to channel set generation via `parse_light_number_to_channels`, it's not really a userland bug that could happen here, it's a platform not defining a set of three channels as exactly r g then b.

So maybe the comment is too harsh, and all we need is documentation on the `parse_light_number_to_channels` interface that it must return an rgb-ordered channel list

PR #1951, opened 2026-01-24
fix randomizer result order when using nonrandom draws with conditionals that change between draws

bosh:
Random event players were reported to be sending events in unexpected order when using forced nonrandom draws and conditionals. Ex:

```yaml
random_event_player:
 not_random_bomb_event_qualify:
    events:
      - moblin1_bombed{device.counters.green_moblin1_hit_counter.value < 3}
      - moblin2_bombed{device.counters.green_moblin2_hit_counter.value < 3}
      - moblin3_bombed{device.counters.green_moblin3_hit_counter.value < 3}
    disable_random: true
```

Where the qualify event was triggering the player, which was not necessarily emitting 1 then 2 then 3, due to the hit counters passing or failing their template eval check differently over time.

EG
If I have events [A,B,C] with conditions all true, and I draw once and get A
But then A's condition changes to false, and I draw again. I would expect to get B, but because the valid set shifted by one, and we move to the next index regardless, we skip B (even though it is still valid) and get C instead.

Anyway this fixes all that!
I'll have to hack it apart to fit with the seeded randomizer work though, so that'll be fun.

### DIFF
--- a/mpf/core/randomizer.py
+++ b/mpf/core/randomizer.py
@@ -6,7 +6,7 @@ class Randomizer:
 
     """Generic list randomizer."""
 
-    def __init__(self, items, machine=None, template_type=None):
+    def __init__(self, items, machine=None, template_type='event'):
         """Initialize Randomizer."""
         self.fallback_value = None
         self.force_different = True
@@ -107,25 +107,37 @@ class Randomizer:
             self.force_all = True
 
     def _next_not_random(self, conditional_args):
-        if self.data['current_item_index'] == len(self.items):
+        total_items = len(self.items)
+        current_item_index = self.data['current_item_index']
+        if current_item_index >= total_items:
             if not self.loop:
                 raise StopIteration
 
-            self.data['current_item_index'] = 0
+            current_item_index %= total_items  # or we could just go to 0
 
-        self.data['current_item'] = (
-            self._get_items(conditional_args)[self.data['current_item_index']][0])
+        next_item_index = current_item_index
 
-        self.data['current_item_index'] += 1
+        rotated_option_list = self.items[next_item_index:] + self.items[:next_item_index]
+        while len(rotated_option_list):
+            next_item = rotated_option_list.pop(0)
+            next_item_index += 1
+            event = next_item[0]
+            if not event.condition or (self._template_type and event.condition.evaluate(conditional_args)):
+                break
 
-        return self.data['current_item']
+        self.data['current_item'] = next_item
+        self.data['current_item_index'] = next_item_index
+
+        return next_item[0].name
 
     @staticmethod
     def _init_data(data_dict):
         """Initialize dict."""
         data_dict['current_item'] = None
         data_dict['items_sent'] = set()
-        data_dict['current_item_index'] = 0  # only used with disable random
+
+        # only used with disable random, represents the index in the list _before_ conditional evaluation check
+        data_dict['current_item_index'] = 0
 
     def get_current(self):
         """Return current item."""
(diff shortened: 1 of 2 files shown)


mpf/core/randomizer.py
```diff
@@ -6,7 +6,7 @@ class Randomizer:
 
     """Generic list randomizer."""
 
-    def __init__(self, items, machine=None, template_type=None):
+    def __init__(self, items, machine=None, template_type='event'):
```
bosh, line 9: I wish I didn't have to set a template type here, but in live code we always pass event, and allowing it to be None is either vestigial or just intended to make tests easier to write.

mpf/core/randomizer.py
```diff
@@ -107,25 +107,35 @@ def loop(self, loop):
...
+        while len(rotated_option_list):
+            next_item = rotated_option_list.pop(0)
+            next_item_index += 1
+            event = next_item[0]
+            if not event.condition or (self._template_type and event.condition.evaluate(conditional_args)):
+                break
 
-        return self.data['current_item']
+        self.data['current_item'] = next_item
+        self.data['current_item_index'] = next_item_index
```
bosh, line 129: My instinct was to mod this here, but we actually need to hit the limit (which is the length, ie 1 out of bounds for index lookups) so that the guard at the top of the function stops us looping after this.

mpf/core/randomizer.py
```diff
@@ -107,25 +107,35 @@ def loop(self, loop):
...
+        current_item_index = self.data['current_item_index']
+        if current_item_index >= total_items:
             if not self.loop:
                 raise StopIteration
 
-            self.data['current_item_index'] = 0
+            current_item_index %= total_items  # or we could just go to 0
 
-        self.data['current_item'] = (
-            self._get_items(conditional_args)[self.data['current_item_index']][0])
```
bosh, line 117: Can't rely on get items when considering current_item_index because the index doesnt account for items dropping out due to template, while the whole point of _get_items is to do that evaluation and filtering.

mpf/tests/test_Randomizer.py
```diff
@@ -17,7 +17,7 @@ def get_machine_path(self):
         return 'tests/machine_files/randomizer/'
 
     def test_one_element_with_force_different(self):
-        r = Randomizer(['1'])
+        r = Randomizer(['1'], self.machine)
```
bosh, line 20: Having to add self.machine in every test case sucks but is necessary for some of the template evaluation stuff to keep types more consistent.

mpf/tests/test_Randomizer.py
```diff
@@ -149,12 +149,51 @@ def test_conditionals(self):
...
+                '4{1 == "whatever"}',
+            ],
+                self.machine,
+                template_type="event"
+        )
+        r.disable_random = True
+
+        for i in range(50):
+            self.assertEqual(next(r), '1')
+            self.assertEqual(next(r), '3')
```
bosh, line 167: This would crash on the fifth draw on the previous code.

mpf/tests/test_Randomizer.py
```diff
@@ -149,12 +149,51 @@ def test_conditionals(self):
...
+            ],
+                self.machine,
+                template_type="event"
+        )
+        r.disable_random = True
+
+        for i in range(50):
+            self.assertEqual(next(r), '2')
+            self.assertEqual(next(r), '4')
+
```
bosh, line 186: This draw sequencing is the reported issue from the user. The crash from the test above wasnt reported (I'm guessing they didnt draw four times against their set of three events)

PR #1956, opened 2026-02-04
Update config_spec.yaml

borgdog:
support for mag and min filter for images. sets scaling algorithm options linear or nearest


bosh (approved):
Seems simple enough! Is "nearest" the existing behavior that mc would have experienced before adding support for these options?

bosh:
I had to do an update process to bring this up to date with recent git history, but I fear I may have altered the state of your github fork's `dev` branch. Sorry about that! It appeared to be a commit on top of v0.57.0.dev33

borgdog:
Dang did I put that as nearest? Linear was the default before putting this in. Nearest was the behavior we were looking to get.

PR #1969, opened 2026-02-11
Bernarma - multiball restart grace period (off of dev)

bosh:
Recreated the https://github.com/missionpinball/mpf/pull/1862 PR with the dev branch as the base. Good stuff!

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -1139,6 +1139,7 @@ multiballs:
     shoot_again: single|template_ms|10s
     hurry_up_time: single|template_ms|0
     grace_period: single|template_ms|0
+    restart_grace_period: single|template_ms|0
     ball_locks: list|machine(ball_devices)|None
     enable_events: event_handler|event_handler:ms|None
     disable_events: event_handler|event_handler:ms|None
--- a/mpf/devices/multiball.py
+++ b/mpf/devices/multiball.py
@@ -239,20 +239,20 @@ class Multiball(EnableDisableMixin, SystemWideDevice, ModeDevice):
 
         restart_grace_period_ms = self.config['restart_grace_period'].evaluate([])
 
-        if not self.machine.game or (self.machine.game.balls_in_play - balls < 1 and restart_grace_period_ms==0):
+        if not self.machine.game or (self.machine.game.balls_in_play - balls < 1 and restart_grace_period_ms == 0):
             self._multiball_end()
-        elif restart_grace_period_ms>0:
+        elif restart_grace_period_ms > 0:
             if self.machine.game.balls_in_play == 0:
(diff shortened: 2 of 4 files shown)


mpf/devices/multiball.py
```diff
@@ -237,15 +237,34 @@ def _ball_drain_count_balls(self, balls, **kwargs):
...
+        restart_grace_period_ms = self.config['restart_grace_period'].evaluate([])
+
+        if not self.machine.game or (self.machine.game.balls_in_play - balls < 1 and restart_grace_period_ms == 0):
+            self._multiball_end()
+        elif restart_grace_period_ms > 0:
+            if self.machine.game.balls_in_play == 0:
+                self.delay.run_now("restart_grace_period")
+            elif self.machine.game.balls_in_play - balls < 1:
+                self.machine.events.post(f"multiball_{self.name}_will_end", grace_period=restart_grace_period_ms)
+                '''event: multiball_(name)_will_end
```
bosh: Note for myself, this is a new event that needs documentation

mpf/devices/multiball.py
```diff
@@ -284,6 +303,14 @@ def event_add_a_ball(self, **kwargs):
 
     def add_a_ball(self):
         """Add a ball if multiball has started."""
+        # remove if it exists as multiball now has multiple balls
+        if self.delay.check("restart_grace_period"):
+            self.delay.remove("restart_grace_period")
+            self.machine.events.post(f"multiball_{self.name}_will_resume")
+            '''event: multiball_(name)_will_resume
```
bosh: Note to self, new event to doc

mpf/config_spec.yaml
```diff
@@ -1139,6 +1139,7 @@ multiballs:
     shoot_again: single|template_ms|10s
     hurry_up_time: single|template_ms|0
     grace_period: single|template_ms|0
+    restart_grace_period: single|template_ms|0
```
bosh, line 1142: NTS: New config option to doc

bosh (commented):
Some docs to write in prep for this. Target is `0.57.5.dev1`

Also need to check over event naming conventions and possibly queue type usage (should we!?)

mpf/devices/multiball.py
bosh: I've documented this but I believe this event is misnamed. It should be `multiball_(name)_restart_grace_period_started`.

Based on the naming conventions of `_will_end`/`_ending` in other events, `x_will_y` is generally used as a non-queue event posting immediately before a queue event named `x_ying`. I would also expect multiballs to generally post events named so generically, rather than only if supporting the restart grace period.

mpf/devices/multiball.py
bosh: I think this should be named `multiball_(name)_restarted`

PR #1970, opened 2026-02-12
Implement chimes enable/disable functionality

borgdog:
Added functionality to only enable chimes fired by score reel advances during gameplay, using ball_started and game_ended events.

### DIFF
--- a/mpf/devices/score_reel_group.py
+++ b/mpf/devices/score_reel_group.py
@@ -73,11 +73,10 @@ class ScoreReelGroup(SystemWideDevice):
                                                 handler=self.chime,
                                                 chime=self.config['chimes'][i])
 
-    @classmethod
-    def chime(cls, chime, **kwargs):
+    def chime(self, chime, **kwargs):
         """Pulse chime if chimes are enabled."""
         del kwargs
-        if cls.chimes_enabled:
+        if self.chimes_enabled:
             chime.pulse()
 
     @event_handler(1)
--- a/mpf/devices/score_reel_group.py
+++ b/mpf/devices/score_reel_group.py
@@ -5,6 +5,7 @@ from mpf.core.system_wide_device import SystemWideDevice
 from mpf.devices.score_reel_controller import ScoreReelController
 from mpf.core.events import event_handler
 
+
 class ScoreReelGroup(SystemWideDevice):
 
     """Represents a logical grouping of score reels in a pinball machine.
@@ -72,21 +73,23 @@ class ScoreReelGroup(SystemWideDevice):
                                                 handler=self.chime,
                                                 chime=self.config['chimes'][i])
 
-    def chime(cls, chime, **kwargs):
+    def chime(self, chime, **kwargs):
         """Pulse chime if chimes are enabled."""
         del kwargs
-        if cls.chimes_enabled:
+        if self.chimes_enabled:
             chime.pulse()
 
     @event_handler(1)
     def event_enable_chimes(self, **kwargs):
         """Event handler to enable chimes."""
+        del kwargs
         self.chimes_enabled = True
         self.log.info('Chimes enabled.')
 
     @event_handler(2)
     def event_disable_chimes(self, **kwargs):
         """Event handler to disable chimes."""
+        del kwargs
         self.chimes_enabled = False
         self.log.info('Chimes disabled.')
 
--- a/mpf/devices/score_reel_group.py
+++ b/mpf/devices/score_reel_group.py
@@ -52,7 +52,7 @@ class ScoreReelGroup(SystemWideDevice):
         self.jump_in_progress = False
         # Boolean attribute that is True when a jump advance is in progress.
 
-        self.chimes_enabled = False 
+        self.chimes_enabled = False
         # Boolean attribute that is True when chimes should play when reels advance.
 
         self._tick_task = None
@@ -85,7 +85,7 @@ class ScoreReelGroup(SystemWideDevice):
     def enable_chimes(self, **kwargs):
         self.chimes_enabled = True
         self.machine.events.post('========== CHIMES ENABLED ==========')
-    
+
     def disable_chimes(self, **kwargs):
         self.chimes_enabled = False
         self.machine.events.post('========== CHIMES DISABLED =========')
--- a/mpf/devices/score_reel_group.py
+++ b/mpf/devices/score_reel_group.py
@@ -73,10 +73,11 @@ class ScoreReelGroup(SystemWideDevice):
                                                 handler=self.chime,
                                                 chime=self.config['chimes'][i])
 
-    def chime(self, chime, **kwargs):
+    @classmethod
+    def chime(cls, chime, **kwargs):
         """Pulse chime if chimes are enabled."""
         del kwargs
-        if self.chimes_enabled:
+        if cls.chimes_enabled:
             chime.pulse()
 
     @event_handler(1)
--- a/mpf/devices/score_reel_group.py
+++ b/mpf/devices/score_reel_group.py
@@ -3,7 +3,7 @@ from collections import deque
 
 from mpf.core.system_wide_device import SystemWideDevice
 from mpf.devices.score_reel_controller import ScoreReelController
-
+from mpf.core.events import event_handler
 
 class ScoreReelGroup(SystemWideDevice):
 
@@ -62,9 +62,6 @@ class ScoreReelGroup(SystemWideDevice):
         self.reels = self.config['reels']
         self.reels.reverse()  # We want our smallest digit in the 0th element
 
-        self.machine.events.add_handler(event='ball_started', handler=self.enable_chimes)
-        self.machine.events.add_handler(event='game_ended', handler=self.disable_chimes)
-
         self.config['chimes'].reverse()
         for i in range(len(self.config['chimes'])):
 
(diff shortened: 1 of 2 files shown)


alobascio-ls (approved):
Why can my work account approve this?

mpf/devices/score_reel_group.py
```diff
@@ -70,10 +76,19 @@ async def _initialize(self):
                                                 chime=self.config['chimes'][i])
 
     @classmethod
-    def chime(cls, chime, **kwargs):
-        """Pulse chime."""
+    def chime(self, chime, **kwargs):
```
bosh: So this is a class method for some reason. That means that `self` is misleading as the first param name (the original was `cls` representing the class, which is new to me but seems a reasonable pattern). We care because `chimes_enabled` is a property on an instance, not the class itself.

The red builds on CI are blowing up on 82, which is due to this.

I'm not sure off the top of my head what the correct solution is -- IMO `chime` should be an instance method, not a class method, then it would all Just Work. But if there's a specific preexisiting reason that it's class instead, then instead we need to find an actual instance and ask it if `instance.chimes_enabled` instead.

mpf/devices/score_reel_group.py
```diff
@@ -70,10 +76,19 @@ async def _initialize(self):
...
+        """Pulse chime if chimes are enabled."""
         del kwargs
-        chime.pulse()
+        if self.chimes_enabled:
+            chime.pulse()
+
+    def enable_chimes(self, **kwargs):
```
bosh: This function needs to either be made implicitly private (add a prefix _ to the name, so `def _enable_chimes(...` and also update the callsite/event binding) or add a docstring (look after the definition of set_value.) The linter is super picky about the syntax, but it's basically: start with three "s and a capital letter and the first line _must_ end with a period. (and three more "s if the string is single line). If multiline, I believe a blank line is mandatory, then the rest of the content, ending with 3 "s on their own line.

Usually I'd say introduce the function as a private unless it needs to be used in a public python context, but enable/disable on an internal setting are pretty reasonable to have public. The docstring can be as simple as:
`"""Enable chimes for the score reel group."""`

mpf/devices/score_reel_group.py
```diff
@@ -70,10 +76,19 @@ async def _initialize(self):
...
+        self.chimes_enabled = True
+        self.log.info('Chimes enabled.')
+
+    def disable_chimes(self, **kwargs):
```
bosh: Ditto previous function comment

mpf/devices/score_reel_group.py
```diff
@@ -52,13 +52,19 @@ def __init__(self, machine, name):
...
+
         self._tick_task = None
 
     async def _initialize(self):
         await super()._initialize()
         self.reels = self.config['reels']
         self.reels.reverse()  # We want our smallest digit in the 0th element
 
+        self.machine.events.add_handler(event='ball_started', handler=self.enable_chimes)
+        self.machine.events.add_handler(event='game_ended', handler=self.disable_chimes)
```
bosh: I think these are reasonable but there may be other events worth listening for by default. Alternatively we could take the approach of providing config options for enable events and disable events (such as flipper enable/disable events: https://github.com/missionpinball/mpf/blob/e50131c0eefd0c9d06cbcaa845f30c03bb0cc969/mpf/devices/flipper.py#L123 and https://github.com/missionpinball/mpf/blob/e50131c0eefd0c9d06cbcaa845f30c03bb0cc969/mpf/config_spec.yaml#L772 ).

In this case though, it's not enable and disable on the score reel group itself, so the config lines to add would be `enable_chimes_events: event_handler|event_handler:ms|whatever_default_enable_event_name,ball_started,etc_event_name` which MPF will automatically wire up to a function defined as: `def event_enable_chimes(self, **kwargs):`

But with these config options provided, along with defaults, then it's a non-breaking upgrade that allows developers to customize their chime management with no further complexity added to MPF.

bosh: The config additions also make this a bit more testable -- we can have a test in https://github.com/missionpinball/mpf/blob/e50131c0eefd0c9d06cbcaa845f30c03bb0cc969/mpf/tests/test_ScoreReels.py#L9C4-L10C1 .... nevermind. No tests!

But actually, I guess that's what I'll be doing in support of this work, is trying to make that test work again. I have some ideas :)

bosh (commented):
I think the bones are good, but there's a bit more to do. Make sure you look over the CI builds for failures ( https://github.com/missionpinball/mpf/actions/runs/21930481476/job/63333128484?pr=1970#step:6:194 ). I think the comment on line 80 will address it.

Also take a look at the prospector run once all the other actions are green - https://github.com/missionpinball/mpf/actions/runs/21930481481/job/63333128300?pr=1970 is the lint errors from the latest run. It's not usually worth checking this until everything else is buttoned up.

mpf/devices/score_reel_group.py
bosh: Turns out the comment was a lie! Brian actually made it work 8 months later and forgot to clean it up!

Okay well so the test changes I would want to see are, delete the outdated comment (8-19 can completely go), then add a test that has a score reel group configured with custom enable and disable chime events, then manually post the events and then make the chime handler fire. Then either assert that the pulse did or didnt happen (or if this is hard, just assert that the score reel group instance's chimes_enabled property is True or False)

mpf/devices/score_reel_group.py
bosh: Thanks for the changes, I'll look into whether we can leave this as an instance method

mpf/devices/score_reel_group.py
```diff
@@ -69,11 +73,25 @@ async def _initialize(self):
...
+            chime.pulse()
+
+    @event_handler(1)
+    def event_enable_chimes(self, **kwargs):
+        """Event handler to enable chimes."""
+        del kwargs
+        self.chimes_enabled = True
+        self.log.info('Chimes enabled.')
+
+    @event_handler(2)
```
bosh, line 89: Is it intentional that the two event handlers are registered at priority 1 and 2? It's okay if no, and these two numbers are probably reasonable picks.

It took me a really long time to understand this btw, and I'm actually not sure what the best numbers to pick are. Anyway, so the # in the parens is used as a priority modifier for the event handlers created via the config options that bind to this.

As in, `enable_chimes_events: ball_started` will look up the `@event_handler` number (1 for that one) and effectively register a hook -- if we wrote it as an event player using dot-priority it would end up as:
``` yaml
event_players:
  ball_started.1: <Python: self.event_enable_chimes>
  game_ended.2: <Python: self.event_disable_chimes>
```

Which, like, who cares?

I _think_ the issue is that a developer might choose the same event for both (why? who knows. maybe it's more common with other device types). So if they do, then not only have they done something a bit daft, but they also end up with a handler resolution order ambiguity.

So I guess (I'm honestly just putting all of this together for the first time myself) we _must_ use different numbers, and we have 1 and 2 here, so that's fine. This means that the disable handler goes first, at priority 2. Which actually means it loses, because it will run, set enabled to False, then the enable handler will catch at priority 1 and run, setting it back to True! So a bit of inversion of expectation around priority and final state.

So really what you have is a decision to make; if a dev really made the event the same for both, since we dont detect and error for them, would you rather have the current situation, where `chimes_enabled == True`, or would you like to flip the 1 and 2 and have the tiebreak end with `chimes_enabled == False`?

It probably doesn't actually matter, since the situation is a bad game code design, but in general this must be why `@event_handler(priority)` exists. Real time learnings!

bosh (changes requested):
This is looking really great!

I still need to check in on the `@classmethod` usage, but it's probably fine.
I also have a minor nit and note, but other than that I believe the code here is exactly what I was looking for. The only other thing left is a structure test and some behavior testing.

For the structure test, I mean adding yaml to a config file that will ensure that the config_spec.yaml matches how you envision a game developer would use the feature.

It's probably easiest to just extend the score reel test. The config file for the test lives at:
https://github.com/missionpinball/mpf/blob/dev/mpf/tests/machine_files/score_reels/config/config.yaml

and the test python can be found at: https://github.com/missionpinball/mpf/blob/dev/mpf/tests/test_ScoreReels.py .

I'd ask that you delete the comment from line 8-19 (it's outdated and incorrect, I checked) and then add a score_reel_group or two to the config.yaml. Actually, I just looked, and there are already two. What I'd probably do to test all the structural variants is make the first not use the settings (therefore uses defaults, proving the config option is not required. You basically get this test value for free :) ), then I'd make one of the other two have custom events.

Then in the python test code, add a new function `def testWhatever(self):` that fires off these events and asserts something about the state of the program before and after each event. It looks like existing tests have mocked out the coil pulse functions, and they can actually directly assert on how many pulses have been fired. `testAdvanceingFailure` has pretty concise and good examples of everything I think you would need. I would probably only test custom events on one instance, and not test the default events specifically (IMO seeing the default value written in config_spec.yaml is enough).

If you haven't been running local tests yet, you can shorten your development loop via running the full suite or a targeted test on the command line - https://missionpinball.org/latest/code/Writing_Tests/RunUnitTests/#running-all-tests-at-once

bosh:
Oh and of course bonus points always for docs :) https://github.com/missionpinball/mpf-docs/blob/main/docs/config/score_reel_groups.md?plain=1

mpf/devices/score_reel_group.py
borgdog: ha, no intention on my part one way or the other, just copying other mode. didn't know that's what that did, but makes sense now that you explain it

bosh: Would you believe I kept thinking about this?... I think MPF should have a preference for when two handlers could want to do conflicting things, and I think I know which I prefer:

Think about weak flippers as a feature - you have the same flipper button and coil defined as two separate `flippers:` configs, with enable and disable events that allow you to swap behaviors. The enable events for the standard flipper are the disable events for the weak flipper and visa versa.
Coils maintain a "current rule" applied, so if the event "enable_weak_flippers" posts, you obviously want the weak flipper rule at the end. So the right order of operations in handlers is: handle the disable first, then handle the enable. And that means disable is higher priority as a rule. Bam. We did it.

bosh:
Per @borgdog's request for backup, I've cloned this branch and added a test commit, PR over here https://github.com/missionpinball/mpf/pull/1972

PR #1973, opened 2026-02-15
UnRARed-bug/mypinballs broken segment display command fix

bosh:
Incorporates and fixes tests for https://github.com/missionpinball/mpf/pull/1880 by @unRARed

### DIFF
--- a/mpf/platforms/mypinballs/mypinballs.py
+++ b/mpf/platforms/mypinballs/mypinballs.py
@@ -27,8 +27,6 @@ class MyPinballsSegmentDisplay(SegmentDisplayPlatformInterface):
             # remove any non-numbers and spaces
             text = re.sub(r'[^0-9 ]', "", text)
 
-            # special char for spaces
-            text = text.replace(" ", "?")
             # set text
             if flashing == FlashingType.FLASH_ALL:
                 cmd = b'2:'
--- a/mpf/tests/test_MyPinballs.py
+++ b/mpf/tests/test_MyPinballs.py
@@ -107,7 +107,7 @@ class MyPinballsPlatformTest(MpfTestCase):
 
     def testPlatform(self):
         self.serialMock.expected_commands = {
-            b'1:1:???1234': False,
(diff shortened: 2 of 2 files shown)


bosh:
However, this thread seems to imply that the ? is actually the correct delimiter for the unit the author over there is using. Hmm https://groups.google.com/g/mpf-users/c/h-fVEUS712Q

unRARed:
Thanks @bosh. I haven't had a chance to get back to this, and we've just been using the patches on our game. It was started in 2018 and it was mostly tabled until 2024. Much of our hardware decisions were seeming antiquated (My Pinballs segment controller and Cypress OPP boards).

unRARed:
> However, this thread seems to imply that the ? is actually the correct delimiter for the unit the author over there is using. Hmm https://groups.google.com/g/mpf-users/c/h-fVEUS712Q

I wonder if there was maybe different versions of the firmware then? Pretty sure it was just an Arduino Uno he sent us and a wiring diagram. Our game would not boot until that line was removed.

unRARed:
If it helps, this was the only documentation we got from the author:

Screenshot 2026-02-15 at 11 34 36 AM
Screenshot 2026-02-15 at 11 35 06 AM

bosh:
@unRARed do your segment displays have the multicolor-with-A-Z support seen in that thread or are they an older unit? The older docs material only mentions scores, not text with spaces. It's possible the protocol did change if space character (" ") support was added later and things needed be to disambiguated between text spaces and protocol separators.

unRARed:
> @unRARed do your segment displays have the multicolor-with-A-Z support seen in that thread or are they an older unit? The older docs material only mentions scores, not text with spaces. It's possible the protocol did change if space character (" ") support was added later and things needed be to disambiguated between text spaces and protocol separators.

Ours are just single color - only numeric, so yea, probably a different firmware. I had forwarded my PR to the My Pinballs guy asking about it, but I never received a reply.

bosh:
Closed per mypinball's rec in original PR

PR #2018, opened 2026-05-09
Fix EventReferenceParser for Python 3.14 (ast.Str → ast.Constant)

Pyrrvs:
## Summary

- Python 3.14 removed the deprecated `ast.Str` node in favour of `ast.Constant`. `EventReferenceParser` was still inspecting `ast.Str`, so any project that ran on 3.14 hit an `AttributeError` during config validation.
- Replaces `ast.Str` with `ast.Constant` and adjusts the value extraction to use `.value` instead of `.s`.

## Test plan

- [x] Existing MPF test suite passes on Python 3.14 with the fix applied
- [x] Verified locally that machine config loads on Python 3.14

### DIFF
--- a/mpf/parsers/event_reference_parser.py
+++ b/mpf/parsers/event_reference_parser.py
@@ -31,15 +31,19 @@ class EventReferenceParser:
                 class_label = None
                 config_section = None
                 for statement in x.body:
-                    if isinstance(statement, ast.Assign) and len(statement.targets) == 1 and \
-                       isinstance(statement.targets[0], ast.Name) and isinstance(statement.value, ast.Str):
+                    if (isinstance(statement, ast.Assign)
+                            and len(statement.targets) == 1
+                            and isinstance(statement.targets[0], ast.Name)
+                            and isinstance(statement.value, ast.Constant)
+                            and isinstance(statement.value.value, str)):
                         if statement.targets[0].id == "class_label":
-                            class_label = str(statement.value.s)
+                            class_label = str(statement.value.value)
                         elif statement.targets[0].id == "config_section":
-                            config_section = [str(statement.value.s)]
+                            config_section = [str(statement.value.value)]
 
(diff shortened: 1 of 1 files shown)


bosh:
I approved the actions run. Could you please re target the base of this PR to `dev`. The versioned .x branches are for release tagging.

Pyrrvs:
Retargeted to `dev` and rebased onto the current tip. Thanks for the review!

bosh:
I'm inclined to approve (after linter fixes), but I don't see anywhere this code is used -- did you run into an issue because of this?

Pyrrvs:
Yes, I should have included that in the PR, my bad. I was trying to port mpf-ls to the latest MPF version when I first ran into this issue. I am no longer actively trying to port mpf-ls to the latest mpf version but figured that it was still an interesting fix.

I fixed the issues reported by the linter.

mpf/parsers/event_reference_parser.py
```diff
@@ -31,15 +31,19 @@ def _parse_file(self, file_name) -> List[EventReference]:
                 class_label = None
                 config_section = None
                 for statement in x.body:
-                    if isinstance(statement, ast.Assign) and len(statement.targets) == 1 and \
-                       isinstance(statement.targets[0], ast.Name) and isinstance(statement.value, ast.Str):
+                    if (isinstance(statement, ast.Assign)
+                            and len(statement.targets) == 1
+                            and isinstance(statement.targets[0], ast.Name)
+                            and isinstance(statement.value, ast.Constant)
+                            and isinstance(statement.value.value, str)):
```
bosh, line 38: I'm slightly suspicious of the added type check here that the final value is a string, but I think it is correct.

bosh (approved):
I'm going to approve and merge even though I'm a little uncertain, because the unmainted LSP seems to be the only usage of this class anyway.

PR #2019, opened 2026-05-09
Add vpx_reset BCP subcommand for VPX-driven state reset

Pyrrvs:
## Summary

Adds a new `vpcom_bridge?subcommand=reset` handler to the `virtual_pinball` platform. When VPX (re)connects, it can ask MPF to wipe its switch mirror and run `machine.reset()` so the next session starts on a clean machine.

This solves a real dev-loop pain when iterating with VPX: today, MPF state (in-progress game, score, mode stack, stuck switches) leaks across VPX restarts. With this change, every (re)connect is automatically clean.

### Consumer

The new subcommand is consumed by [Pyrrvs/mpf-vpx-plugin](https://github.com/Pyrrvs/mpf-vpx-plugin) — a Visual Pinball X plugin (C++ shared library, dropped into VPX's plugin directory) that bridges VPX's VBS environment to MPF over BCP. The plugin replaces the legacy `vpinmame.controller` COM object, lets a designer's existing VPX table drive MPF as if it were the physical machine, and gives MPF reverse control of coils, lights, segment displays, and hardware rules.

The companion plugin PR ([Pyrrvs/mpf-vpx-plugin#1](https://github.com/Pyrrvs/mpf-vpx-plugin/pull/1)) adds connect-retry, mid-session reconnect, and a switch-state mirror that gets replayed after `vpx_reset` so the playfield state survives the wipe.

### What `vpx_reset` does

`vpx_reset` is an `async` handler that:
1. Iterates the platform's switch mirror and forces every active switch back to inactive (so reset-phase handlers see a cleaned state instead of stale readings from the prior VPX session).
2. `await self.machine.reset()` — drives `machine_reset_phase_1/2/3` and `reset_complete`.
3. Returns `True` so the BCP reply contains a `result` (no `error`).

Mechanisms (servos, motors, drop targets, diverters) reset automatically because they already subscribe to `machine_reset_phase_3`. Switch state on the plugin side is replayed by the plugin **after** the reset so the new MPF immediately knows the playfield state (which switches are active right now).

### Dispatcher tweak

`_dispatch` now awaits coroutine handlers — backward-compatible with all existing sync `vpx_*` handlers (the `inspect.iscoroutine` check is a no-op for non-coroutine returns). This is what allows `vpx_reset` to be `async`.

### Note about the first commit

The first commit on this branch is the Python 3.14 compatibility fix (`ast.Str` → `ast.Constant`), already submitted as #2018. Either PR can merge first; if #2018 lands first I'll rebase this PR.

## Test plan

- [x] `test_vpx_reset_clears_switches_and_resets_machine` — pins switch mirror wipe + `machine_reset_phase_3` posting.
- [x] `test_vpx_reset_ends_active_game` — pins in-progress-game-ends behavior (skipped in the VPX test fixture because it has no ball-source device, but the logic is exercised end-to-end via the broader machine.reset() suite).
- [x] Full MPF test suite passes on Python 3.14 (882/882).
- [x] End-to-end smoke tested against the companion `mpf-vpx-plugin` change (5 scenarios: cold connect, MPF-first connect, VPX restart mid-game, MPF restart mid-game with switch replay, graceful degradation).

### DIFF
--- a/mpf/platforms/virtual_pinball/virtual_pinball.py
+++ b/mpf/platforms/virtual_pinball/virtual_pinball.py
@@ -1,5 +1,6 @@
 """VPX platform."""
 import asyncio
+import inspect
 from typing import Dict, List, Optional
 
 import logging
@@ -200,6 +201,8 @@ class VirtualPinballPlatform(LightsPlatform, SwitchPlatform, DriverPlatform, Seg
 
         try:
             result = method(**kwargs)
+            if inspect.iscoroutine(result):
+                result = await result
         # pylint: disable-msg=broad-except
         except Exception as e:
             self.machine.bcp.transport.send_to_client(client, "vpcom_bridge_response",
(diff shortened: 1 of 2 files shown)


bosh:
I approved the actions run. Could you please re target the base of this PR to `dev`. The versioned .x branches are for release tagging.

Pyrrvs:
Retargeted to `dev` and rebased onto the current tip. Also dropped the python314 fix commit from this branch (it lives in #2018 now), so this PR is just the `vpx_reset` changes.

bosh (approved):
Nice! Great test too! The only thing I find funny is that the bcp encoded message is just "reset" instead of "vpx_reset", but I see that that's the way every other bcp message is handled as well.

PR #2023, opened 2026-05-18
Reset available_balls count on eject timeouts

avanwinkle:
## Summary

This PR attempts to mitigate a bug in ball count handling where orphaned `available_balls` causes the game to soft lock.

## Description

When a ball device is the eject target of another's ejection, the `available_balls` count is incremented to allow eject chains to map a path through the available devices to the target. When the eject completes, the incoming device's `balls` count is incremented and the available balls are resolved.

There is a bug where if an eject fails, the available_balls count is not adjusted. This can lead to disparate values between `available_balls` and `balls`.

When a player's turn ends, the game waits for all balls to drain before proceeding. However it does this by counting `available_balls` and not `balls`, which can lead to a soft-lock if a previous eject failure caused the device (playfield) to have orphaned available balls.

## The Fix

This PR implements two changes:
* When an eject fails, decrement the `available_balls` of the target device. This keeps the balls and available balls in sync, and the available balls will be incremented again on the next eject attempt.
* When the playfield drains, do a safety check to align the available balls and balls. This should no longer happen with the above fix, but it's better to catch and address a mismatch than to soft-lock the game.

These fixes were build by Copilot based on analysis of my own game log wherein a multiball eject failure caused a later downstream lock. Tests were added to validate the fix.

balls balls balls

### DIFF
--- a/mpf/core/ball_controller.py
+++ b/mpf/core/ball_controller.py
@@ -188,6 +188,15 @@ class BallController(MpfController):
         while True:
             found_balls = False
             for playfield in playfields:
+                # Reconcile inconsistent playfield state: available_balls may
+                # sometimes be left >0 due to eject/timeout race conditions.
+                # Use the authoritative `balls` count and keep `available_balls`
+                # in sync to avoid waiting forever on stale state.
+                if playfield.available_balls != playfield.balls:
+                    self.debug_log('Reconciling %s: available_balls=%s, balls=%s',
+                                   playfield.name, playfield.available_balls, playfield.balls)
+                    playfield.available_balls = playfield.balls
+
                 if playfield.available_balls > 0:
                     self.info_log('Found %s ball(s) on %s.', playfield.available_balls, playfield.name)
                     found_balls = True
--- a/mpf/devices/ball_device/outgoing_balls_handler.py
+++ b/mpf/devices/ball_device/outgoing_balls_handler.py
@@ -357,12 +357,12 @@ class OutgoingBallsHandler(BallDeviceStateHandler):
             else:
                 self.debug_log('No available_balls to decrement on %s (value: %s)',
                                target.name, getattr(target, 'available_balls', None))
-        except Exception:
+        except Exception:  # pylint: disable=broad-exception-caught
             # Don't let cleanup raise an exception
             try:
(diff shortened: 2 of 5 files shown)


avanwinkle:
Working through the test failures, hard to tell if it's an actual regression or a byproduct of the test cases compensating for the available_balls bug. Will have more updates later this week.

bosh:
Interesting -- @borgdog's SW soft locks on occasion in between balls (I think while trying to stage the next ball) and my theory was that the start of the just-drained ball mightve had a failed eject or other unusual state happen much earlier on, but the problem would only manifest at start of next ball.

I've PMed him the branch, maybe we can see if it fixes it for NWPAS or at least some test runs before :)

IIRC the setup was that the trough eject device was stuck in `waiting_for_target_ready` while the plunger was in `waiting_for_ball`

One thing I notice is that these are both "outgoing_balls_handler" states, so maybe the plunger is stuck waiting for a signal, and the trough is waiting for that to clear before ejecting

borgdog:
I manually added this to my machine and have been playing it a bunch.  Unfortunately I have never been able to reliably recreate the issue, and it pretty much exclusively happens at shows. That said I have not had the issue, but did notice one occasion where there was a pause after bonus count before next ball, normally it kicks out almost immediately, and I thought  it was going to have the issue, but after about 2 seconds it launched a ball.  I don't know if that could be a result of this PR, but thought I'd note it.  Game will again be at NWPAS show in a couple weeks where I first had the problem last year.

cobra18t:
> I manually added this to my machine and have been playing it a bunch. Unfortunately I have never been able to reliably recreate the issue, and it pretty much exclusively happens at shows. That said I have not had the issue, but did notice one occasion where there was a pause after bonus count before next ball, normally it kicks out almost immediately, and I thought it was going to have the issue, but after about 2 seconds it launched a ball. I don't know if that could be a result of this PR, but thought I'd note it. Game will again be at NWPAS show in a couple weeks where I first had the problem last year.

How long have you waited for soft locks in the past? Unpredictable delays between balls or games may also be a memory leak. I was accidentally creating thousands of shows that were stacking on top of each other and were not terminated until the end of a game. This led to long delays between games. You might set this up before the NWPAS show:
https://missionpinball.org/latest/troubleshooting/debugging_memory_leaks/

PR #2024, opened 2026-05-19
Update PinProc

atummons:
Removed support for Python 3.8 and 3.9 (not supported on 0.80)
Added support for Python 3.13 and 3.14


bosh (approved):
Nice! Merging to dev, which would go for 0.81 release in probably a few months.

Do you think this is worth making a 0.80.1 backport for? I'm happy to do all the git-fu if you think this warrants faster distribution

atummons:
I think it's fine to just go with what's next. A person asked for it, so I built them. They can go on dev branch to get it, would be my opinion. Especially if you're thinking there's a general timeframe for the upgrade.

PR #2028, opened 2026-05-25
Soft power down event hook and FAST button press integration

bosh:
This adds soft power down requesting as well as post-exit command support.

Flow 1: FAST soft power:
Note: watchdog runs double frequency vs config value, for safety. This commit also doubles the frequency, to improve resolution. This means that after a state change, there will be up to .25 seconds extra measured at the release moment. Though I suppose this also happens in reverse at the start moment. Oh well.

WD is constantly issued to the NET connection, normally receiving P response (WD:P)
If soft power button is pressed, response is WP:P

Commit adds processors for both messages, plus time tracking and resetting.
Note that the FAST system will auto power down on 7s hold, so the threshold should be comfortably lower than that to avoid accidental overpress. (Picked 2.5 default, which tbh feels a touch slow)

If time between press and release was long enough, asks the machine to attempt a soft power down.

Machine generically requests request_shutdown, allowing other hooks to reject the boolean event and keep the machine alive.
Nothing by default will hook in, so this should noop right into the handler, which marks the soft power down intention and signals the stop future to uh, stop.

The machine shutdown hooks into platforms, which means the neuron receives a stopping() call. Because the soft power off could be event-based, it checks whether it was generally a soft power down (not a fast-specific one) and attempts a WP:1000, which is a 4096ms delayed soft power down of the neuron. That 4s could probably be a setting, since users might need the neuron to power down dead last. If the neuron is _not_ on soft power, WP:XXXX will not do anything. (Well, it wont de-power, I need to check with the FAST devs on whether there are other side effects)

Flow 2: Event-based
Instead of using the neuron dedicated pathway, this introduces an event that attempts the shutdown generically, with the same request boolean event and followup. Just post "attempt_soft_power_down" :)

~~~~


I want to get some testing on this and reconsider some of the naming. Plus maybe testing, though all of this resides in the realm of hard-to-test.

This allows a dev to fire the event `attempt_soft_power_down` OR hold the FAST soft power button down for 2.25001 seconds to initiate soft power down.

Ways MPF can exit:
1. python issues in MPF startup
2. runtime python issues
3. config validation issues
4. Ctrl C (user kill)
5. hardware disconnection (port disconnection)
6. BCP client issues quit (godot esc)
7. soft power event hook
8. soft power hardware press

I've tested all these cases, and found that the exit command fires only as expected, for soft power event or button hold.

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -581,6 +581,8 @@ fast_net:
     mute_unconfigured_switches: single|bool|false
     gi_hz: single|int|30
     lamp_hz: single|int|30
+    soft_power_hold_ms: single|ms|2500
+    soft_power_powerdown_delay: single|ms|4000
 fast_io_loop:
     model: single|str|  # FP-I/O-0804, etc.
     order: single|enum(1,2,3,4,5,6,7,8,9)|
@@ -942,6 +944,7 @@ machine:
     __type__: config
     balls_installed: single|int|1
     min_balls: single|int|1
+    soft_shutdown_exit_command: single|str|None
 machine_vars:
     __valid_in__: machine
     __type__: config_dict
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -573,7 +573,7 @@ fast_net:
     console_log: single|enum(none,basic,full)|none
     file_log: single|enum(none,basic,full)|basic
     controller: single|enum(neuron,nano,sys11,wpc89,wpc95)|
-    watchdog: single|ms|1000
+    watchdog: single|ms|500
     default_quick_debounce_open: single|ms|2
     default_quick_debounce_close: single|ms|2
     default_normal_debounce_open: single|ms|4
--- a/mpf/core/machine.py
+++ b/mpf/core/machine.py
@@ -2,6 +2,7 @@
 """Contains the MachineController base class."""
 import asyncio
 import logging
+import subprocess
 import sys
 import threading
 from typing import Any, Callable, Dict, List, Set, Optional, TYPE_CHECKING
@@ -106,7 +107,8 @@ class MachineController(LogMixin):
                  "stop_future", "events", "switch_controller", "mode_controller", "settings",
                  "bcp", "ball_controller", "show_controller", "placeholder_manager", "device_manager", "auditor",
                  "tui", "service", "switches", "shows", "coils", "ball_devices", "lights", "playfield", "playfields",
-                 "autofire_coils", "_crash_handlers", "__dict__", "mpf_config", "is_shutting_down"]
+                 "autofire_coils", "_crash_handlers", "__dict__", "mpf_config", "is_shutting_down",
+                 "soft_power_down_active"]
 
     # pylint: disable-msg=too-many-statements
     def __init__(self, options: dict, config: MpfConfig) -> None:
@@ -116,6 +118,7 @@ class MachineController(LogMixin):
         self.log.info("Mission Pinball Framework Core Engine v%s", __version__)
         self._crash_handlers = []   # type: List[Callable]
         self.is_shutting_down = False
+        self.soft_power_down_active = False
 
         self.log.info("Command line arguments: %s", options)
         self.options = options
@@ -396,6 +399,7 @@ class MachineController(LogMixin):
     def _register_system_events(self) -> None:
         """Register default event handlers."""
         self.events.add_handler('quit', self.stop)
+        self.events.add_handler('request_soft_shutdown', self.request_soft_shutdown)
         self.events.add_handler(self.config['mpf']['switch_tag_event'].
                                 replace('%', 'quit'), self.stop)
 
@@ -790,6 +794,56 @@ class MachineController(LogMixin):
         self.clock.loop.stop()
         self.clock.loop.run_forever()
         self.clock.loop.close()
+
+        if self.soft_power_down_active:
+            self._execute_soft_shutdown_command()
+
+    def _execute_soft_shutdown_command(self) -> None:
+        cmd = self.config['machine'].get('soft_shutdown_exit_command')
+        if not cmd:
+            return
+
+        self.info_log('Executing exit command %s', cmd)
+        try:
+            # pylint: disable=consider-using-with
+            subprocess.Popen(
+                cmd,
+                shell=True,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                start_new_session=True
+            )
+        except (subprocess.SubprocessError, OSError) as e:
+            self.error_log("Failed to execute exit command: %s", e)
+
+    def request_soft_shutdown(self, **_kwargs) -> None:
+        """Attempt to soft shut down, allowing hooks to deny the request."""
+        self.warning_log("Attempting soft shutdown")
+        self.events.post_boolean('machine_request_shutdown', callback=self._result_of_shutdown_request)
+
+    def _result_of_shutdown_request(self, ev_result=True):
+        """Handle the result of the shutdown request.
+
+        Called after the *machine_request_shutdown* event is posted.
+
+        If the boolean result is True, the stop future will be resolved (ending the run loop)
+        and the event 'machine_will_shutdown' is posted. If the result is False,
+        the event 'machine_abort_shutdown' is posted.
+
+        Args:
+        ----
+            ev_result : Bool result of the boolean event
+                *machine_request_shutdown.* If any registered event handler did not
+                want the machine to shut down, this will be False, else it's True.
+        """
+        if ev_result is False:
+            self.warning_log('Soft shutdown was denied')
+            self.events.post('machine_abort_shutdown')
+        else:
+            self.warning_log('Soft shutdown proceeding...')
+            self.events.post('machine_will_shutdown')
+            self.soft_power_down_active = True
+            self.stop('Soft shutdown')
 
     def _run_loop(self) -> None:    # pragma: no cover
         # Main machine run loop with when the default platform interface
--- a/mpf/platforms/fast/communicators/net_neuron.py
+++ b/mpf/platforms/fast/communicators/net_neuron.py
@@ -271,10 +271,21 @@ class FastNetNeuronCommunicator(FastSerialCommunicator):
         await self.send_and_wait_for_response_processed('SA:', 'SA:')
 
     def _process_wp(self, msg):
-        return
+        if msg == 'P':
+            if not self.platform.soft_power_held_time:
+                self.platform.soft_power_held_time = self.machine.clock.get_time()
+                self.info_log('Soft power down press started.')
+                self.machine.events.post('fast_soft_power_switch_active')
 
     def _process_wd(self, msg):
-        return
+        if msg == 'P':
+            if self.platform.soft_power_held_time:
+                delta_seconds = self.machine.clock.get_time() - self.platform.soft_power_held_time
+                self.machine.events.post('fast_soft_power_switch_inactive')
+                self.platform.soft_power_held_time = None
+                self.info_log('Soft power down released after %s seconds.', delta_seconds)
+                if delta_seconds * 1000 > self.platform.soft_power_hold_ms:
+                    self.platform.report_soft_power_down_request()
 
     def _process_sa(self, msg):
         if not self.platform.switches_initialized:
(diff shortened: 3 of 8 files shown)


mpf/core/machine.py
```diff
@@ -106,7 +107,7 @@ class MachineController(LogMixin):
                  "stop_future", "events", "switch_controller", "mode_controller", "settings",
                  "bcp", "ball_controller", "show_controller", "placeholder_manager", "device_manager", "auditor",
                  "tui", "service", "switches", "shows", "coils", "ball_devices", "lights", "playfield", "playfields",
-                 "autofire_coils", "_crash_handlers", "__dict__", "mpf_config", "is_shutting_down"]
+                 "autofire_coils", "_crash_handlers", "__dict__", "mpf_config", "is_shutting_down", "soft_power_down_active"]
```
bosh: machine state variable: soft_power_down_active

mpf/core/machine.py
```diff
@@ -396,6 +398,7 @@ def validate_machine_config_section(self, section: str) -> None:
     def _register_system_events(self) -> None:
         """Register default event handlers."""
         self.events.add_handler('quit', self.stop)
+        self.events.add_handler('attempt_soft_power_down', self.attempt_soft_power_down)
```
bosh: event: attempt_soft_power_down

mpf/core/machine.py
```diff
@@ -786,6 +789,40 @@ def shutdown(self) -> None:
         self.clock.loop.run_forever()
         self.clock.loop.close()
 
+        if self.soft_power_down_active and self.config['machine'].get('exit_command'):
```
bosh: machine: config setting: exit_command

mpf/core/machine.py
```diff
@@ -786,6 +789,40 @@ def shutdown(self) -> None:
...
+            cmd = self.config['machine']['exit_command']
+            self.info_log('Executing exit command %s', cmd)
+            try:
+                subprocess.Popen(cmd, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
bosh: Probably go somewhere other than dev null huh?

mpf/core/machine.py
```diff
@@ -786,6 +789,40 @@ def shutdown(self) -> None:
...
+            except Exception as e:
+                self.error_log("Failed to execute exit command: %s", e)
+
+    def attempt_soft_power_down(self, **kwargs) -> None:
+        self.warning_log("Attempting soft shutdown")
+        self.events.post_boolean('request_shutdown', callback=self.result_of_shutdown_request)
```
bosh: Note that nothing actually uses this, it's just an available hook

mpf/core/machine.py
```diff
@@ -786,6 +789,40 @@ def shutdown(self) -> None:
...
+        request was denied by some handler.
+
+        Args:
+        ----
+            ev_result : Bool result of the boolean event
+                *request_shutdown.* If any registered event handler did not
+                want the machine to shut down, this will be False, else it's True.
+        """
+        if ev_result is False:
+            self.debug_log("Soft shutdown was denied")
```
bosh: warnings probably

mpf/core/machine.py
```diff
@@ -786,6 +789,40 @@ def shutdown(self) -> None:
...
+        else:
+            self.debug_log("Soft shutdown proceeding...")
+            self.soft_power_down_active = True
+            self.stop_future.set_result(True)
```
bosh: seemed like the best way to hook into the normal shutdown process

mpf/platforms/fast/communicators/net_neuron.py
```diff
@@ -22,7 +23,7 @@ class FastNetNeuronCommunicator(FastSerialCommunicator):
     MAX_IO_BOARDS = 9
     MAX_SWITCHES = 104
     MAX_DRIVERS = 48
-    IGNORED_MESSAGES = ['WD:P', 'TL:P']
+    IGNORED_MESSAGES = ['TL:P']
```
bosh, line 25: must actually handle these now

mpf/platforms/fast/communicators/net_neuron.py
```diff
@@ -268,6 +271,19 @@ async def update_switches_from_hardware(self):
         """Process an SA: message and update switch states."""
         await self.send_and_wait_for_response_processed('SA:', 'SA:')
 
+    def _process_wp(self, msg):
+        if not self.platform.soft_power_held_time:
+            self.platform.soft_power_held_time = time.time()
+            self.warning_log('Soft power down press started.')
```
bosh: could be just info?

mpf/platforms/fast/communicators/net_neuron.py
```diff
@@ -268,6 +271,19 @@ async def update_switches_from_hardware(self):
...
+
+    def _process_wd(self, msg):
+        if self.platform.soft_power_held_time:
```
bosh: fast: platform config setting: soft_power_held_time

could be moved to fast_net: subconfig though tbh

mpf/platforms/fast/communicators/net_neuron.py
```diff
@@ -268,6 +271,19 @@ async def update_switches_from_hardware(self):
...
+            delta_seconds = time.time() - self.platform.soft_power_held_time
```
bosh: Not a dynamic lookup, but could be pushed down to here or wrapped in a lookup function

mpf/platforms/fast/communicators/net_neuron.py
```diff
@@ -268,6 +271,19 @@ async def update_switches_from_hardware(self):
...
+            self.platform.soft_power_held_time = None
+            self.warning_log('Soft power down released after %s seconds.', delta_seconds)
```
bosh: match the other log above if it changes

mpf/platforms/fast/communicators/net_neuron.py
```diff
@@ -338,3 +354,5 @@ def _process_switch_closed(self, msg):
     def stopping(self):
         """Stop the Neuron processor and disable the watchdog."""
         self.send_and_forget('WD:1')
+        if self.machine.soft_power_down_active:
+            self.send_and_forget('WP:1000')
```
bosh: could be a config based number instead. Note this is 0x1000, or 4096 ms=4.1s

mpf/config_spec.yaml
```diff
@@ -561,7 +562,7 @@ fast_net:
     console_log: single|enum(none,basic,full)|none
     file_log: single|enum(none,basic,full)|basic
     controller: single|enum(neuron,nano,sys11,wpc89,wpc95)|
-    watchdog: single|ms|1000
+    watchdog: single|ms|500
```
bosh, line 576: watchdog cut in half to improve resolution on soft power checkins. Otherwise, a release at 2.1s of hold will round up to 2.5 (watchdog is sent 2x as often as the watchdog ms here. So really, it's setting the watchdog coverage duration here from 1s to .5s, which changes the frequency from .5s to .25s.)

avanwinkle: Would it make sense to adjust the frequency when the first `WP` is detected? This would still have a rounding error of up to the initial watchdog time (did they press it 0.1 seconds ago or 0.9 seconds ago?) but could have a high-resolution detection during the press and then revert after the press is released.

mpf/platforms/fast/communicators/net_neuron.py
avanwinkle: Nit: Might be cleaner to do an early exit `if not self.platform.soft_power_held_time: return` so the rest of the code isn't nested in an `if `block`

mpf/platforms/fast/communicators/net_neuron.py
avanwinkle: The MPF loop keeps a clock that's generally used for time tracking via `self.machine.clock.get_time()` which is... more accurate? Or at least internally accurate? That's what's used throughout the codebase.

mpf/platforms/fast/communicators/net_neuron.py
```diff
@@ -37,6 +38,8 @@ def __init__(self, platform, processor, config):
         self.switches = list()
         self.drivers = list()
 
+        self.message_processors['WD:'] = self._process_wd
+        self.message_processors['WP:'] = self._process_wp
```
avanwinkle, line 41: I'm torn between having separate WD and WP commands versus having a different response for the WD...

mpf/core/machine.py
avanwinkle: I'm wondering if it might be valuable to have another event as soon as the press is detected? So the UI could communicate that the hold is being tracked, and also an event when the time has passed? That way the user knows when they've held it long enough, because the moment they release it the shutdown will begin and there won't be any time to give feedback before MPF shuts down.

avanwinkle (commented):
Very cool! This is a long-awaited feature!

mpf/core/machine.py
bosh: Yeah that's probably useful

mpf/platforms/fast/communicators/net_neuron.py
bosh: They have to be different, WP: triggers the beginning of a soft power down automatically, and should basically never be used except during the soft power down process.

bosh: And these are the processors, which actually don't have a choice -- the WP and WD prefix is determined by the soft power state, and both are responses to WD:#### commands.

mpf/platforms/fast/communicators/net_neuron.py
bosh: Probably, but there is no rest of the code, and I like the signalling of this style better than the early return that would be the last line of the function in either case. It's really up to a future patch to continue with the appropriate style

mpf/platforms/fast/communicators/net_neuron.py
bosh: The issue is not the precision of the timekeeper, it's the precision of the watchdog command frequency. It is not possible to get a tighter signal than the watchdog rate, so this PR also reduces that to .25s.

mpf/config_spec.yaml
bosh: Yes probably but TBH I'd rather do nothing and leave the default at 1000 than add more code to manage a variable watchdog checkin frequency. I'll just change my personal game configs to 500.

mpf/platforms/fast/communicators/net_neuron.py
avanwinkle: Fair enough. I'm typically in favor of early returns because I find it easier to read, but that's purely stylistic. :)

mpf/platforms/fast/communicators/net_neuron.py
avanwinkle: Sure sure, I didn't mean to suggest there's anything _in_accurate about `time`, just that most of the MPF codebase uses the clock timer to maintain accuracy across threads (or something?). Just the convention I was taught, only MPF tests and benchmarks use the system time, all machine code uses the clock time.

mpf/platforms/fast/communicators/net_neuron.py
avanwinkle: Yeah, I know it's not our choice, more of a consideration for maybe asking FAST for consistent WD for watchdog-related commands but your handling here is really elegant so that actually makes me more a fan of how it's being done :)

mpf/platforms/fast/communicators/net_neuron.py
bosh: Ah, the reason is for testability -- much easier to manage an injected or constructed clock. Fixed

mpf/platforms/fast/communicators/net_neuron.py
bosh: :) thanks!

mpf/core/machine.py
bosh: I made the press and release events FAST-specific because it's really about the FAST hardware built-in switch handling, which isn't actually visible to MPF as a normal switch. I'll also add an event to go before the boolean event, so that generic attempts can be hooked onto without risking hooking onto the boolean event itself.

mpf/core/machine.py
bosh: attempt_soft_shutdown

mpf/core/machine.py
bosh: soft_shutdown_exit_command

mpf/core/machine.py
bosh: added start_new_session=True

mpf/platforms/fast/communicators/net_neuron.py
bosh: moved to net, added additional param:
    soft_power_powerdown_delay: single|int|4000

avanwinkle (approved):

bosh:
Thanks for the approve :) I need to retest everything now that I've refactored (in fact I just noticed and fixed a bug with using **_kwargs)

mpf/config_spec.yaml
```diff
@@ -561,14 +561,16 @@ fast_net:
...
     default_quick_debounce_open: single|ms|2
     default_quick_debounce_close: single|ms|2
     default_normal_debounce_open: single|ms|4
     default_normal_debounce_close: single|ms|4
     mute_unconfigured_switches: single|bool|false
     gi_hz: single|int|30
     lamp_hz: single|int|30
+    soft_power_hold_time: single|float|2.5
+    soft_power_powerdown_delay: single|int|4000
```
bosh: TBH I'm not super happy about using two different types of numeric value. And probably I should default to adding things with template support so that settings can be used.

PR #2035, opened 2026-06-10
Auditor and file saving fixes

bosh:
developers were using game_ended etc and getting double counted audits because they did not realize the default auditor config already includes it

also adds a bunch of file management improvements and safety for all the builtin yaml files in data/
- yaml interface is threadsafe
- yaml interface uses fsync
- file manager uses directory fsync (and anticipates windows exceptions with permissionerror)
- data manager accepts config option on whether to use fsync (default to None which handles OS preferences, windows to False, else True)
- data manager uses fresh data dump if dirty on thread stop

### DIFF
--- a/mpf/config_spec.yaml
+++ b/mpf/config_spec.yaml
@@ -1026,6 +1026,10 @@ mpf:
     paths: ignore
     report_crashes: single|enum(ask,never,always)|never
     rgbw_white_behavior: single|enum(white_only,min_rgb,duck_rgb)|duck_rgb
+data_manager:
+    __valid_in__: machine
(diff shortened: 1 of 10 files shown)


bosh:
Oh I didnt note it in the commit message but the test setup caused an expected red on https://github.com/missionpinball/mpf/blob/757ac9f1a8e2d4b039b7739aa8680a64806fbf85/mpf/tests/test_Auditor.py#L139 getting 2 instead of 1. I didn't prove player vars are _necessary_ to have as a set, but there's no use case where list makes more sense anyway

PR #2041, opened 2026-06-12
utility function string_to_ms should not crash with more obvious cases

bosh:
It sketches me out how many of these were crashes before. ~~I almost worry that somehow something depended on things like "" crashing instead of converting to zero.~~ And who knows what this would do with negative numbers! (edit: left undefined for now! No opinion!)

On the other hand, this is a low level utility function, and I think it should be as permissive as possible in converting, so as to avoid crashes at runtime (since this is used all over the place). I'm not 100% sure I like converting the native True False and Nones, ~~but really the thing I'm unsettled on is whether "" should crash still (so we dont blindly start letting errors where a string should have come through and instead was truncated) or whether it should become 0. Maybe some special cases deserve logging.~~ Don't do it! Crash on blanks!

### DIFF
--- a/mpf/core/utility_functions.py
+++ b/mpf/core/utility_functions.py
@@ -588,13 +588,16 @@ class Util:
         if isinstance(time_string, (int, float)):
             return int(time_string)
 
-        time_string = str(time_string).upper()
+        time_string = str(time_string).upper().strip()
 
-        if time_string.endswith('MS') or time_string.endswith('MSEC'):
-            return int(time_string[:-2])
+        if time_string == 'NONE':
+            return 0
+
+        if time_string.endswith('MS'):
+            return int(float(time_string[:-2]))
 
         if time_string.endswith('MSEC'):
-            return int(time_string[:-4])
+            return int(float(time_string[:-4]))
 
         if time_string.endswith('D'):
             return int(float(time_string[:-1]) * 86400 * 1000)
@@ -611,7 +614,7 @@ class Util:
         if time_string.endswith('SEC'):
             return int(float(time_string[:-3]) * 1000)
 
-        return int(time_string)
+        return int(float(time_string))
 
     @staticmethod
     def string_to_secs(time_string: str) -> float:
--- a/mpf/tests/test_Utility_Functions.py
+++ b/mpf/tests/test_Utility_Functions.py
@@ -5,13 +5,60 @@ from mpf.core.utility_functions import Util
 class TestUtil(unittest.TestCase):
 
     def test_string_to_ms(self):
-        self.assertEqual(86400000, Util.string_to_ms('1d'))
-        self.assertEqual(3600000, Util.string_to_ms('1h'))
-        self.assertEqual(60000, Util.string_to_ms('1m'))
-        self.assertEqual(1000, Util.string_to_ms('1s'))
-        self.assertEqual(1, Util.string_to_ms('1ms'))
-        self.assertEqual(0, Util.string_to_ms(None))
-        self.assertEqual(0, Util.string_to_ms(False))
+        # Base Units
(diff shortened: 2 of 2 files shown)


mpf/core/utility_functions.py
```diff
@@ -588,13 +588,16 @@ def string_to_ms(time_string: str) -> int:
         if isinstance(time_string, (int, float)):
             return int(time_string)
 
-        time_string = str(time_string).upper()
+        time_string = str(time_string).upper().strip()
```
bosh, line 591: Adds whitespace handling. This could be cut as it might let weird things through that shouldnt be.

mpf/core/utility_functions.py
```diff
@@ -588,13 +588,16 @@ def string_to_ms(time_string: str) -> int:
...
 
-        if time_string.endswith('MS') or time_string.endswith('MSEC'):
-            return int(time_string[:-2])
+        if time_string in ('NONE', ''):
```
bosh: Really unsure about "" going to 0, but NONE seems reasonable.

mpf/core/utility_functions.py
```diff
@@ -588,13 +588,16 @@ def string_to_ms(time_string: str) -> int:
         if isinstance(time_string, (int, float)):
             return int(time_string)
 
-        time_string = str(time_string).upper()
+        time_string = str(time_string).upper().strip()
 
-        if time_string.endswith('MS') or time_string.endswith('MSEC'):
```
bosh, line 593: msec was cut at the end by 2 chars, becoming X ms, but that would leave ms in the int()'d string, which doesnt work

mpf/core/utility_functions.py
```diff
@@ -588,13 +588,16 @@ def string_to_ms(time_string: str) -> int:
...
-            return int(time_string[:-2])
+        if time_string in ('NONE', ''):
+            return 0
+
+        if time_string.endswith('MS'):
+            return int(float(time_string[:-2]))
 
         if time_string.endswith('MSEC'):
-            return int(time_string[:-4])
+            return int(float(time_string[:-4]))
```
bosh, line 600: All string=>ints should go through float conversion to allow for 100.0 etc to cleanly convert instead of crashing. They're close enough to be right to be worth it.

mpf/core/utility_functions.py
```diff
@@ -611,7 +614,7 @@ def string_to_ms(time_string: str) -> int:
         if time_string.endswith('SEC'):
             return int(float(time_string[:-3]) * 1000)
 
-        return int(time_string)
+        return int(float(time_string))
```
bosh, line 617: Fallback wasnt handling "100.0" as 100, instead preferring to crash

mpf/tests/test_Utility_Functions.py
```diff
@@ -5,13 +5,55 @@
...
+        # Whitespace handling
+        self.assertEqual(Util.string_to_ms('  500ms  '), 500)
+        self.assertEqual(Util.string_to_ms('2s '), 2000)
+
+    def test_string_to_ms_unhandled(self):
+        with self.assertRaises(ValueError):
+            Util.string_to_ms('pinball')
+
+        with self.assertRaises(ValueError):
+            Util.string_to_ms('True')
```
bosh, line 55: This could be argued like `"None"` to convert to a 1 or 0 instead of crash. Iono

mpf/core/utility_functions.py
bosh: Let's play it safe and try to avoid crashes

mpf/core/utility_functions.py
bosh: This is the only one I'm unsure of still -- blank seems like a true mistake or bug rather than an intentional 0. Could add a warning?

PR #2043, opened 2026-06-13
FAST: fix serial response timeout/retry that froze startup on a lost reply

### LINKED ISSUE #2036: FAST: a single lost serial response freezes MPF startup forever (timeout/retry in send_and_wait_for_response_processed never engages)
﻿
**Affected:** 0.80.0 stable and current dev (`757ac9f`), `mpf/platforms/fast/communicators/base.py`

## Symptom

Roughly every other boot, MPF freezes silently during FAST EXP initialization. The
log just stops â€” no error, no timeout, no traceback. Last lines look like:

```
INFO [FAST [EXP_INT]] Verifying hardware for EXP "neuron" (FP-EXP-2000, @48) with ID string "EXP FP-EXP-2000  0.48", board address 48
INFO [FAST [EXP_INT]] Verifying hardware for EXP "neuron" (FP-EXP-2000, @48) with ID string "BRK FP-EXP-2000  0.48", board address 480
```

(i.e. it hangs awaiting the response to the *next* command â€” in our case the
`BR@48:` board reset or the next board's `ID@` query.)

The "every other run" pattern comes from unclean shutdowns: if the previous MPF
process was killed while streaming LED data, the EXP board can be left
mid-message, and one response to the next session's init commands gets lost.
Hardware: Neuron (FP-CPU-2000), built-in FP-EXP-2000 plus FP-EXP-0071, Windows 11.

## Root cause

`send_and_wait_for_response_processed()` puts the `asyncio.wait_for()` timeout
around the wrong await:

```python
while max_retries == -1 or retries <= max_retries:
    try:
        await asyncio.wait_for(self.send_and_wait_for_response(msg, pause_sending_until,
                                                               log_msg), timeout=timeout)
        break
    except asyncio.TimeoutError:
        self.log.error("Timeout waiting for response to %s. Retrying...", msg)
        retries += 1

await self.done_waiting.wait()
```

Two problems:

1. `send_and_wait_for_response()` does not wait for the response â€” it awaits
   `no_response_waiting`, queues the message via `send_with_confirmation()`
   (a non-blocking `put_nowait`), and returns. So the `wait_for(timeout=1)`
   virtually always completes instantly and **the timeout guards nothing**.
2. The *actual* wait for the response is the final
   `await self.done_waiting.wait()` â€” which has **no timeout**. If the response
   never arrives, this awaits forever, silently. The retry loop above is
   unreachable in practice, and even if it fired, the resent message would
   deadlock: `no_response_waiting` is still cleared and the writer's
   `pause_sending` state is still pending from the lost response.

All callers that rely on retries (e.g. `ID:` polls with `max_retries=-1`) are
equally affected: they only ever "retry" if the *send queue* stalls, not if the
board fails to answer.

## Fix

Move the timeout onto the `done_waiting` wait, release the gating events before
resending, and fail loudly when retries are exhausted instead of hanging:

```python
self.done_waiting.clear()

retries = 0

while max_retries == -1 or retries <= max_retries:
    await self.send_and_wait_for_response(msg, pause_sending_until, log_msg)
    try:
        await asyncio.wait_for(self.done_waiting.wait(), timeout=timeout)
        return
    except asyncio.TimeoutError:
        self.log.error("Timeout waiting for response to %s. Retrying...", msg)
        self._resume_sending()
        self.no_response_waiting.set()
        retries += 1

raise AssertionError(f"FAST board on {self.config['port']} never responded to '{msg}' "
                     f"(tried {retries} times)")
```

We'd also suggest bumping the default `max_retries` from 0 to a small number
(we use 3) â€” every message sent through this path (`ID@`, `BR@`, `NN:`, `SA:`,
driver configs) is idempotent, and a default of 0 means a single lost response
is fatal even with the timeout fixed.

Related: #1900 fixed the same kind of dead retry path in the port detector.

We've been running this fix locally (0.80.0, FAST Neuron) â€” the every-other-run
freeze is gone across repeated boot cycles, including immediately after hard
kills with the LED stream active. Happy to submit it as a PR (optionally with a
mock-serial test that drops one ID response and asserts init still completes).

sjmiles:
Fixes #2036.

## Problem

Roughly every other boot, MPF freezes silently during FAST EXP init — no error, no timeout, no traceback. It happens when a single serial response is lost (e.g. the previous process was killed while streaming LED data, leaving the EXP board mid-message).

`send_and_wait_for_response_processed()` puts the `asyncio.wait_for()` timeout around the wrong await:

* `send_and_wait_for_response()` does **not** wait for the response — it awaits the send-queue gate, queues the message via a non-blocking `put_nowait`, and returns. So `wait_for(..., timeout=1)` virtually always completes instantly and **the timeout guards nothing**.
* The *actual* wait is the final `await self.done_waiting.wait()`, which has **no timeout**. If the response never arrives, it awaits forever, silently. The retry loop is effectively unreachable, and even if it fired the resend would deadlock (the gating events are still pending from the lost response).

## Fix

* Move the timeout onto `done_waiting.wait()` — the wait that actually corresponds to the board's response.
* On timeout, release the gating state (`_resume_sending()` + `no_response_waiting.set()`) before resending, so the retry doesn't deadlock in `send_and_wait_for_response()`.
* Raise `AssertionError` (naming the port and message) when retries are exhausted, instead of hanging.
* Bump default `max_retries` from `0` to `3`. Every message on this path (`ID@`, `BR@`, `NN:`, `SA:`, driver configs) is idempotent, so a single lost response shouldn't be fatal even with the timeout fixed.

Related: #1900 fixed the same kind of dead retry path in the port detector.

## Tests

`mpf/tests/test_Fast_Communicator.py` — drives `send_and_wait_for_response_processed()` with a stubbed I/O surface:
* `test_returns_when_response_processed` (happy path, exactly one send)
* `test_raises_instead_of_hanging_when_no_response` (retries then raises; asserts 4 sends fired — the retry path is now live — and never hangs)
* `test_unlimited_retries_eventually_succeed` (`max_retries=-1`, used by `ID:` polls, succeeds once the board finally answers)

All new tests plus the full existing FAST suite pass.

Verified locally (0.80.0, FAST Neuron): the every-other-run freeze is gone across repeated boot cycles, including immediately after hard kills with the LED stream active.

## Follow-up: clear leftover serial bytes so the reply isn't lost in the first place

The timeout/retry above converts a silent hang into a loud failure, but it surfaced a second, related cause from the same unclean-shutdown scenario — and here the board **does** answer; the reply is just lost on our side.

After a hard kill mid LED-blast, the board or OS buffer can still hold a burst of leftover bytes with **no trailing `<CR>`**. `reset_input_buffer()` only flushes what was buffered at that instant; bytes still draining over USB arrive just afterwards and fuse onto the front of the next real message — `<junk>ID:exp ...`. Because messages are framed on `<CR>`, the junk and the genuine reply become one segment, which then either:

* fails to UTF-8 decode → raises and kills `read_task` (with `ignore_decode_errors` False during init), or
* dispatches with a junk 3-char header → the real reply is silently dropped.

Either way the handshake reply is lost and we fall into the timeout/retry path even though the board responded.

Two fixes (commit `5aafaff8d`):

* `_drain_serial()` — before starting the read loop, read and discard inbound bytes until the stream is quiet for 100 ms (capped at 2 s total). The leftover burst is finite, so this clears it and the handshake reply arrives clean. Root-cause fix.
* `_resync_segment()` — when a segment fails to decode, scan for the earliest registered message header and dispatch from there instead of dropping the segment or raising. Recovers a real message fused onto the tail of junk; defense-in-depth for junk that fuses in the same read after init.

Covered by the existing FAST integration suites (Exp / Neuron / Nano / Retro etc.), which exercise `connect()` and the parse path during board init. Full FAST suite passes.

### DIFF
--- a/mpf/platforms/fast/communicators/base.py
+++ b/mpf/platforms/fast/communicators/base.py
@@ -146,6 +146,13 @@ class FastSerialCommunicator(LogMixin):
 
         await self.clear_board_serial_buffer()
 
+        # reset_input_buffer() above only flushes what was buffered at that
+        # instant. After an unclean shutdown the board can still be clocking out
+        # leftover bytes (often un-terminated binary LED data) that arrive just
+        # afterwards. Drain until the stream goes quiet so the first handshake
+        # response isn't fused onto the tail of that garbage.
+        await self._drain_serial()
+
         self.ignore_decode_errors = False
 
         self.write_task = asyncio.create_task(self._socket_writer())
@@ -157,6 +164,33 @@ class FastSerialCommunicator(LogMixin):
     async def clear_board_serial_buffer(self):
         """Clear out the serial buffer."""
         self.write_to_port(b'\r\r\r\r')
+
+    async def _drain_serial(self, quiet_period=0.1, max_drain_time=2.0):
+        """Read and discard inbound bytes until the serial stream goes quiet.
+
+        After an unclean shutdown (e.g. Ctrl-C mid LED-blast) the board or OS
+        buffer can still hold a burst of leftover bytes with no trailing <CR>.
+        If that reaches the parser it fuses onto the front of the next real
+        message (``<junk>ID:exp ...``), so the response is dropped or raises a
+        decode error. This runs before the read loop starts, so it can read the
+        stream directly: keep reading until nothing has arrived for
+        ``quiet_period`` seconds, bounded by ``max_drain_time`` total.
+        """
+        deadline = self.machine.clock.loop.time() + max_drain_time
+        while True:
+            try:
+                chunk = await asyncio.wait_for(self.reader.read(1024), quiet_period)
+            except asyncio.TimeoutError:
+                return  # nothing for quiet_period -> stream is drained
+            if not chunk:
+                return  # port closed; the normal read path will report it
+            if self.port_debug:
+                self.log.info("Drained %s leftover serial byte(s): %s", len(chunk), chunk)
+            if self.machine.clock.loop.time() >= deadline:
+                self.log.warning(
+                    "Serial port still producing data after %ss; proceeding with handshake anyway.",
+                    max_drain_time)
+                return
 
     async def init(self):
         """Initialize the communicator with any board-specific logic."""
@@ -407,14 +441,46 @@ class FastSerialCommunicator(LogMixin):
                 if self.machine.is_shutting_down:
                     return
 
-                self.log.warning("Interference / bad data received: %s", msg)
-                if not self.ignore_decode_errors:
-                    raise
+                # Un-terminated leftover junk fuses onto the front of the next
+                # real message. Try to resync to a known header before treating
+                # this as bad data, so we don't drop (or crash on) a response
+                # sitting at the tail of the garbage.
+                recovered = self._resync_segment(msg)
+                if recovered is not None:
+                    self.log.warning("Recovered '%s' after discarding leading junk bytes", recovered)
+                    msg = recovered
+                else:
+                    self.log.warning("Interference / bad data received: %s", msg)
+                    if not self.ignore_decode_errors:
+                        raise
+                    continue
 
             if self.port_debug:
                 self.log.info("<<<< %s", msg)
 
             self._dispatch_incoming_msg(msg)
+
+    def _resync_segment(self, raw):
+        """Recover a valid message from a segment polluted with leading junk.
+
+        Un-terminated leftover binary has no <CR> of its own, so it fuses onto
+        the front of the following message. Rather than dropping the whole
+        segment (or raising on the un-decodable junk), find the earliest
+        registered message header in the segment and return the message from
+        there. Returns the decoded string, or None if no known header is found
+        or the recovered remainder still won't decode.
+        """
+        earliest = None
+        for header in self.message_processors:
+            idx = raw.find(header.encode())
+            if idx != -1 and (earliest is None or idx < earliest):
+                earliest = idx
+        if earliest is None:
+            return None
+        try:
+            return raw[earliest:].decode()
+        except UnicodeDecodeError:
+            return None
 
     def _dispatch_incoming_msg(self, msg):
         # Figures out what to do with incoming messages
--- a/mpf/platforms/fast/communicators/base.py
+++ b/mpf/platforms/fast/communicators/base.py
@@ -11,6 +11,16 @@ from mpf.core.utility_functions import Util
 
 MIN_FW = version.parse('0.00')  # override in subclass
 
+# After a response wait_for() times out, drain at most this many event-loop
+# passes (re-checking done_waiting) before treating it as a real timeout. Under
+# cooperative scheduling the timeout timer and the inbound response bytes can
+# become ready in the same pass, so wait_for() raises while the reply is still a
+# few parse callbacks from being handled; draining lets it land instead of
+# triggering a spurious resend. Sized well above the observed worst case (a
+# board's full init parse chain); only a genuinely lost response runs the full
+# count.
+RESPONSE_DRAIN_PASSES = 50
+
 
 # pylint: disable-msg=too-many-instance-attributes
 class FastSerialCommunicator(LogMixin):
@@ -269,7 +279,7 @@ class FastSerialCommunicator(LogMixin):
 
     # pylint: disable-msg=too-many-arguments
     async def send_and_wait_for_response_processed(self, msg, pause_sending_until, timeout=1,
-                                                   max_retries=0, log_msg=None):
+                                                   max_retries=3, log_msg=None):
         """Send a message and wait for the response to be processed.
 
         Unlike send_and_wait_for_response(), this method will not release the wait when the response is received.
@@ -285,7 +295,9 @@ class FastSerialCommunicator(LogMixin):
                 If a response is not received by then (based on the pause_sending_until), the message will be resent.
                 Defaults to 1.
             max_retries (int, optional): How many times the message will be resent if the response is not
-                received by the timeout. -1 means unlimited retries. Defaults to 0.
+                received by the timeout. -1 means unlimited retries. Defaults to 3. Commands on this path
+                should be idempotent, since a retry resends them; when retries are exhausted an
+                AssertionError is raised rather than hanging.
             log_msg (_type_, optional): Optional version of the message that will be used in logs.
                 Typically used with binary messages so the longs can contain human readable versions.
                 Defaults to None which means the actual msg will be used in the logs.
@@ -295,15 +307,45 @@ class FastSerialCommunicator(LogMixin):
         retries = 0
 
         while max_retries == -1 or retries <= max_retries:
-            try:
-                await asyncio.wait_for(self.send_and_wait_for_response(msg, pause_sending_until,
-                                                                       log_msg), timeout=timeout)
-                break
-            except asyncio.TimeoutError:
-                self.log.error("Timeout waiting for response to %s. Retrying...", msg)
-                retries += 1
+            await self.send_and_wait_for_response(msg, pause_sending_until, log_msg)
+            # done_waiting is set by done_processing_msg_response() once the
+            # reply has been handled; wait for it with a timeout so a lost
+            # response can't hang here forever.
+            if await self._wait_for_processed(timeout):
+                return
 
-        await self.done_waiting.wait()
+            self.log.error("Timeout waiting for response to %s. Retrying...", msg)
+            # Timing out means the writer is stuck paused but we don't expect any
+            # more response. To avoid deadlocking, release no_response_waiting and
+            # clear pause_sending_flag/until.
+            self._resume_sending()
+            self.no_response_waiting.set()
+            retries += 1
+
+        raise AssertionError(
+            f"FAST board on {self.config['port']} never responded to '{msg}' after {retries} attempt(s)")
+
+    async def _wait_for_processed(self, timeout):
+        """Wait up to ``timeout`` seconds for the response to be processed.
+
+        Returns True if done_processing_msg_response() fired, False on timeout.
+
+        A response delivered right at the deadline can still be a couple of
+        callbacks short of being processed when wait_for() times out (under
+        cooperative scheduling the timeout and the inbound bytes can become
+        ready in the same loop pass). Let those pending callbacks drain and
+        re-check before declaring a real timeout, so we don't resend — and
+        double-process — a response that actually arrived.
+        """
+        try:
+            await asyncio.wait_for(self.done_waiting.wait(), timeout)
+            return True
+        except asyncio.TimeoutError:
+            for _ in range(RESPONSE_DRAIN_PASSES):
+                await asyncio.sleep(0)
+                if self.done_waiting.is_set():
+                    return True
+            return False
 
     def done_processing_msg_response(self):
         """Releases the wait for the response to be processed.
--- a/mpf/platforms/fast/communicators/base.py
+++ b/mpf/platforms/fast/communicators/base.py
@@ -11,15 +11,15 @@ from mpf.core.utility_functions import Util
 
 MIN_FW = version.parse('0.00')  # override in subclass
 
-# After a response wait_for() times out, drain at most this many event-loop
+# After a response wait_for() times out, yield the event loop at most this many
 # passes (re-checking done_waiting) before treating it as a real timeout. Under
 # cooperative scheduling the timeout timer and the inbound response bytes can
 # become ready in the same pass, so wait_for() raises while the reply is still a
-# few parse callbacks from being handled; draining lets it land instead of
-# triggering a spurious resend. Sized well above the observed worst case (a
-# board's full init parse chain); only a genuinely lost response runs the full
-# count.
-RESPONSE_DRAIN_PASSES = 50
+# few parse callbacks from being handled; letting the loop settle lets it land
+# instead of triggering a spurious resend. The observed worst case (a board's
+# full init parse chain) was ~6 passes; 20 leaves margin. Each pass is just an
+# asyncio.sleep(0) yield, so only a genuinely lost response runs the full count.
+RESPONSE_SETTLE_PASSES = 20
 
 
 # pylint: disable-msg=too-many-instance-attributes
@@ -367,15 +367,15 @@ class FastSerialCommunicator(LogMixin):
         A response delivered right at the deadline can still be a couple of
         callbacks short of being processed when wait_for() times out (under
         cooperative scheduling the timeout and the inbound bytes can become
-        ready in the same loop pass). Let those pending callbacks drain and
-        re-check before declaring a real timeout, so we don't resend — and
-        double-process — a response that actually arrived.
+        ready in the same loop pass). Let the loop settle and re-check before
+        declaring a real timeout, so we don't resend — and double-process — a
+        response that actually arrived.
         """
         try:
             await asyncio.wait_for(self.done_waiting.wait(), timeout)
             return True
         except asyncio.TimeoutError:
-            for _ in range(RESPONSE_DRAIN_PASSES):
+            for _ in range(RESPONSE_SETTLE_PASSES):
                 await asyncio.sleep(0)
                 if self.done_waiting.is_set():
                     return True
@@ -441,46 +441,14 @@ class FastSerialCommunicator(LogMixin):
                 if self.machine.is_shutting_down:
                     return
 
-                # Un-terminated leftover junk fuses onto the front of the next
-                # real message. Try to resync to a known header before treating
-                # this as bad data, so we don't drop (or crash on) a response
-                # sitting at the tail of the garbage.
-                recovered = self._resync_segment(msg)
-                if recovered is not None:
-                    self.log.warning("Recovered '%s' after discarding leading junk bytes", recovered)
-                    msg = recovered
-                else:
-                    self.log.warning("Interference / bad data received: %s", msg)
-                    if not self.ignore_decode_errors:
-                        raise
-                    continue
+                self.log.warning("Interference / bad data received: %s", msg)
+                if not self.ignore_decode_errors:
+                    raise
 
             if self.port_debug:
                 self.log.info("<<<< %s", msg)
 
             self._dispatch_incoming_msg(msg)
-
-    def _resync_segment(self, raw):
-        """Recover a valid message from a segment polluted with leading junk.
-
-        Un-terminated leftover binary has no <CR> of its own, so it fuses onto
-        the front of the following message. Rather than dropping the whole
-        segment (or raising on the un-decodable junk), find the earliest
-        registered message header in the segment and return the message from
-        there. Returns the decoded string, or None if no known header is found
-        or the recovered remainder still won't decode.
-        """
-        earliest = None
-        for header in self.message_processors:
-            idx = raw.find(header.encode())
-            if idx != -1 and (earliest is None or idx < earliest):
-                earliest = idx
-        if earliest is None:
-            return None
-        try:
-            return raw[earliest:].decode()
-        except UnicodeDecodeError:
-            return None
 
     def _dispatch_incoming_msg(self, msg):
         # Figures out what to do with incoming messages
--- a/mpf/platforms/fast/communicators/base.py
+++ b/mpf/platforms/fast/communicators/base.py
@@ -11,15 +11,15 @@ from mpf.core.utility_functions import Util
 
 MIN_FW = version.parse('0.00')  # override in subclass
 
-# After a response wait_for() times out, drain at most this many event-loop
+# After a response wait_for() times out, yield the event loop at most this many
 # passes (re-checking done_waiting) before treating it as a real timeout. Under
 # cooperative scheduling the timeout timer and the inbound response bytes can
 # become ready in the same pass, so wait_for() raises while the reply is still a
-# few parse callbacks from being handled; draining lets it land instead of
-# triggering a spurious resend. Sized well above the observed worst case (a
-# board's full init parse chain); only a genuinely lost response runs the full
-# count.
-RESPONSE_DRAIN_PASSES = 50
+# few parse callbacks from being handled; letting the loop settle lets it land
+# instead of triggering a spurious resend. The observed worst case (a board's
+# full init parse chain) was ~6 passes; 20 leaves margin. Each pass is just an
+# asyncio.sleep(0) yield, so only a genuinely lost response runs the full count.
+RESPONSE_SETTLE_PASSES = 20
 
 
 # pylint: disable-msg=too-many-instance-attributes
@@ -365,15 +365,15 @@ class FastSerialCommunicator(LogMixin):
         A response delivered right at the deadline can still be a couple of
         callbacks short of being processed when wait_for() times out (under
         cooperative scheduling the timeout and the inbound bytes can become
-        ready in the same loop pass). Let those pending callbacks drain and
-        re-check before declaring a real timeout, so we don't resend — and
-        double-process — a response that actually arrived.
+        ready in the same loop pass). Let the loop settle and re-check before
+        declaring a real timeout, so we don't resend — and double-process — a
+        response that actually arrived.
         """
         try:
             await asyncio.wait_for(self.done_waiting.wait(), timeout)
             return True
         except asyncio.TimeoutError:
-            for _ in range(RESPONSE_DRAIN_PASSES):
+            for _ in range(RESPONSE_SETTLE_PASSES):
                 await asyncio.sleep(0)
                 if self.done_waiting.is_set():
                     return True
@@ -439,46 +439,14 @@ class FastSerialCommunicator(LogMixin):
                 if self.machine.is_shutting_down:
                     return
 
-                # Un-terminated leftover junk fuses onto the front of the next
-                # real message. Try to resync to a known header before treating
-                # this as bad data, so we don't drop (or crash on) a response
-                # sitting at the tail of the garbage.
-                recovered = self._resync_segment(msg)
-                if recovered is not None:
-                    self.log.warning("Recovered '%s' after discarding leading junk bytes", recovered)
-                    msg = recovered
-                else:
-                    self.log.warning("Interference / bad data received: %s", msg)
-                    if not self.ignore_decode_errors:
-                        raise
-                    continue
+                self.log.warning("Interference / bad data received: %s", msg)
+                if not self.ignore_decode_errors:
+                    raise
 
             if self.port_debug:
                 self.log.info("<<<< %s", msg)
 
             self._dispatch_incoming_msg(msg)
-
-    def _resync_segment(self, raw):
-        """Recover a valid message from a segment polluted with leading junk.
-
-        Un-terminated leftover binary has no <CR> of its own, so it fuses onto
-        the front of the following message. Rather than dropping the whole
-        segment (or raising on the un-decodable junk), find the earliest
-        registered message header in the segment and return the message from
-        there. Returns the decoded string, or None if no known header is found
-        or the recovered remainder still won't decode.
-        """
-        earliest = None
-        for header in self.message_processors:
-            idx = raw.find(header.encode())
-            if idx != -1 and (earliest is None or idx < earliest):
-                earliest = idx
-        if earliest is None:
-            return None
-        try:
-            return raw[earliest:].decode()
-        except UnicodeDecodeError:
-            return None
 
     def _dispatch_incoming_msg(self, msg):
         # Figures out what to do with incoming messages
--- a/mpf/platforms/fast/communicators/base.py
+++ b/mpf/platforms/fast/communicators/base.py
@@ -11,6 +11,16 @@ from mpf.core.utility_functions import Util
 
 MIN_FW = version.parse('0.00')  # override in subclass
 
+# After a response wait_for() times out, drain at most this many event-loop
+# passes (re-checking done_waiting) before treating it as a real timeout. Under
+# cooperative scheduling the timeout timer and the inbound response bytes can
+# become ready in the same pass, so wait_for() raises while the reply is still a
+# few parse callbacks from being handled; draining lets it land instead of
+# triggering a spurious resend. Sized well above the observed worst case (a
+# board's full init parse chain); only a genuinely lost response runs the full
+# count.
+RESPONSE_DRAIN_PASSES = 50
+
 
 # pylint: disable-msg=too-many-instance-attributes
 class FastSerialCommunicator(LogMixin):
@@ -267,7 +277,7 @@ class FastSerialCommunicator(LogMixin):
 
     # pylint: disable-msg=too-many-arguments
     async def send_and_wait_for_response_processed(self, msg, pause_sending_until, timeout=1,
-                                                   max_retries=0, log_msg=None):
+                                                   max_retries=3, log_msg=None):
         """Send a message and wait for the response to be processed.
 
         Unlike send_and_wait_for_response(), this method will not release the wait when the response is received.
@@ -283,7 +293,9 @@ class FastSerialCommunicator(LogMixin):
                 If a response is not received by then (based on the pause_sending_until), the message will be resent.
                 Defaults to 1.
             max_retries (int, optional): How many times the message will be resent if the response is not
-                received by the timeout. -1 means unlimited retries. Defaults to 0.
+                received by the timeout. -1 means unlimited retries. Defaults to 3. Commands on this path
+                should be idempotent, since a retry resends them; when retries are exhausted an
+                AssertionError is raised rather than hanging.
             log_msg (_type_, optional): Optional version of the message that will be used in logs.
                 Typically used with binary messages so the longs can contain human readable versions.
                 Defaults to None which means the actual msg will be used in the logs.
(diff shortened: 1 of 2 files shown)


mpf/platforms/fast/communicators/base.py
```diff
@@ -267,7 +267,7 @@ async def send_and_wait_for_response(self, msg, pause_sending_until, log_msg=Non
 
     # pylint: disable-msg=too-many-arguments
     async def send_and_wait_for_response_processed(self, msg, pause_sending_until, timeout=1,
-                                                   max_retries=0, log_msg=None):
```
bosh, line 272: You're probably right that this by default should just retry things, but I'll have to check all the callsites thoroughly to be sure

mpf/platforms/fast/communicators/base.py
```diff
@@ -283,7 +283,9 @@ async def send_and_wait_for_response_processed(self, msg, pause_sending_until, t
                 If a response is not received by then (based on the pause_sending_until), the message will be resent.
                 Defaults to 1.
             max_retries (int, optional): How many times the message will be resent if the response is not
-                received by the timeout. -1 means unlimited retries. Defaults to 0.
+                received by the timeout. -1 means unlimited retries. Defaults to 3. Every message sent
+                through this path (ID@, BR@, NN:, SA:, driver configs) is idempotent, so retrying a lost
```
bosh: I worry that the specific listing of commands will become stale, so I'd probably cut them and say that most commands, or that commands _should_ be idempotent if using the retry

mpf/platforms/fast/communicators/base.py
```diff
@@ -293,15 +295,50 @@ async def send_and_wait_for_response_processed(self, msg, pause_sending_until, t
...
-                await asyncio.wait_for(self.send_and_wait_for_response(msg, pause_sending_until,
-                                                                       log_msg), timeout=timeout)
-                break
-            except asyncio.TimeoutError:
-                self.log.error("Timeout waiting for response to %s. Retrying...", msg)
-                retries += 1
+            await self.send_and_wait_for_response(msg, pause_sending_until, log_msg)
+            # The actual response is signalled by done_waiting, which
+            # done_processing_msg_response() sets once the reply has been
+            # handled. The previous code wrapped the timeout around
```
bosh: There's no reason to mention previous code in a final diff as it's only relevant to the issue/moment in time this was fixed, not for forward maintenance

mpf/platforms/fast/communicators/base.py
```diff
@@ -293,15 +295,50 @@ async def send_and_wait_for_response_processed(self, msg, pause_sending_until, t
...
+            # a non-blocking put_nowait), so the timeout guarded nothing and the
+            # real, unguarded wait below hung forever on a lost response. #2036
+            if await self._wait_for_processed(timeout):
+                return
+
+            self.log.error("Timeout waiting for response to %s. Retrying...", msg)
+            # The lost response left the gating state pending: the writer is
+            # still paused waiting for a confirmation header that never came,
+            # and no_response_waiting is still cleared. Release both or the
+            # resend deadlocks in send_and_wait_for_response().
```
bosh: Tighten up copy, something like:

```
Timing out means the writer is stuck paused but we dont expect any more response.
To avoid deadlocking, release no_response_waiting and clear pause_sending_flag/until
```

mpf/platforms/fast/communicators/base.py
```diff
@@ -293,15 +295,50 @@ async def send_and_wait_for_response_processed(self, msg, pause_sending_until, t
...
+        ``asyncio.wait_for(self.done_waiting.wait(), timeout)``. wait_for() arms
+        a timer, and MPF's cooperative unit-test loop (mpf/tests/loop.py)
+        advances time to the nearest timer *before* servicing ready serial FDs,
+        so that timer fires before an in-flight response is delivered and
+        triggers a spurious retry. Polling re-checks the event after each step,
+        so a response that is mid-delivery is still observed, while a genuinely
+        lost response still times out.
+        """
+        loop = asyncio.get_event_loop()
+        deadline = loop.time() + timeout
```
bosh: I think you're likely correct here, but I need to dig into the mpf machine clock management -- there may be a more standard way of reading the time in a way that tests can hook into (I mention because of the comment's note of mpf/tests/loop.py)

mpf/tests/test_Fast_Communicator.py
```diff
@@ -0,0 +1,120 @@
...
+                comm.send_and_wait_for_response_processed(
+                    'ID@48:', 'ID:', timeout=0.01, max_retries=-1),
+                timeout=5)
+            await responder_task
+
+        # Must complete without raising.
+        asyncio.run(scenario())
+
+
+if __name__ == '__main__':
```
bosh: Cut please -- maybe these are super useful and we should have them everywhere, but I want to be more intentional about that decision rather than take them in bugfix PRs

bosh (changes requested):
This is really good! But it's a tricky enough thing that I want to pull it down and test it myself on live hardware before approving, and I also want to check in on the callsites of the send_..._response_processed function to be sure we want retries everywhere

sjmiles:
Heads up — I force-pushed a revision to fix the CI failure on this PR (it was mine, not flaky).

**What broke:** my first cut used `asyncio.wait_for(self.done_waiting.wait(), timeout=...)`. That arms a timer, and the cooperative unit-test loop (`mpf/tests/loop.py`) advances time to the nearest timer *before* servicing ready serial FDs. During EXP init the 1s timer fired before the in-flight `ID:` response was delivered, triggering a spurious retry that re-sent `ID@` after `_process_id` had already reset `active_board` → `TypeError: 'NoneType' object is not subscriptable` across the FAST suites. It only showed up on ubuntu/macos because callback ordering on windows happened to deliver the response first.

**Fix:** wait for `done_waiting` by polling it against the loop clock instead of arming a `wait_for` timer. Polling re-checks the event after each step, so a mid-delivery response is still observed, while a genuinely lost response still times out. Same behavior on real hardware; no preempting timer in the harness. Updated commit message and the `_wait_for_processed()` docstring explain it inline.

Locally the previously-failing suites pass (test_Fast_Exp/Neuron/Nano/Retro) along with the new test_Fast_Communicator unit tests. The new run is sitting in `action_required` — whenever one of you can approve the workflow, it should go green. Thanks!

sjmiles:
Thanks for the thorough review, bosh — really appreciate the careful read. Revised and force-pushed; here's how I handled each point:

- **`max_retries` default:** kept at 3 per your lean, but cut the specific command list from the docstring (agreed it'd drift out of date) — it now just notes that commands on this path should be idempotent since a retry resends them. Easy to drop back to 0 if your callsite audit turns up anything that shouldn't be retried by default.
- **"previous code" comment:** removed — agreed, that's point-in-time context that belongs in the issue, not in the source for forward maintenance.
- **deadlock-release comment:** tightened to your suggested wording.
- **time source:** switched from `asyncio.get_event_loop().time()` to `self.machine.clock.get_time()` — that's the standard, test-hookable clock you were pointing at, and it let me drop the `tests/loop.py` reference from the docstring entirely.
- **`test_Fast_Communicator.py`:** cut. The existing FAST integration suites (Exp/Neuron/Nano/Retro) already exercise this path during board init and caught the original regression, so coverage stays without introducing the new stubbing pattern in a bugfix PR.

No rush on the live-hardware test — ping me if anything turns up and I'll jump on it. Thanks again!

bosh:
Gotcha, no worries about force pushes, that's how I keep all my branches up to date with rebases before merging in as a single change from the latest HEAD. I just pushed a fix to dev that handles the coverall breakage as well, so if you just missed it, you'll want to rebase on that one too

sjmiles:
For the record Claude (AI) is helping me write these PRs, although I am keeping watch on him. Some of the wordiness is his, ahem. I'm going to check the tests against my Mac here too, fwiw.

sjmiles:
Thanks again for the detailed review — pushed a revision that addresses your comments, and along the way found (and fixed) a macOS-only regression worth explaining.

**Your review comments:**
- **`max_retries` default** — kept at 3, but cut the specific command list from the docstring (agreed it'd go stale); it now just says commands on this path should be idempotent. Happy to drop to 0 if your callsite audit prefers it.
- **"previous code" comment** — removed; agreed it's point-in-time context, not forward maintenance.
- **deadlock-release comment** — tightened to your wording.
- **`test_Fast_Communicator.py`** — cut. The existing FAST integration suites already exercise this path during init.
- **time source / `tests/loop.py` reference** — this one drove a rethink (below); the final version doesn't read the clock manually at all, so the reference is gone.

**The macOS regression + final approach.** My interim version (polling `done_waiting` against the machine clock) passed on Windows but, when tested on macOS, regressed `test_Fast_Exp.test_leds` and `test_Fast_Retro.test_matrix_lights`: any manual clock-advancing poll during init shifts the absolute virtual-clock phase, and those tests assert exact intermediate fade bytes that depend on it. (Pure `sleep(0)` instead starves the time-travel loop and hangs.)

The version now pushed keeps init timing parent-identical:
```python
try:
    await asyncio.wait_for(self.done_waiting.wait(), timeout)
    return True
except asyncio.TimeoutError:
    # the time-travel test loop jumps the virtual clock to the timeout timer
    # while an inbound response is still a few parse callbacks out; drain and
    # re-check before declaring a real timeout so we don't resend a reply that
    # arrived (which would double-process it)
    for _ in range(RESPONSE_DRAIN_PASSES):
        await asyncio.sleep(0)
        if self.done_waiting.is_set():
            return True
    return False
```

**Verified on macOS** (Apple Silicon, Py 3.12): the full FAST suite passes (`Ran 27 … OK`), including the two tests that regressed. The drain bound is the one spot I'd flag for your eye — I bisected the worst-case board init at ~6 passes and set `RESPONSE_DRAIN_PASSES = 50` for margin, but a "drain until the loop idles" bound would be more principled if you'd rather (it'd need to peek at the loop's ready queue, which felt too internal for here). Your call — and no rush given you wanted to hardware-test it anyway.

tlpinball:
powershell_PZjPPUoT6I

I implemented the fix on my system and its working 95% of the time, here is the one failure i had

bosh:
> I implemented the fix on my system and its working 95% of the time, here is the one failure i had

Can you consistently cause this error? What are the conditions that cause this in the 5%? If you execute those same conditions without the branch changes, what happens then?
I want to understand if this error is a regression or new, or just an extra level of bad luck

tlpinball:
seems that my error only now occurs if a Ctrl-C in powershell and then try and restart in the same powershell instance. if i close the powershell instance without Ctrl-C and then open a new instance it works. here are some logs of it failing
[startup logs - Pass then Fail - 1.txt](https://github.com/user-attachments/files/28986038/startup.logs.-.Pass.then.Fail.-.1.txt)
[startup logs - Pass then Fail.txt](https://github.com/user-attachments/files/28986040/startup.logs.-.Pass.then.Fail.txt)

tlpinball:
if i execute the same way without the code change it just hangs ~1oo2 times at the EXP board

sjmiles:
The leftover-bytes case @tlpinball hit (Ctrl-C in PowerShell, then restart in the **same** PowerShell instance) is now addressed in `5aafaff8d`.

Root cause: on a hard kill mid LED-blast, the board/OS buffer still holds a burst of leftover bytes with no trailing `<CR>`. `reset_input_buffer()` only flushes what's buffered at that instant — bytes still draining over USB arrive just after and fuse onto the front of the next real message (`<junk>ID:exp ...`). That fused segment either fails to decode (killing the read task) or dispatches with a junk header (dropping the reply), so the handshake reply is lost even though the board *did* answer. Restarting in the same shell reproduces it because the OS port handle isn't torn down, so the in-flight bytes survive.

Two fixes: `_drain_serial()` discards inbound bytes until the stream is quiet (100 ms, capped at 2 s) before the handshake, and `_resync_segment()` recovers a real message fused onto the tail of junk instead of dropping/raising.

@tlpinball if you can re-test with this commit using your same Ctrl-C-then-restart-in-same-shell steps, that'd be a great confirmation.

tlpinball:
[startup logs - commit 5aafaff8d_test1.txt](https://github.com/user-attachments/files/28996418/startup.logs.-.commit.5aafaff8d_test1.txt)
[startup logs - commit 5aafaff8d_test2.txt](https://github.com/user-attachments/files/28996420/startup.logs.-.commit.5aafaff8d_test2.txt)

i have tried all of my startup and hard/soft shutdown/quit methods and have not experienced a failure with this change. 
i can see in the logs that the Fast board is still outputting bad data on some attempts but the change is allowing it to retry and continue on

Thank you

jsm174:
Observing the same behavior with an old v0.57.4.dev3 install running an OPI 5 we made a few years ago for @scottacus64 's Cuphead. 

Ironically, last night I got my first flipper flipping using FAST and that MPF setup.

While learning, was using the starter configs just trying to make a LED blink. Every other attempt, it would hang.

https://github.com/user-attachments/assets/776c7a61-df9b-4768-9781-b3712b82d742

Will start working on upgrading, but glad to see this issue being worked on!

bosh:
Nice! I think I'll have time to test this on hardware tonight.

One note -- I think there was a miscommunication -- with the test file I was saying cut the self-running block that was put on the end of it (`if __name__ == '__main__':` etc) NOT cut the whole test! I think it would be extremely valuable to test what we can here. It's just that the self-run pattern is something that I'd want to add intentionally across all the test files, or not add at all (really I just want consistency), so I didn't want to accidentally start establishing a pattern in unrelated bugfix PRs

sjmiles:
Ah, my mistake — I read "cut the test file" as the whole file rather than just the trailing `if __name__ == '__main__':` block. Restored `test_Fast_Communicator.py` (sans self-run block) in d58aac35e.

Since you're keen on coverage, I also added unit tests for the new drain/resync fix from 5aafaff8d, which had none: `_resync_segment` header recovery, the franken-string parse path (an `ID:` reply fused onto leftover junk), and the connect-time `_drain_serial`. Full FAST suite green (36 tests).

Agreed on keeping the self-run pattern out unless it's applied consistently across the whole suite — happy to follow whatever convention you land on there.

mpf/platforms/fast/communicators/base.py
```diff
@@ -363,15 +439,47 @@ def parse_incoming_raw_bytes(self, msg):
...
+                    if not self.ignore_decode_errors:
+                        raise
+                    continue
 
             if self.port_debug:
                 self.log.info("<<<< %s", msg)
 
             self._dispatch_incoming_msg(msg)
 
+    def _resync_segment(self, raw):
```
bosh: I think this is really cool, but that we actually don't want to recover messages in almost any cases, vs just retrying.

What I can think of:
1. it's a slightly corrupted ID: response -- probably reasonable to use, but effectively no cost to retry
2. it's binary lighting data incompletely written to the line -- it's possible that the binary data could accidentally match a real response (esp with RGB values as the raw bits) and we shouldn't accidentally treat it as valid
3. it's something worse -- the device is in a bad enough state that truly unexpected noise is appearing on the line, and shouldnt be used

But also this is in the base communicator, so we could be getting NET switch data back, or other FAST network types (which I'm less familiar with). Maybe those are reasonable places to do resync recovery on, but I'm not sure. Other than the fact that recovery is possible (for instance, I noticed a case where the corrupted data was a valid ID response shifted by 2 bits), are there specific use cases we really want recovery on, instead of just retrying or bailing?

bosh: I could see us keeping this function around though, for selective usage or wrapped in a debugging option that recovers and prints (maybe not processes) the message

mpf/platforms/fast/communicators/base.py
```diff
@@ -293,15 +339,45 @@ async def send_and_wait_for_response_processed(self, msg, pause_sending_until, t
...
+        cooperative scheduling the timeout and the inbound bytes can become
+        ready in the same loop pass). Let those pending callbacks drain and
+        re-check before declaring a real timeout, so we don't resend — and
+        double-process — a response that actually arrived.
+        """
+        try:
+            await asyncio.wait_for(self.done_waiting.wait(), timeout)
+            return True
+        except asyncio.TimeoutError:
+            for _ in range(RESPONSE_DRAIN_PASSES):
```
bosh: I wonder about the naming -- this is really sleep passing, and the drain is happening elsewhere.

mpf/platforms/fast/communicators/base.py
```diff
@@ -134,6 +144,13 @@ async def connect(self):
 
         await self.clear_board_serial_buffer()
 
+        # reset_input_buffer() above only flushes what was buffered at that
+        # instant. After an unclean shutdown the board can still be clocking out
+        # leftover bytes (often un-terminated binary LED data) that arrive just
+        # afterwards. Drain until the stream goes quiet so the first handshake
+        # response isn't fused onto the tail of that garbage.
+        await self._drain_serial()
```
bosh, line 154: I wonder if `self.clear_board_serial_buffer()` a few lines up is intended to fulfill the same purpose as this new `_drain_serial`. In the Neuron code, it's actually just a pass function (noop), while the Nano sees 4x<CR>.

Since this is the base communicator, I wonder if it's actually wrong to put the specific serial draining we have here in the generic class (and instead move the implementation to the neuron's `clear_board_serial_buffer` override). I don't know enough about the other FAST controller products to say (though I do think the drain_serial function is probably generally safe and cross compatible). Def okay to push back here, I'm just trying to get the architecture juuust right.

mpf/platforms/fast/communicators/base.py
```diff
@@ -11,6 +11,16 @@
...
 
+# After a response wait_for() times out, drain at most this many event-loop
+# passes (re-checking done_waiting) before treating it as a real timeout. Under
+# cooperative scheduling the timeout timer and the inbound response bytes can
+# become ready in the same pass, so wait_for() raises while the reply is still a
+# few parse callbacks from being handled; draining lets it land instead of
+# triggering a spurious resend. Sized well above the observed worst case (a
+# board's full init parse chain); only a genuinely lost response runs the full
+# count.
+RESPONSE_DRAIN_PASSES = 50
```
bosh:
```suggestion
RESPONSE_DRAIN_PASSES = 12
```
let's do 12 and dial it in, if 50 really is leaps and bounds above max realistic.

EDIT: Left this when I thought it was doing 50 drains, not 50 sleep 0 passes. Maybe 20+ is fine, I don't really know how to judge this

mpf/platforms/fast/communicators/base.py
bosh: I think I'd probably slightly rather each callsite decide its own retry, but we've also seen them just not do that for all of time until now, so I think let's keep the default and tweak it if devs report issues.

You can resolve this item noop :)

bosh (dismissed):
I think this is really close, thanks for working with me on it!

sjmiles:
Thanks, this was a good round. Addressed in `4225c88eb`:

**Resync (`_resync_segment`)** — removed it and reverted `parse_incoming_raw_bytes` to just raise/drop. Your point #2 is the convincing one: partial binary lighting data can coincidentally match a real header (RGB bytes as raw bits), and processing that is worse than retrying. Since `_drain_serial` now clears the leftover startup burst *before* the handshake, the real franken-string case is handled there, and the parser no longer guesses. One flag: tlpinball's confirmed build had both drain *and* resync, so I'll ask him to re-confirm the drain-only build to be sure drain+retry fully cover his case.

**`RESPONSE_DRAIN_PASSES` naming** — good catch, the name collided with the actual `_drain_serial`. Renamed to `RESPONSE_SETTLE_PASSES` (it's event-loop yield passes in `_wait_for_processed`, not serial draining) and dialed 50 → 20 (observed worst case ~6; each pass is just an `asyncio.sleep(0)` yield).

**`_drain_serial` placement** — kept it in the base class for now. It's a host-side read-drain (read inbound and discard until quiet), which is orthogonal to `clear_board_serial_buffer`'s job of *writing* board-specific reset bytes to prompt the board — so it's not really the same purpose, and the read-drain is generic across FAST serial types. Happy to move it into a per-board override if you'd rather, but it seemed to belong with the connect logic. Open to your call.

**`max_retries` default** — kept at 3, resolving that thread noop as suggested.

sjmiles:
@tlpinball one more re-test request when you get a chance. The build you confirmed had two mechanisms: the serial **drain** on connect, and a **resync** that recovered a reply fused onto junk. We've since removed the resync (it could misread partial binary LED data as a valid response), so the current branch relies on drain + retry alone.

Could you re-run your same Ctrl-C-in-PowerShell-then-restart-in-the-same-shell steps on the latest commit (`4225c88eb`) and confirm it still recovers cleanly? Want to be sure drain + retry fully cover your case before this merges. Thanks again for all the testing!

bosh:
Hell yes. Now let's both go to bed lol. I just finished backfilling tests for https://github.com/missionpinball/mpf/pull/2028

tl-wfc:
Will run the testing tomorrow and report back

bosh (commented):
I think this is ready! Great work!

I'll test this version in the next day as well, and hopefully have this merged and released as dev3 by Saturday :)

bosh:
Also, it may be that we do want the command recovery option or just a general clear-buffer-between-retries (edit: in a followup after testing on this as is). Octo just reported an issue on dev2, which adds the ER light config block. He gets noise + an ER success (P) seemingly in between header configurations (maybe `ER@840:3` ? So weird because it's not the first header on the board, and nothing should be trying to write lighting info to the buffer at that moment)

```rb
2026-06-17 20:41:24,872 : DEBUG : FASTLED : Creating FAST RGB LED on expansion board at hardware address: 48067
2026-06-17 20:41:24,875 : INFO : EventManager : Event: ======'drop_target_bank_randall_1_bank_up'====== Args={}
2026-06-17 20:41:24,875 : INFO : EventManager : Event: ======'drop_target_bank_randall_2_bank_up'====== Args={}
2026-06-17 20:41:27,255 : INFO : SwitchController : <<<<<<< 's_trough_ball_1' active >>>>>>>
2026-06-17 20:41:27,255 : INFO : SwitchController : <<<<<<< 's_trough_ball_2' active >>>>>>>
2026-06-17 20:41:27,255 : INFO : SwitchController : <<<<<<< 's_trough_ball_3' active >>>>>>>
2026-06-17 20:41:27,255 : INFO : SwitchController : <<<<<<< 's_trough_ball_4' active >>>>>>>
2026-06-17 20:41:27,255 : INFO : SwitchController : <<<<<<< 's_trough_ball_5' active >>>>>>>
2026-06-17 20:41:27,255 : INFO : SwitchController : <<<<<<< 's_boo_home' active >>>>>>>
2026-06-17 20:41:27,285 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@480:0,0,00,20
2026-06-17 20:41:27,286 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@480:1,0,20,20
2026-06-17 20:41:27,286 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@480:2,0,40,20
2026-06-17 20:41:27,286 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@480:3,0,60,20
2026-06-17 20:41:27,317 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@840:0,0,00,20
2026-06-17 20:41:27,317 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@840:1,0,20,20
2026-06-17 20:41:27,317 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@840:2,0,40,20
2026-06-17 20:41:27,317 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@840:3,0,60,20
2026-06-17 20:41:27,333 : WARNING : FAST [EXP] : Interference / bad data received: b'\x825\xffER:P'
2026-06-17 20:41:27,333 : ERROR : asyncio : Exception in callback Util.raise_exceptions() at C:\Users\octo\AppData\Local\Programs\Python\Python313\Lib\site-packages\mpf\core\utility_functions.py:789
handle: <Handle Util.raise_exceptions() at C:\Users\octo\AppData\Local\Programs\Python\Python313\Lib\site-packages\mpf\core\utility_functions.py:789>
  File "C:\Users\octo\AppData\Local\Programs\Python\Python313\Lib\asyncio\events.py", line 89, in _run
    self._context.run(self._callback, *self._args)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\octo\AppData\Local\Programs\Python\Python313\Lib\site-packages\mpf\core\utility_functions.py", line 797, in raise_exceptions
    future.result()
    ~~~~~~~~~~~~~^^
  File "C:\Users\octo\AppData\Local\Programs\Python\Python313\Lib\site-packages\mpf\platforms\fast\communicators\base.py", line 413, in _socket_reader
    self.parse_incoming_raw_bytes(resp)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "C:\Users\octo\AppData\Local\Programs\Python\Python313\Lib\site-packages\mpf\platforms\fast\communicators\base.py", line 360, in parse_incoming_raw_bytes
    msg = msg.decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x82 in position 0: invalid start byte
2026-06-17 20:41:27,336 : INFO : Machine : Starting the main run loop with active modes: []
2026-06-17 20:41:27,336 : WARNING : Machine : Attract mode is not active, game will not be playable. Please check your attract mode configuration.
2026-06-17 20:41:27,336 : INFO : Machine : Shutting down...
2026-06-17 20:41:27,337 : INFO : EventManager : Event: ======'shutdown'====== Args={}
2026-06-17 20:41:27,348 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@841:0,0,00,20
2026-06-17 20:41:27,348 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@841:1,0,20,20
2026-06-17 20:41:27,349 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@841:2,0,40,20
2026-06-17 20:41:27,349 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@841:3,0,60,20
2026-06-17 20:41:27,380 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@900:0,0,00,20
2026-06-17 20:41:27,380 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@900:1,0,20,20
2026-06-17 20:41:27,381 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@900:2,0,40,20
2026-06-17 20:41:27,381 : DEBUG : FAST [EXP] : Sending LED port config to breakout. ER@900:3,0,60,20
```

sjmiles:
Agreed — let's merge this as-is and take the ER-noise as a followup. A few notes from digging into octo's trace so the followup is well-scoped:

- It's the same junk-fused-onto-a-reply pattern (`\x82\x35\xff` + `ER:P`), but it lands mid-init during the `ER@` light-config block, *after* connect — so `_drain_serial` (connect-only) won't catch it, and it'd repro on this branch too. Fair to treat as separate.
- That `ER@` light config uses `send_with_confirmation(..., 'ER:P')` (fire-and-pause), not the retry path — so there's no timeout there. A corrupted/lost `ER:P` either crashes the reader on decode or stalls sending on the pause flag.

So for the followup I'm leaning toward your clear-buffer-between-retries idea over content recovery: (a) make a decode error non-fatal (drop + warn instead of killing `read_task`), and (b) bound the confirmation pause with a timeout that clears the input buffer and resends once. That avoids the false-match risk of recovering binary as a real response.

I'll prototype it on a side branch and validate on my Mac + with octo/tl before opening anything, once this one's merged. Sound right?

sjmiles:
Followup for octo's ER-noise crash is now its own PR: #2051. It's independent of this one (the crash repros on dev2 without these changes), CI is green, and it simply stops a `UnicodeDecodeError` mid-init from killing the reader. Keeping it separate so this PR stays focused on the #2036 timeout/drain work.

bosh:
Thanks! I played around with this a bit last night, and things seemed fine on my test setup. I'll test on a fully built machine later today.

However, I also haven't been able to get the dev code without the fix to hang or have EXP communication issues. My setup is an 0081 with two light strings on headers 1 and 2, and an 8 digit segment display on header 4, with a random event player constantly changing the segment text with a random event player on a 10ms loop (trying to get partially-written light info to be left over in the buffer). Any tips on how I can actually cause a bad state to prove the hang?

sjmiles:
@bosh what OS?

bosh:
My test bench is on my main Windows PC, but my game is run on Linux

sjmiles:
Here's a minimal repro for the EXP startup hang. Confirmed on my hardware: it **hangs every other run** on unpatched `dev`, and boots straight through with this PR.

It's stripped to just the Neuron's NET + EXP buses and its built-in EXP LEDs, driven by a continuous looping color show — so there's constant binary `RD@` traffic on the EXP bus. No coils/switches/flippers/modes. The continuous LED stream is the key: it's what leaves the board mid-message when you kill MPF.

**Repro steps (single terminal):**
1. `python -m mpf <this-machine-folder> -t`
2. Let the EXP LEDs cycle for ~5s.
3. Ctrl-C while they're still cycling.
4. Immediately re-run in the **same** shell (don't close the window — keeping the OS COM handle open is what leaves the board mid-message).
5. Repeat a few times. On `dev` it freezes during EXP init every other run or so — last log line around the EXP connect, no error, no further output. On this PR it logs `Timeout waiting for response to ID@..., Retrying...` (and/or drains leftover bytes) and continues.

Adjust the two COM ports for your Neuron (and add your external EXP board(s) under `boards:` if you want — the built-in `FP-EXP-2000` alone reproduces it):

```yaml
#config_version=6
hardware:
  platform: fast

fast:
  net:
    controller: neuron
    port: com5            # <-- your Neuron NET port
    baud: 921600
  exp_int:
    port: com6            # <-- your Neuron EXP port
    boards:
      neuron:
        model: FP-EXP-2000

machine:
  balls_installed: 0
  min_balls: 0

lights:
  led0: {number: neuron-1-1, type: rgb}
  led1: {number: neuron-1-2, type: rgb}
  led2: {number: neuron-1-3, type: rgb}
  led3: {number: neuron-1-4, type: rgb}

show_player:
  machine_reset_phase_3:
    exp_stream:
      loops: -1
      speed: 1

shows:
  exp_stream:
    - duration: 42ms
      lights:
        led0: {color: red,   fade: 42ms}
        led1: {color: green, fade: 42ms}
        led2: {color: blue,  fade: 42ms}
        led3: {color: white, fade: 42ms}
    - duration: 42ms
      lights:
        led0: {color: green, fade: 42ms}
        led1: {color: blue,  fade: 42ms}
        led2: {color: white, fade: 42ms}
        led3: {color: red,   fade: 42ms}
    - duration: 42ms
      lights:
        led0: {color: blue,  fade: 42ms}
        led1: {color: white, fade: 42ms}
        led2: {color: red,   fade: 42ms}
        led3: {color: green, fade: 42ms}
    - duration: 42ms
      lights:
        led0: {color: white, fade: 42ms}
        led1: {color: red,   fade: 42ms}
        led2: {color: green, fade: 42ms}
        led3: {color: blue,  fade: 42ms}
```

bosh:
No dice, that config works perfectly well between CtrlC'd restarts on unmodified `dev`. I tried on neuron as well as a 0081, no problems either way. Tried different fades as well. Maybe I'll try a 0071 later. (edit: 0071 had no issues on plain dev either)

One thing to note is that your example used the `exp_int` subconfig instead of `exp`. In practice this is fine, but that one is actually intended for the raspberry pi's direct-mount interface to the Neuron EXP net (which requires [exp_int ](https://missionpinball.org/latest/config/fast/fast_exp/#using-exp_int)as a workaround to get the normal EXP network PLUS the Neuron direct)

sjmiles:
Tracked down why it hangs, and it's a **board-firmware behavior, not MPF** — reproduced in CoolTerm with MPF entirely out of the picture.

**Observed** — Neuron `FP-CPU-2000` v02.26, EXP `FP-EXP-2000` **fw 0.48**, Windows, 921600 baud. Typed by hand in CoolTerm:
- `br@48:` ×15 → acks come back in a steady period-3 pattern: `BR:P`, `BR:P`, *(silent)*, repeating (~6/15 with no ack).
- `br@480:` ×15 → same period-3 pattern, clean from the first command (5/15 silent).
- `id@48:` ×15 → **15/15** acked.

So the board reliably answers `ID@` but **drops roughly every 3rd `BR@` ack.** That's exactly what init hangs on: `reset()` sends `BR@` and waits for `BR:P`; on a dropped beat the ack never comes, so unpatched MPF waits forever. It also explains the periods we saw — a period-3 drop × the number of `BR@` per boot/shutdown cycle (which is why removing the shutdown `BR@` shifted every-other → every-third, and why *waiting* for `BR:P` didn't help: the ack genuinely never comes).

**Hypothesis:** an internal board cadence (counter/rate?) we can't see from the host — and quite possibly **firmware-version-specific**, which would explain your no-repro. Mine is `FP-EXP-2000` **fw 0.48**; you mentioned testing a `0081`.

**Could you check your firmware version** — and, if you have a minute, try the same in a terminal? Just send `br@<addr>:` ~15× to one of your EXP boards and see whether you get the same `BR:P, BR:P, silent` period-3, or a clean ack every time. If yours never drops, that pins it as firmware-specific.

Either way, MPF can't force the board to ack, so the correct MPF-side behavior is to not hang on it — which is what this PR does. I'd love to get the `BR@`-ack-drop in front of FAST as well; I haven't had luck reaching them directly, so if you have a line to them, the CoolTerm repro above is self-contained for their firmware folks.

sjmiles:
Root cause is a FAST firmware behavior, now tracked at fastpinball/fast-firmware#2. Measured on our `FP-EXP-2000` (fw 0.48): the board drops **every 3rd `BR@`** (count-based, persists across idle — still every-3rd at 30s spacing), plus a fixed ~1.19s dead window after each drop. `ID@`/idle/`FP-EXP-0081` unaffected (→ the no-repro on `0081`). MPF's init `reset()` waits on a `BR:P` that's dropped when the init `BR@` lands on the every-3rd → the #2036 hang; the persistent count makes it deterministic and wait-independent.

This PR's retry is the right MPF-side mitigation; the firmware fix is tracked upstream.

One correction to the record: our earlier comments here pinned this on leftover binary LED data from an unclean mid-LED-blast shutdown fusing onto replies. That was wrong — this hang is the `BR@` firmware behavior above, unrelated to LED traffic or leftover bytes (the LED streaming in the repro turned out to be irrelevant; it hangs simply because each run sends `BR@`). The separate decode crash in #2051 is its own issue.

bosh:
I'll try manual BRs in a bit. I mentioned trying 0071, 0081, and Neuron all, with no issues on just dev (and no issues on this branch as well  :) )

All my firmwares are updated, I wrote the version bump to .48 into MPF and the ER command support that necessitated it.

bosh:
Bad news is that my Neuron, 71, and 81 all do not have any issue with repeated BR:s. I'm on a Neuron 2000-5, and 0081-2 and 0071-2

sjmiles:
Not having a repro always sux. I'm on a Neuron 2000-7, that's the only difference I'm aware of at this point.

bosh (approved):
Lets do it :) I have a neuron-7 and some testers to help in case anything comes up

PR #2051, opened 2026-06-19
FAST: don't crash the serial reader on an undecodable message

sjmiles:
## Problem

When MPF receives an undecodable serial segment after connect, `parse_incoming_raw_bytes()` re-raises the `UnicodeDecodeError` (since `ignore_decode_errors` is `False` post-connect). That exception propagates out of `_socket_reader` and, via the `read_task`'s `Util.raise_exceptions` done-callback, **crashes MPF**.

This happens when line noise or leftover binary fuses onto the front of a real reply — e.g. octo hit it mid-init during the EXP `ER@` light-config block, where `b'\x82\x35\xffER:P'` (junk + a real `ER:P`) failed to decode and took the whole process down:

```
WARNING : FAST [EXP] : Interference / bad data received: b'\x825\xffER:P'
...
  File ".../fast/communicators/base.py", line 360, in parse_incoming_raw_bytes
    msg = msg.decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x82 in position 0: invalid start byte
```

## Fix

Drop the corrupted segment and keep reading instead of re-raising, so a single noisy read can't kill the reader. The send/confirmation path re-requests anything genuinely lost. Unexpected (non-ignored) junk is still logged as a warning so it stays visible.

## Tests

`mpf/tests/test_Fast_Communicator.py` drives `parse_incoming_raw_bytes()` directly: a corrupted segment is dropped and a following valid message in the same buffer is still dispatched (i.e. the read loop survives); junk during the connect window is dropped quietly. Full FAST suite passes.

## Relationship to #2043

Independent of #2043 (this crash reproduces on released dev2, without it) and touches a different code path, so it can land on its own. Heads up: #2043 also adds a `test_Fast_Communicator.py` (covering the timeout/retry + drain work). Whichever merges second, I'll reconcile the two test files — they're both mine and the merge is trivial.

### DIFF
--- a/mpf/platforms/fast/communicators/base.py
+++ b/mpf/platforms/fast/communicators/base.py
@@ -441,9 +441,14 @@ class FastSerialCommunicator(LogMixin):
                 if self.machine.is_shutting_down:
                     return
 
-                self.log.warning("Interference / bad data received: %s", msg)
+                # Line noise or leftover binary can appear mid-stream (e.g. a
+                # board left mid-message by an unclean shutdown, fusing junk onto
+                # a real reply). Drop the corrupted segment and keep reading
+                # rather than raising, which would kill read_task and crash MPF.
+                # The send/confirmation path re-requests anything genuinely lost.
                 if not self.ignore_decode_errors:
-                    raise
+                    self.log.warning("Interference / bad data received, dropping: %s", msg)
+                continue
 
             if self.port_debug:
                 self.log.info("<<<< %s", msg)
--- a/mpf/platforms/fast/communicators/base.py
+++ b/mpf/platforms/fast/communicators/base.py
@@ -446,8 +446,8 @@ class FastSerialCommunicator(LogMixin):
                 # a real reply). Drop the corrupted segment and keep reading
                 # rather than raising, which would kill read_task and crash MPF.
                 # The send/confirmation path re-requests anything genuinely lost.
-                if not self.ignore_decode_errors:
-                    self.log.warning("Interference / bad data received, dropping: %s", msg)
+                # Always warn so the underlying noise stays visible for debugging.
+                self.log.warning("Interference / bad data received, dropping: %s", msg)
                 continue
 
             if self.port_debug:
--- a/mpf/tests/test_Fast_Communicator.py
+++ b/mpf/tests/test_Fast_Communicator.py
@@ -1,7 +1,7 @@
 # mpf.tests.test_Fast_Communicator
 """Unit tests for FastSerialCommunicator's response and connect/parse paths.
 
-Two related FAST startup failures are covered here:
+Three related FAST startup failures are covered here:
 
 * Response timeout/retry (#2036): a single lost serial response used to freeze
   MPF init forever, because the asyncio timeout was wrapped around
@@ -9,9 +9,12 @@ Two related FAST startup failures are covered here:
   immediately) while the real wait on done_waiting had no timeout.
 * Connect-time drain: after an unclean shutdown the board can clock out
   leftover, un-terminated bytes that would otherwise fuse onto the front of the
-  next real message. connect() drains that burst before the handshake; the
-  parser itself still raises (or drops, when decode errors are ignored) on
-  undecodable data rather than guessing at it.
+  next real message. connect() drains that burst before the handshake.
+* Undecodable data mid-stream (#2051): a UnicodeDecodeError in
+  parse_incoming_raw_bytes used to be re-raised after connect, which kills
+  read_task via its raise_exceptions done-callback and crashes MPF whenever
+  line noise or leftover binary fuses onto a real reply. The parser now drops
+  the corrupted segment (with a warning) and keeps reading.
 
 These drive the relevant methods directly with a stubbed I/O surface so the
 paths are exercised deterministically without a real serial port.
@@ -148,16 +151,19 @@ class TestFastCommunicatorRetry(unittest.TestCase):
 
 class TestFastCommunicatorInbound(unittest.TestCase):
 
-    def test_parse_raises_on_undecodable_data(self):
-        """Un-decodable data raises during init (ignore_decode_errors False), so
-        a board in a bad state surfaces loudly instead of being guessed at."""
+    def test_drops_undecodable_data_and_keeps_reading(self):
+        """A corrupted segment is dropped rather than raised, so a single noisy
+        read can't kill the read loop (which would crash MPF). A following valid
+        message in the same buffer is still dispatched."""
         comm = _make_comm()
         comm.ignore_decode_errors = False
-        comm.message_processors = {'ID:': lambda m: None}
-        with self.assertRaises(UnicodeDecodeError):
-            comm.parse_incoming_raw_bytes(b'\x81\xb5\xc1\r')
+        received = []
+        comm.message_processors = {'ID:': received.append}
+        # junk fused onto an ER:P reply, then a clean ID: response
+        comm.parse_incoming_raw_bytes(b'\x825\xffER:P\rID:exp fp-exp-0081 0.48\r')
+        self.assertEqual(received, ['exp fp-exp-0081 0.48'])
 
-    def test_parse_drops_undecodable_data_when_ignoring(self):
+    def test_drops_undecodable_data_when_ignoring(self):
         """With ignore_decode_errors set (e.g. during connect), undecodable
         data is dropped rather than raised."""
         comm = _make_comm()


sjmiles:
@bosh heads-up — this is the small independent fix for the crash octo reported on #2043 (the `b'\x82\x35\xffER:P'` decode error during EXP light config). It's deliberately scoped to just stop a `UnicodeDecodeError` from killing `read_task`; it doesn't depend on #2043 and the crash repros on released dev2, so it stands on its own. CI's green across the matrix.

Could you loop octo in to confirm on hardware when convenient? That's the real-world validation. And if it's easy, feel free to fold it into the dev3 batch alongside #2043.

One note carried over from the description: this also adds a `test_Fast_Communicator.py`, which overlaps the one #2043 introduces — whichever lands second I'll reconcile the two (both mine, trivial).

mpf/platforms/fast/communicators/base.py
```diff
@@ -363,9 +363,14 @@ def parse_incoming_raw_bytes(self, msg):
                 if self.machine.is_shutting_down:
                     return
 
-                self.log.warning("Interference / bad data received: %s", msg)
+                # Line noise or leftover binary can appear mid-stream (e.g. a
```
bosh, line 444: We shouldnt stop warning, that'll just make debugging harder in the future.

mpf/platforms/fast/communicators/base.py
```diff
@@ -363,9 +363,14 @@ def parse_incoming_raw_bytes(self, msg):
...
+                # board left mid-message by an unclean shutdown, fusing junk onto
+                # a real reply). Drop the corrupted segment and keep reading
+                # rather than raising, which would kill read_task and crash MPF.
+                # The send/confirmation path re-requests anything genuinely lost.
                 if not self.ignore_decode_errors:
```
bosh: I think just using this option is the actual answer for Octo if he is the only one reporting this specific timing of noise on the line.

bosh (commented):
I don't think this is the right approach. I was investigating a bit myself, and my angle was on that I implemented ER and didn't have a specific reason to choose
`communicator.send_with_confirmation`
over the newly improved
`communicator.send_and_wait_for_response_processed`
and maybe our improved method would be better.

I've been playing with those two (and the `_process_er` hook) but don't think new version helped. I did expect send_and_wait to be better, especially as it adds retries, but it corrupted the commands somehow and ended up incorrectly changing the ER effects in a way I can't explain (definitely the light strings were not rechanneled correctly).

sjmiles:
Thanks for digging in, @bosh — and good to know the `ER@` → `send_and_wait_for_response_processed` route corrupted the channeling; that matches my read that the ER path is touchier than it looks (part of why I kept this PR off it).

I'd frame #2051 a bit differently than "the octo fix": it's really a defensive guard — `parse_incoming_raw_bytes` shouldn't be able to take down `read_task` (and the whole process) on *any* undecodable byte, whatever the source. With the warning kept, we're not masking anything — we just don't crash, and the existing confirmation/retry machinery re-requests whatever was lost. So it stands on its own as robustness, independent of whether octo's the only one seeing this particular noise.

Restored the always-warn per your comment (460ce0f45). Happy to dig into *why* the ER retry-path conversion re-channeled incorrectly as a separate thread — that smells like a real ordering/state bug worth understanding — but it'd need your repro or hardware, since the Windows/CI suite doesn't surface it.

mpf/platforms/fast/communicators/base.py
```diff
@@ -363,9 +363,14 @@ def parse_incoming_raw_bytes(self, msg):
                 if self.machine.is_shutting_down:
                     return
 
-                self.log.warning("Interference / bad data received: %s", msg)
-                if not self.ignore_decode_errors:
```
bosh, line 445: I dont think removing the ignore_decode_errors option is right in general. But if it turns out to be unhelpful and removal is right, we need to also clean up all the places that set it, since this is the only actual conditional logic based on it

bosh (changes requested):
My comment last night was somewhat mistaken - there is no direct config control over the flag value, so Octo can't actually just flip something to ignore noise trivially. (I was thinking of the FAST EXP board option ignore_led_errors https://github.com/missionpinball/mpf/blob/447feba9507ff99974ad13f647cc6b43b3da2a00/mpf/config_spec.yaml#L589 ).

I think we need to figure out whether this is electrical noise on the line or is serial communications stepping on each other. If it is noise, maybe we need a FAST config setting that allows a game builder to just ignore it. But if it is bad serial message sequencing code, then we might need to rethink larger parts of the FAST communicators.

sjmiles:
Good question to nail down before we settle the shape. I traced the init phases and I think we can **rule out "serial messages stepping on each other"** — at least MPF stepping on itself:

- The `ER@` light config runs in `soft_reset` → **`init_phase_1`** (`fast.py` `add_async_handler('init_phase_1', self.soft_reset)`).
- The `RD@` binary LED writes (`update_leds`) are only scheduled by `start_tasks` at **`init_phase_3`** (`_start_communicator_tasks`).

phase 1 < phase 3, so `update_leds` isn't even scheduled yet when the `ER@` config runs — MPF isn't writing any lighting data at that moment, which matches your instinct that "nothing should be writing to the buffer there." So the `\x82\x35\xff` isn't our own `RD@` binary from this session.

It's a stray 3-byte fragment fused onto an **intact** `ER:P` (the response itself is fine). That points to leftover/external bytes rather than garbled sequencing — most likely **leftover `RD@` LED binary from a prior unclean shutdown** (octo's previous run *was* blasting LEDs) surfacing mid-init just after the connect-time drain window — i.e. the same family as tlpinball's franken-string, only landing a little later. The other candidate is plain electrical noise. Either way it doesn't look like "rethink the communicators" territory.

To tell noise vs leftover apart, the deciding data point is octo's: **does it happen on a cold boot, or only after a Ctrl-C / unclean restart?** (And `port_debug` on to capture the raw inbound around the event.) Cold-boot-too → electrical/board noise → a FAST config setting to tolerate it makes sense. Only-after-unclean → leftover data → drain/guard territory.

On `ignore_decode_errors`: you're right that this change orphaned it — that conditional was its only reader. Since the guard's whole point is *not* to crash the reader, the flag has nothing left to gate, so I'd either (a) remove it and clean up its setters, or (b) repurpose it as the real config-controlled "tolerate noise" knob you mentioned (better as a followup, and it lines up with the noise case if that's what octo's data shows). Which way do you prefer?

Regardless of a/b: a `UnicodeDecodeError` taking down `read_task` and the whole process is never the right outcome, and the warning stays — so I think the guard itself should land. Happy to do the flag cleanup whichever way you call it.

sjmiles:
Picking this back up with what we learned chasing the related EXP hang (#2043 / fastpinball/fast-firmware#2):

On **"noise vs sequencing"** — we ruled out MPF stepping on itself: the `ER@`/LED-port config runs in `soft_reset` (init_phase_1) while the `RD@` LED task is only scheduled later (init_phase_3), so nothing of ours is writing lighting data during config. The EXP serial issues are firmware-side — e.g. for the #2036 hang we measured the board silently dropping every 3rd `BR@` ack. octo's `\x82\x35\xffER:P` decode crash is a *separate* junk-on-the-line case we didn't fully root-cause, but it's the same family (FAST EXP firmware quirks), not a communicator bug to "rethink." So the minimal correct MPF behavior is just: a stray undecodable byte must not be able to kill `read_task` and take down the whole process.

On **`ignore_decode_errors`** — you're right it's now vestigial (the raise was its only reader). Two ways to resolve, your call:

**(a) Keep it, give it a real job:** gate the *log level* — during the connect/clear window (where junk is expected) log at debug; elsewhere log at warning. Keeps the flag meaningful and keeps the warning exactly where it's useful (unexpected interference), nothing masked.

**(b) Remove it + clean up the setters** (`__init__`, `connect`, `__slots__`) since nothing reads it anymore.

I lean (a), since you said you'd rather not drop the option and it keeps the warning for the cases that matter. Tell me which and I'll push it.

bosh:
> Good question to nail down before we settle the shape. I traced the init phases and I think we can **rule out "serial messages stepping on each other"** — at least MPF stepping on itself:
> 
>     * The `ER@` light config runs in `soft_reset` → **`init_phase_1`** (`fast.py` `add_async_handler('init_phase_1', self.soft_reset)`).
> 
>     * The `RD@` binary LED writes (`update_leds`) are only scheduled by `start_tasks` at **`init_phase_3`** (`_start_communicator_tasks`).
> 
> 
> phase 1 < phase 3, so `update_leds` isn't even scheduled yet when the `ER@` config runs — MPF isn't writing any lighting data at that moment, which matches your instinct that "nothing should be writing to the buffer there." So the `\x82\x35\xff` isn't our own `RD@` binary from this session.
> 
> It's a stray 3-byte fragment fused onto an **intact** `ER:P` (the response itself is fine). That points to leftover/external bytes rather than garbled sequencing — most likely **leftover `RD@` LED binary from a prior unclean shutdown** (octo's previous run _was_ blasting LEDs) surfacing mid-init just after the connect-time drain window — i.e. the same family as tlpinball's franken-string, only landing a little later. The other candidate is plain electrical noise. Either way it doesn't look like "rethink the communicators" territory.
> 
> To tell noise vs leftover apart, the deciding data point is octo's: **does it happen on a cold boot, or only after a Ctrl-C / unclean restart?** (And `port_debug` on to capture the raw inbound around the event.) Cold-boot-too → electrical/board noise → a FAST config setting to tolerate it makes sense. Only-after-unclean → leftover data → drain/guard territory.
> 
> On `ignore_decode_errors`: you're right that this change orphaned it — that conditional was its only reader. Since the guard's whole point is _not_ to crash the reader, the flag has nothing left to gate, so I'd either (a) remove it and clean up its setters, or (b) repurpose it as the real config-controlled "tolerate noise" knob you mentioned (better as a followup, and it lines up with the noise case if that's what octo's data shows). Which way do you prefer?
> 
> Regardless of a/b: a `UnicodeDecodeError` taking down `read_task` and the whole process is never the right outcome, and the warning stays — so I think the guard itself should land. Happy to do the flag cleanup whichever way you call it.

I agree, likely not lighting data. Leftover data from before doesnt make sense to me, as why would it surface several messages deep into the board setup processes?
I think we really can't make much decision and need more info on frequency from Octo. Maybe I can get him to hack in more logging statements.

bosh:
> Picking this back up with what we learned chasing the related EXP hang (#2043 / [fastpinball/fast-firmware#2](https://github.com/fastpinball/fast-firmware/issues/2)):
> 
> On **"noise vs sequencing"** — we ruled out MPF stepping on itself: the `ER@`/LED-port config runs in `soft_reset` (init_phase_1) while the `RD@` LED task is only scheduled later (init_phase_3), so nothing of ours is writing lighting data during config. The EXP serial issues are firmware-side — e.g. for the #2036 hang we measured the board silently dropping every 3rd `BR@` ack. octo's `\x82\x35\xffER:P` decode crash is a _separate_ junk-on-the-line case we didn't fully root-cause, but it's the same family (FAST EXP firmware quirks), not a communicator bug to "rethink." So the minimal correct MPF behavior is just: a stray undecodable byte must not be able to kill `read_task` and take down the whole process.
> 
> On **`ignore_decode_errors`** — you're right it's now vestigial (the raise was its only reader). Two ways to resolve, your call:
> 
> **(a) Keep it, give it a real job:** gate the _log level_ — during the connect/clear window (where junk is expected) log at debug; elsewhere log at warning. Keeps the flag meaningful and keeps the warning exactly where it's useful (unexpected interference), nothing masked.
> 
> **(b) Remove it + clean up the setters** (`__init__`, `connect`, `__slots__`) since nothing reads it anymore.
> 
> I lean (a), since you said you'd rather not drop the option and it keeps the warning for the cases that matter. Tell me which and I'll push it.

I'm inclined to do nothing right now. I think that raising an error is reasonable for many cases, and seems to not be causing anyone else issues. So we need more info on the specifics of this user issue, and not to spin our wheels jumping to early conclusions or making solutions in search of a problem. My apologies for prematurely raising this issue without enough information

sjmiles:
No need to apologize — agreed, better to wait for real signal from octo than ship a guess. Happy to leave this parked.

The most useful data from him would be: (1) does it happen on a **cold boot**, or only after a **Ctrl-C / unclean restart**, and (2) a `port_debug` capture of the raw inbound around the crash. Whenever there's enough signal, the guard (or a config-controlled "tolerate noise" knob) is a quick follow-up.

bosh:
Multiple people are running into the unexpected byte sequence `b'\x82\x35\xffER:P'` and a couple have reported that this fix works. I'm just hesitant to merge this and cover up what seems like a consistent message on the wire that isn't being handled correctly. If the "noise" was actually random or occurred at other points in time during startup, I'd probably have it in already. But for now, since 81 is still a dev release, not a full release, I'm going to keep the error live.

If you do have a chance to rebase on the latest dev just once (dont need to keep bumping it), it would help me out, and if I dont have a better fix for the specific byte sequence error before 81 wants to go to full release, I'll merge it.

sjmiles:
@bosh Rebased onto latest dev as requested (one-time bump, per your note). Two things folded in:

- The test-file reconciliation flagged in the description: #2043's `test_Fast_Communicator.py` is now on dev, so this branch's version merged into it — its retry/drain tests are kept intact, and only `test_parse_raises_on_undecodable_data` (which asserted the old raise-and-crash behavior this PR removes) is replaced by this PR's drop-and-keep-reading test.
- `base.py` carried over cleanly, including the always-warn behavior from review.

CI is green across the full matrix and the PR shows mergeable. Parked here until the 0.81 call.

sjmiles:
@bosh I think we can now root-cause the byte sequence we'd left open — the "consistent message on the wire."

fastpinball/fast-firmware#2 documents that the affected FP-EXP-2000 **emits a short garbage burst on TX whenever its processor resets** — including after a successful `BR@`: *"an acked `BR@48:` is followed by one reset cycle (~1.5s unresponsive, garbage burst, then stable again)."*

Line that up with our init: we send `BR@{addr}:` during board verification (which per the issue *triggers* a reset on the affected board), then immediately proceed into the `ER@` light-config block. ~1.2–1.5s after the `BR:P` ack, the reset burst arrives on the inbound stream — right in the middle of the `ER@` exchange, fusing onto the front of the next intact `ER:P`. That's exactly the observed failure: junk + intact reply, always mid-`ER@`-block.

It also explains why multiple users see the *identical* bytes: a deterministic reset produces the same TX artifact every time, on every affected board. (That consistency is also what rules out my earlier leftover-LED-data theory — different machines' RGB data would differ.)

FAST has a pre-release firmware (Eli's candidate build for the fast-firmware#2 reset bug) that should solve this at the source — with the reset behavior gone, the garbage burst and this byte sequence go with it.

PR #2056, opened 2026-07-04
Conditional-only handler cleanup

bosh:
Conditional-only event listeners attach to their relevant AST
nodes for player variables or other variables as appropriate.
Specifically with player-related bindings, additional subscriptions
are made to player turn rotation events, to cancel the future and dump
the handler.

The issue is that with these several subscriptions, only one will ever
occur, and the others need to be cleaned up. This is addressed by
adding cancellation signals to the unused hooks when the parent
future is done.

```yaml
event_player:
  {current_player.score > 0 and current_player.score < 10000}: keep_going
```
So in this example, every time the player's score changes, _two_ subscriptions to the current player's score player variable change event are made, as well as two subscriptions to player turn rotation events (to ensure the listeners are cleaned up if unused.)

The fix is that any of the subscribed events should cause the rest to cancel, so that futures arent left waiting around until the ball end turn rotation catches up to them.

Note that this PR does NOT fix that the above example binds to current_player.score two times. It's weird, but it also doesn't multi-trigger the handler, so it's only as bad as the number of times a dev repeats the same variable in the same conditional hook line.

### DIFF
--- a/mpf/core/events.py
+++ b/mpf/core/events.py
@@ -458,6 +458,8 @@ class EventManager(MpfController):
                                                              _future=future,
                                                              _keys=keys,
                                                              event=event_name)))
+
+        future.add_done_callback(lambda f: [self.remove_handler_by_key(k) for k in keys])
         return future
 
     def _wait_handler(self, _future: asyncio.Future, _keys: List[EventHandlerKey], **kwargs):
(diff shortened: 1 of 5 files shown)


avanwinkle (approved):
why not?

PR #2079, opened 2026-08-11
Reset combo switch state when its mode stops

Pyrrvs:
## Problem

A mode-scoped `combo_switches` device fires **exactly once per MPF process** when its own event ends the mode that owns it.

`ComboSwitch` keeps three pieces of state: `_switches_1_active`, `_switches_2_active` and `_state`. `device_removed_from_mode` removes the switch handlers and kills the pending delays, but leaves all three untouched.

The switches are typically still held at that moment. That is the normal case for any combo whose own event ends the mode that owns it. Consider a "hold both buttons to cancel the game" combo defined in a game mode: `events_when_both` posts `end_game`, that stops the mode synchronously, and only afterwards does the player let go. With the handlers already removed the release is never seen, so the device stays latched at `both` with both groups marked active for the life of the process.

Every later attempt is then dropped twice over:

* `_switch_1_went_active` and `_switch_2_went_active` each guard on `if self._switches_N_active: return`, so neither group ever re-arms.
* `_switch_state` guards on `if state == self.state: return`, so even a re-armed group could not re-post `events_when_both`.

Reloading the mode does not clear it either. `add_switch_handler_obj` only back-fills switches that are already in the requested state for handlers registered with `ms > 0`, and `ComboSwitch` registers with `ms=0`, so re-registering the handlers on the next mode start does not re-arm anything.

## Reproduction

On a machine, hold LEFT FLIPPER + START to cancel a game, using a combo defined in a game mode. It works. Start another game, repeat the identical gesture, and nothing happens for the rest of the machine's uptime.

The same shape affects any combo whose event stops its own mode, including an attract-mode gesture that starts a game.

Observed on a live machine. The relevant ordering, measured from the moment both buttons went down:

```
+2.002s  end_game            (combo fires)
+2.003s  ball_will_end
+2.006s  mode_..._stopped    <- handlers removed, buttons STILL DOWN
+2.008s  buttons released    <- nobody is listening any more
```

## Fix

Reset the three attributes in `device_removed_from_mode`. An unloaded device observes nothing, so holding no state is the only accurate option, and a reload re-arms from the next real switch change.

## Test

`test_combo_switches_in_mode` already covers mode-scoped combos, but it releases the switches *before* stopping the mode, so it never reached this path. The new `test_combo_switches_in_mode_stopped_while_held` stops the mode with both switches still down, releases them, restarts the mode, and asserts the identical gesture works again.

| | `test_ComboSwitches.py` |
|---|---|
| without the fix | 1 failed, 10 passed |
| with the fix | 11 passed |

Full suite on this branch: **895 passed, 15 skipped, no failures.** (The 34 collection errors are pre-existing `fixture 'config_file' not found` artifacts from the helper functions in `MpfTestCase.py`, present with or without this change.)

### DIFF
--- a/mpf/devices/combo_switch.py
+++ b/mpf/devices/combo_switch.py
@@ -88,16 +88,7 @@ class ComboSwitch(SystemWideDevice, ModeDevice):
         self._remove_switch_handlers()
         self._kill_delays()
 
-        # The switches may still be held when the mode stops. This happens
-        # whenever the combo's own event ends the mode that owns it, e.g. a
-        # "hold both buttons to cancel the game" combo defined in a game mode.
-        # With the handlers gone the eventual release is never seen, so any
-        # state left here would latch for the life of the process:
-        # _switch_1_went_active and _switch_2_went_active both return early on
-        # "if self._switches_N_active", and _switch_state returns early on
-        # "if state == self.state". The combo would then fire exactly once and
-        # never again. The device observes nothing while unloaded, so the only
-        # honest state to hold is none at all.
+        # Ensure reset to inactive state when mode ends, in case combo switch event is reason mode ended.
         self._state = 'inactive'
         self._switches_1_active = False
         self._switches_2_active = False
--- a/mpf/devices/combo_switch.py
+++ b/mpf/devices/combo_switch.py
@@ -88,6 +88,20 @@ class ComboSwitch(SystemWideDevice, ModeDevice):
         self._remove_switch_handlers()
         self._kill_delays()
 
+        # The switches may still be held when the mode stops. This happens
+        # whenever the combo's own event ends the mode that owns it, e.g. a
+        # "hold both buttons to cancel the game" combo defined in a game mode.
+        # With the handlers gone the eventual release is never seen, so any
+        # state left here would latch for the life of the process:
+        # _switch_1_went_active and _switch_2_went_active both return early on
+        # "if self._switches_N_active", and _switch_state returns early on
+        # "if state == self.state". The combo would then fire exactly once and
+        # never again. The device observes nothing while unloaded, so the only
+        # honest state to hold is none at all.
+        self._state = 'inactive'
+        self._switches_1_active = False
+        self._switches_2_active = False
+
     def _register_switch_handlers(self):
         for switch in self.config['switches_1']:
             self._switch_handlers.append(switch.add_handler(self._switch_1_went_active, state=1, return_info=True))
--- a/mpf/devices/combo_switch.py
+++ b/mpf/devices/combo_switch.py
@@ -90,7 +90,8 @@ class ComboSwitch(SystemWideDevice, ModeDevice):
         self._remove_switch_handlers()
         self._kill_delays()
 
-        # Ensure reset to inactive state when mode ends, in case combo switch event is triggered mode end, or combo switch was partially engaged as mode ended.
+        # Ensure reset to inactive state when mode ends, in case combo switch event triggers mode end,
+        # or the combo switch was partially engaged as mode ended.
         self._reset_state()
 
     def _reset_state(self):
--- a/mpf/devices/combo_switch.py
+++ b/mpf/devices/combo_switch.py
@@ -90,7 +90,7 @@ class ComboSwitch(SystemWideDevice, ModeDevice):
         self._remove_switch_handlers()
         self._kill_delays()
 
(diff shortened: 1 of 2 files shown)


mpf/devices/combo_switch.py
```diff
@@ -88,6 +88,20 @@ def device_removed_from_mode(self, mode):
...
+        # The switches may still be held when the mode stops. This happens
+        # whenever the combo's own event ends the mode that owns it, e.g. a
+        # "hold both buttons to cancel the game" combo defined in a game mode.
+        # With the handlers gone the eventual release is never seen, so any
+        # state left here would latch for the life of the process:
+        # _switch_1_went_active and _switch_2_went_active both return early on
+        # "if self._switches_N_active", and _switch_state returns early on
+        # "if state == self.state". The combo would then fire exactly once and
+        # never again. The device observes nothing while unloaded, so the only
+        # honest state to hold is none at all.
```
bosh:
```suggestion
        # Ensure reset to inactive state when mode ends, in case combo switch devices are active while mode is ending.
```

mpf/devices/combo_switch.py
```diff
@@ -88,6 +88,20 @@ def device_removed_from_mode(self, mode):
...
+        self._state = 'inactive'
+        self._switches_1_active = False
+        self._switches_2_active = False
```
bosh, line 100: I think this solution works for 90% of cases but itself introduces a new possible device-state-synchronization bug. Indeed if the mode ends while the combo is held (whether because of the combo event itself or some other reason), it _is_ possible that the mode _will start again_ while the combo is still held. Seems exceedingly unlikely, but what that means is that assuming that mode-end means release may not be exactly correct either.

TBH I think the inactive/false cleanup is probably reasonable, but that with or without it, the solution is at mode-start-time for a combo switch in a mode, it should check the live states and sync itself to them then. However, this might mean you have to deal with combos that behave kinda funny if a switch was held while it activates. I'm not sure if you'd run into "that hold doesnt count, start again" kind of non-detections, or whether the pre-held switch would allow just the addition of the other combo member to cause the combo to fire .... and I'm not sure which of those two options is more correct.

bosh: I've thought more on the right behavior. I think the root issue is that combo switches are more like switches (machine file only) than shots (mode config only) - they represent a hardware detection interface, not a domain layer rules management object. IMO that means you should only ever define them in machine files in the first place, and the mode-scoped variant is something that I would deprecate (to align with sequence_shots). But ofc I'm not trying to do that either :)

The best option for the mode-specific version of the combo switch is to assume it needs to fully reset every time it starts, forcing the player to begin the combo from scratch after mode start. I think this aligns it with how other mode devices work. The more complex options for how we could handle combo switch state (e.g. detect-initial-state or persist-across-stop/start) can be handled by a machine combo switch and a mode-scoped event player.

So maybe the best move is to make a private reset_state fn here and call it both on mode start and stop.

Pyrrvs:
Ok I think I understand your reasoning. I should probably move my combos to the machine file instead of the mode layer and have my mode react to events posted by the machine layer combo, right?

In the meantime, I updated the PR based on your suggestions: the reset logic moved to _reset_state() and it now runs on both mode start and stop, so a switch already held when the mode starts does not count. Also added tests for a mode restarted while its switches are held.

mpf/devices/combo_switch.py
```diff
@@ -88,6 +90,14 @@ def device_removed_from_mode(self, mode):
         self._remove_switch_handlers()
         self._kill_delays()
 
+        # Ensure reset to inactive state when mode ends, in case combo switch event is reason mode ended.
```
bosh:
```suggestion
        # Ensure reset to inactive state when mode ends, in case combo switch event is triggered mode end, or combo switch was partially engaged as mode ended.
```

bosh (approved):
Thanks for touching up it! Good fix :)

mpf/devices/combo_switch.py
```diff
@@ -88,6 +90,14 @@ def device_removed_from_mode(self, mode):
         self._remove_switch_handlers()
         self._kill_delays()
 
+        # Ensure reset to inactive state when mode ends, in case combo switch event is triggered mode end, or combo switch was partially engaged as mode ended.
```
bosh:
```suggestion
        # Ensure reset to inactive state when mode ends, in case combo switch event triggers mode end,
        # or the combo switch was partially engaged as mode ended.
```

